Issue #212, opened 2012-04-10
synchrony read timeout


eabbott:
In 1.9.2 with synchrony and hiredis there is no read timeout.
To simulate use iptables locally to drop outgoing packets to redis after making the connection.
1. start EM/rails/stack with a redis timeout value set
2. run a request against redis to open the connection
3. use iptables to drop packets: iptables -A OUTPUT -p tcp --dport 6379 -j DROP
4. run a request against redis, note the timeout is ineffective

PR #213, opened 2012-04-10
read timeout for synchrony driver

eabbott:
Add a read timeout for the synchrony driver as in issue #212


djanowski:
If anyone wants to revisit this, please be my guest. We'll probably remove built-in support for `em-synchrony` soon.

PR #313, opened 2013-02-02
Add support for CLIENT commands

bitterb:
This pull request adds support for `CLIENT [LIST | KILL | GETNAME | SETNAME]` commands. 

I think that some of tests might be able to be improved. For example, the current test for `CLIENT KILL` requires Redis 2.6.9 or later because the test depends on `CLIENT SETNAME`. However, the test should not use such newer commands and run on Redis 2.4.0 or later. If you have any ideas for the improvement of tests, please let me know.

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -7,8 +7,6 @@ class Redis
     $stderr.puts "\n#{message} (in #{trace})"
   end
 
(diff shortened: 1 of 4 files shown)


djanowski:
Implemented in 4.0.

PR #329, opened 2013-03-25
Added subscriptions with timeout

somebody32:
Hi,

I've added subscribe and psubscribe with timeout methods. That is useful when publisher could die for an any reason and subscriber should not hang endlessly waiting. 

So if there is no data published for a timeout period then Redis::Timeout error will be raised.


djanowski:
This landed in master (and 3.3).

PR #354, opened 2013-07-26
Fix distributed db ring collision

axos88:
I have added a test to the HashRing#add_node that will test if a node ID's CRC checksum is already present and will throw an exception in such a case.
This enables to prevent quiet CRC collisions and adding databases which have the same ID, but are not identical. I had this error when adding multiple MockRedis objects to the ring which had the same ID.


djanowski:
Sorry for the delay. Thank you.

djanowski:
This is available in 3.2.2.

axos88:
lolk, your are welcome :)

PR #453, opened 2014-07-23
enable to set options by query_string

masarakki:
options:
- timeout
- inherit_socket

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ redis = Redis.new(:host => "10.0.1.1", :port => 6380, :db => 15)
 You can also specify connection options as an URL:
 
 ```ruby
-redis = Redis.new(:url => "redis://:p4ssw0rd@10.0.1.1:6380/15")
+redis = Redis.new(:url => "redis://:p4ssw0rd@10.0.1.1:6380/15?timeout=3")
 ```
 
 By default, the client will try to read the `REDIS_URL` environment variable
@@ -166,7 +166,7 @@ end
       to redis, AND
     - your own code prevents the parent process from using the redis
       connection while a child is alive
-   
+
    Improper use of `inherit_socket` will result in corrupted and/or incorrect
    responses.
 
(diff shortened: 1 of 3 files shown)


djanowski:
I don't want to get too philosophical, but URLs describe the resource -- not all the implementation details when you try to access it. Similarly, a HTTP URL doesn't allow you to set the `Accept-Encoding` header.

For instance, if you're on a PaaS like Heroku, would it make sense that they could inject the connection timeout or other app-specific logic options via the URL? I don't think so.

If the motivation for this change is having a single configuration setting to set up the Redis client for convenience, then I don't favor it. If there's more that I'm not seeing, please let me know.

beedub:
@djanowski I think your point makes sense. How about having a separate ENV variables like `REDIS_TCP_TIMEOUT` and `REDIS_WRITE_TIMEOUT`? A natural consequence is that redis might accept dozens of configuration env vars, but that doesn't seem wrong in of itself.

djanowski:
@beedub Thank you for your response.

Right now I'm leaning towards 'roll your own'. Have a single `REDIS_OPTIONS`, query-string encoded maybe, and parse it yourself. If we see this is something many people want, then we could consider adding it into the gem.

beedub:
I don't disagree with only including it if "many people want", but what would be the criteria? Atleast two people ( me and @masarakki ) so far have pull requested two code changes with variants of the underlying need. I'd also imagine there are atleast a few more that would find this useful.

The reason why I'd like to see this merged is because I need to set tcp timeout on every app deployed to EC2 and GCE (related to https://github.com/redis/redis-rb/pull/482). Currently, this means having to copy-and-paste the redis config decoding code into each app and incur the obvious costs of copy-and-pasted code.

mrgordon:
Bumping @beedub's comment

djanowski:
If env variables are an option, then why not simply:

```ruby
redis = Redis.new(inherit_socket: ENV["REDIS_INHERIT_SOCKET"] == "1")
```

PR #481, opened 2014-12-03
Parse query string as options

beedub:
I really only needed this for `tcp_keepalive`, but I added `driver`, `timeout`, and `inherit_socket` as well.


djanowski:
Please see [my comment](https://github.com/redis/redis-rb/pull/453#issuecomment-71667464) on #453.

PR #482, opened 2014-12-08
Write timeout

beedub:
Writes will block indefinitely without timeout if the socket's write buffer space is exhausted and the host is unreachable. This happens in EC2 when a connection has been closed and is idle for more than 60 seconds.

I couldn't figure out a way to simulate host unreachable (as that probably requires going into the OS' tcp layer), so I added a kinda disgusting test using monkeypatching. Any suggestions on how to make it less disgusting?

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -13,6 +13,7 @@ class Redis
       :path => nil,
       :timeout => 5.0,
       :connect_timeout => 5.0,
+      :write_timeout => nil,
       :password => nil,
       :db => 0,
       :driver => nil,
@@ -426,6 +427,8 @@ class Redis
       else
         options[:timeout]
       end
+
+      options[:write_timeout] = options[:write_timeout] ? options[:write_timeout].to_f : options[:timeout]
 
       options[:db] = options[:db].to_i
       options[:driver] = _parse_driver(options[:driver]) || Connection.drivers.last
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -12,7 +12,7 @@ class Redis
       def initialize(*args)
         super(*args)
 
-        @timeout = nil
+        @timeout = @write_timeout = nil
         @buffer = ""
       end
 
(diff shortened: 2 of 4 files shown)


djanowski:
I spent quite some time trying to figure out a way to simulate the condition. I was about to give up when I re-read your description and realized I needed to exhaust the socket buffer.

This is what I came up with: https://github.com/redis/redis-rb/compare/beedub-write_timeout?expand=1#diff-49f3002b4c26172aa701e9046574e1eaR163

What do you think?

beedub:
@djanowski posted comments on https://github.com/redis/redis-rb/commit/d26f6e18698342f461380e11819e36df629250a4

djanowski:
Fixed stuff based on your comments.

Would you like to merge that branch and update this PR?

beedub:
@djanowski I merged your changes. On cursory glance the test failures look like they're subject to platform differences?

badboy:
The test makes no sense for the proposed feature. 
You're setting a write timeout, but writing against an idle server will still succeed (and it does succeed in under 0.2 seconds), that's why the outer `timeout` in the test fires, leading to the failing test.

beedub:
oh duh. @badboy any thoughts about a test that would work? I originally had https://github.com/beedub/redis-rb/commit/fea04ea18e367f73f04842411372d6baf115fb2c#diff-49f3002b4c26172aa701e9046574e1eaR163, a solution that felt dirty but worked

badboy:
Aaah, sorry for the quick and harsh words, @djanowski informed me that it does fail when the bufferi s exceeded. On Linux it's probably bigger

beedub:
the closest thing I could find to determining the size of the socket write buffer is using

```ruby
val = TCPSocket.new("127.0.0.1", 6000).getsockopt(Socket::SOL_SOCKET, Socket::SO_SNDBUF).unpack("i")[0]
=> 50748
```
but I tested this on ubuntu 10.04 and the write would block somewhere between 3-4x the size in `val`. I'm not sure how it is on other servers.

djanowski:
This is merged in master.

PR #494, opened 2015-01-07
Sentinel: Be more resilient when trying to connect to sentinels

zmack:
Ran into this thing the other day.

The issue is that if one of your sentinels is down ( given that the gem tries them in order, say the first one ), or the connection times out or whatever, this will just throw up at that point.

The behavior I would like to see is for it to try all the other sentinels and if it can't connect to any of them, bail.

Doesn't really make sense otherwise, since instead of being more resilient, your setup now relies on all the sentinels needing to be up, which feels slightly strange as a design choice.

Let me know if I'm reading this wrong.

Couldn't find any sentinel tests to update around the test folder, so I'm assuming that'll be a separate issue ( actually writing tests for the sentinel implementation )

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -461,7 +461,9 @@ class Redis
     end
 
     class Connector
+      attr_reader :logger
       def initialize(options)
+        @logger = options[:logger]
         @options = options
       end
 
@@ -522,6 +524,11 @@ class Redis
                 @sentinels.unshift(sentinel)
 
                 return result
+              end
+            rescue CannotConnectError
+              if logger
+                msg = "[Redis] Could not connect to sentinel #{sentinel[:host]}:#{sentinel[:port]}"
+                logger.debug(msg)
               end
             ensure
               client.disconnect


nick-desteffen:
I just noticed this as well.  If it bombs on the fist sentinel it should most definitely try the next, as that is the whole point of giving it a pool of sentinels.

badboy:
Will take a look at it tomorrow.

badboy:
Sorry for the late reply. Idea & code look good (afterall it's just a 7 line change, right?)
Test failures should be gone with cc634ea73ac86e9fbfb4d4b0c1c327b2e6ecc7a6

Sign-off from me. :)

nick-desteffen:
So any chance this could be merged?

badboy:
People are busy. I'm sure this will be merged soon.

zmack:
@nick-desteffen fwiw we just monkeypatched it in.

If a new version gets cut, I'm hoping this'll make it in :)

djanowski:
Looking at it now.

djanowski:
I went ahead and implemented what you described: try all sentinels, 3 times each, and raise if none is available.

https://github.com/redis/redis-rb/compare/zmack-fix_unconnectable_sentinel

What do you all think?

nick-desteffen:
:+1:

nick-desteffen:
Thank you for taking the time to look into this! :dog:

djanowski:
@badboy @zmack I'll wait for your +1s before merging.

zmack:
Looks good :+1:

I'll see if i can't get some work done around adding tests and making all of those options customizable ( having a mandatory .3 sleep in there makes me a bit antsy ).

Thanks for getting this sorted, really appreciate it.

badboy:
I'm on @zmack's side here. A mandatory sleep seems not to good.
Do we really need to try 3 times?
@zmack's PR here has no good way to let the application using it know no Sentinel could be contacted (only that it was impossible to fetch the right role, but this can be to other errors as well), so I get why we would want to raise a new exception here.

So summarizing, my suggestion:

* Only iterate once
* leave out the `sleep`
* Properly document what exception can be raised

djanowski:
@badboy I decided to iterate because of the very short connect timeout hard-coded in there. Maybe we should honor the `reconnect_attempts` option, if given?

badboy:
We already got an option for that? Sounds good then

djanowski:
Assuming we merge #505, then honoring `reconnect_attempts` looks like this: https://github.com/redis/redis-rb/compare/sentinel-tests...zmack-fix_unconnectable_sentinel

djanowski:
Thanks @zmack.

PR #496, opened 2015-01-11
Preliminary native SSL support

tarcieri:
This adds two options which can be used to configure SSL support:

* ssl: (true/false) - Enable SSL for this connection
* ssl_params: (Hash) - Params to OpenSSL::SSL::SSLContext to provide. Each param must be a method name on the OpenSSL::SSL::SSLContext object.

It can be used like follows:

```Ruby
Redis.new(:host => "127.0.0.1", :port => 6379, :ssl => true)
```


badboy:
build failures are unrelated to your changes. Seems like some of the test environments have a super new Redis already with Quicklist integrated. Also: we have the assertion argument order the wrong way which lead to this subtle unclear error message.

djanowski:
@badboy Your fix for the unrelated failure is in, so @tarcieri can merge master here.

I wonder if this should be a different scheme in the URL? Like `redis+ssl`?

Also, I'd like to add some tests to this. Let me know if you can do it or I'll do it.

badboy:
`redis+ssl` instead of `:ssl => true`?

Before this can be merged it definitely needs tests and documentation.

tarcieri:
I'd also note that upstream support for SSL hasn't been merged into Redis yet, however the person working on it was blocked on having Ruby support so I went ahead and made this.

tarcieri:
I'm also okay with something like `redis+ssl` as a URL in lieu of `:ssl => true`, but I would definitely like to have `:ssl_params` (we need to configure certificates and keys among other things)

djanowski:
I'll be delighted to revisit once SSL is supported by Redis. Thanks!

tarcieri:
@djanowski that's likely to be never given the state of [Redis #2178](https://github.com/antirez/redis/issues/2178) however this is still useful for anyone running an SSL terminating proxy on the Redis side as it means they don't have to run SSL terminating proxies on both sides.

djanowski:
I want to merge this but I'm thinking SSL should be supported in our two main drivers, Ruby and hiredis. @tarcieri can you look into that? I'm happy to add the missing tests etc

tarcieri:
@djanowski as I understand it, hiredis performs I/O operations in C and does not have native support for SSL, so I don't think hiredis+ssl is trivial unless there's native SSL support I'm unaware of

tarcieri:
This unceremoniously closed issue would seem to confirm that hiredis lacks SSL support:

https://github.com/antirez/redis/issues/675

tarcieri:
I just rebased on master. We'll see if the tests pass now...

djanowski:
Yeah, hiredis doesn't support SSL. What I meant is, can you try to add it?

tarcieri:
@djanowski no. That would require modifying hiredis's C source code, and integrating it with a library like OpenSSL, effectively closing https://github.com/antirez/redis/issues/675

The reason this patch is so small is because it's using a socket duck type in conjunction with the Ruby code.

There's a lot of other stuff this patch could probably use: documentation, tests, verifying hostnames(!), but I don't think a native hiredis SSL adapter is going to happen.

tarcieri:
Also this is passing on Travis but failing on CircleCI. The CircleCI failure seems unrelated: it can't start Redis?

tarcieri:
@djanowski anything I can do here to help get this merged?

tarcieri:
@djanowski to add a bit of urgency to this, it seems to be a blocker for getting SSL support merged upstream, as Redis uses redis-rb in its test suite...

djanowski:
Tests would be nice + explicit exception when trying to use ssl with a
driver other than ruby...

tarcieri:
@djanowski thanks!

What do you think about "redis+ssl" URLs as @badboy noted? (`rediss://`?)

Apparently this is a blocker for upstream Redis SSL support as `redis-rb` is part of the overall Redis test suite, so it'd be good to get this fixed!

badboy:
The [IANA uri scheme for Redis](http://www.iana.org/assignments/uri-schemes/prov/redis) lists `rediss:` as used by some clients already.

tarcieri:
We've been trying to terminate Redis client SSL natively at Square and have something internally that's working.

I just pushed 6a02e77 which should be much cleaner than my previous attempts. It adds `Redis::Connection::SSLSocket` which includes `Redis::Connection::SocketMixin`, and is used if the `rediss://` URI scheme is passed in.

Still no tests, but I think I could probably whip up a `MockSslRedisServer` or something to that effect which responds to `PING` messages, and make sure it's able to successfully send a ping through that.

djanowski:
> Still no tests, but I think I could probably whip up a `MockSslRedisServer` or something to that effect which responds to `PING` messages, and make sure it's able to successfully send a ping through that.

Sounds good!

tarcieri:
@badboy @djanowski I just pushed my latest changes to the PR, and I'd actually consider this complete aside from actually writing `test/ssl_test.rb`. I did the preliminary work on `test/support/redis_mock.rb` to add SSL support to `RedisMock::Server` so I can use it in tests as well.

We are running a monkeypatch of this against `redis-rb` in production now, and it seems to be working. If this looks good I can write `test/ssl_test.rb` (hardest part is generating all of the certs/keys needed for the test, which I'd want to do programmatically)

badboy:
Looks quite minimal in deed. I like that. Tests would be appreciated. I'll check the implementation later today. 
Test failures are Ruby 1.8.7 related, but not due to your changes (I'll take care of that ASAP)

djanowski:
Definitely option 2.

tarcieri:
Updated with client certs and tests. It seems the `IO::WaitReadable` API for non-blocking I/O was not standardized until 2.0.0, so for the time being the SSL tests target that unless I can figure out a backward compatible option.

Some of the tests are failing with `SSL_accept returned=1 errno=0 state=unknown state: unknown protocol`. Not sure why. Other than that I'd call this complete. I'll continue investigating the SSL errors.

tarcieri:
@badboy @djanowski tests are green now!

lib/redis/connection/ruby.rb
```diff
@@ -2,6 +2,7 @@
 require "redis/connection/command_helper"
 require "redis/errors"
 require "socket"
+require "openssl"
```
tarcieri: Of anything I could foresee actually causing problems it's this. How do you feel about a hard dependency on the OpenSSL extension? It's generally part of the standard library but I could see this causing problems for people who disable it...

tarcieri:
@badboy @djanowski I'd call this good to go now. It now raises `NotImplementedError` if you try to use the hiredis or synchrony drivers with SSL, and I also added documentation to the README.

lib/redis/connection/ruby.rb
indirect: Speaking from direct experience with Bundler, _tons_ of people have Ruby installed without being able to require `openssl`. 😕 Since Ruby will happily compile and install without any OpenSSL support, I suspect you'll want to `rescue LoadError` here.

tarcieri: All right, I'll add fallback support for if OpenSSL is unavailable

tarcieri: 38dd370 should ensure the absence of the OpenSSL extension is tolerated

tarcieri:
@badboy @djanowski okay one last update, it should now load on Rubies that do not have the OpenSSL extension enabled (however SSL will not work for obvious reasons)

I think this is finally complete. The "wip" label can be safely removed at this point :wink:

badboy:
Thanks for the work. I will check the code this week, so we hopefully can land this soon

tarcieri:
@badboy awesome, thanks!

badboy:
LGTM. Merge is up to @djanowski

tarcieri:
@djanowski WDYT?

djanowski:
Already merged in the 3.3 branch, we got stuck with a potential memory leak
(unrelated to this) before the release, but since we can't confirm it 100%,
I'll probably release today or tomorrow.
On Apr 16, 2016 18:48, "Tony Arcieri" <notifications@github.com> wrote:

@djanowski <https://github.com/djanowski> WDYT?

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/redis/redis-rb/pull/496#issuecomment-210910577>

tarcieri:
@djanowski awesome, thanks! Great to hear it!

lib/redis/connection/ruby.rb
```diff
@@ -3,12 +3,22 @@
...
+
 class Redis
   module Connection
     module SocketMixin
 
       CRLF = "\r\n".freeze
 
+      # Exceptions raised during non-blocking I/O ops that require retrying the op
+      NBIO_EXCEPTIONS = [Errno::EWOULDBLOCK, Errno::EAGAIN]
+      NBIO_EXCEPTIONS << IO::WaitReadable if RUBY_VERSION >= "1.9.3"
```
nerdrew: @tarcieri Thoughts on using nio4r here instead of re-implementing the `IO.select` logic around `IO::WaitReadable` and `IO::WaitWritable`>

We are seeing `OpenSSL::SSL::SSLErrorWaitWritable` errors when writing large keys. cc @mdayaram

tarcieri:
```ruby
>> OpenSSL::SSL::SSLErrorWaitWritable.ancestors
 => [OpenSSL::SSL::SSLErrorWaitWritable, IO::WaitWritable, OpenSSL::SSL::SSLError, OpenSSL::OpenSSLError, StandardError, Exception, Object, Kernel, BasicObject]
```

That should be a `IO::WaitWritable`

Issue #501, opened 2015-01-19
Requests end up interleaved on Ruby 2.1 when Timeout module is used


tarcieri:
The following snippet is able to reproduce a condition whereby the response of one command is never consumed, and the next command ends up consuming the response from the previous command:

https://gist.github.com/tarcieri/303c5dcf803b0e5d021d

Here's another reproduction of the same issue, I believe:

https://gist.github.com/stevehodgkiss/10204388

The issue manifests for me after 1000+ iterations of the loop. It *only* manifests on Ruby 2.1, and not on previous versions.

It may very well be a bug in Ruby's Timeout implementation, but I thought I'd start here.

antirez:
Thank you @tarcieri, made an hypothesis about what could be the cause in issue #490. Probably Damian will identify the root cause much easily btw.

tarcieri:
Given this only manifests on Ruby 2.1, I'm actually leaning towards it being a bug in Ruby itself (specifically in the 2.1+ implementation of Timeout)

antirez:
Oh strange it only affects only 2.1, yep probably you are right. In the general form the bug looks like something every client out there should have a regression for... very easy to have or re-introduce it.

badboy:
It might very well be the `Timeout` class, it's [broken afterall](http://headius.blogspot.de/2008/02/rubys-threadraise-threadkill-timeoutrb.html). Maybe 2.1 is just faster by now and thus this racing conditions manifest. I'll try tracking it down tomorrow.

zanker:
One thing to note, is that while the timeout issue is the most common case this happens, killing a thread can cause it too regardless of version (or more accurately, I've only tested on 2.0/2.1/JRuby 1.7).

Keeping track of whether or not you have a request outstanding when writing data on anything that's not thread safe handles all of the corruption problems though (like in https://github.com/mperham/connection_pool/pull/67, https://github.com/mperham/connection_pool/pull/70)

tamird:
> The issue manifests for me after 1000+ iterations of the loop.

This is because the timeout is very high 0.5s -- setting it to 0.001s manifests the issue after one iteration.

> It only manifests on Ruby 2.1, and not on previous versions.

Definitely manifests on 2.2 as well, but interestingly not on 2.0/rbx/jruby.

zmack:
I've run into this one as well, and I've seen similar bugs happen with the ruby mogilefs client.

[...]

PR #502, opened 2015-01-27
Fix inconsistent connections when using Timeout under MRI 2.1+.

djanowski:
This should address #501.

I like this solution because the performance and complexity penalty of keeping counters is limited to the pure Ruby driver. Edit: From the comments below, I'll test what happens with Hiredis (I thought it wasn't an issue).

The 'test cases' were [this](https://gist.github.com/tarcieri/303c5dcf803b0e5d021d) and [this](https://gist.github.com/stevehodgkiss/10204388). I'll try to make those proper tests cases in our code base.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -77,6 +77,8 @@ class Redis
       @connection = nil
       @command_map = {}
 
+      @pending_reads = 0
+
       if options.include?(:sentinels)
         @connector = Connector::Sentinel.new(@options)
       else
@@ -243,12 +245,15 @@ class Redis
 
     def read
       io do
-        connection.read
+        value = connection.read
+        @pending_reads -= 1
+        value
       end
     end
 
     def write(command)
       io do
+        @pending_reads += 1
         connection.write(command)
       end
     end
@@ -315,6 +320,7 @@ class Redis
       @options[:port] = server[:port]
 
       @connection = @options[:driver].connect(server)
+      @pending_reads = 0
     rescue TimeoutError,
            Errno::ECONNREFUSED,
            Errno::EHOSTDOWN,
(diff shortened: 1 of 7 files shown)


tamird:
can you add some test cases to demonstrate that this is correct?

tamird:
i believe in our experience hiredis was also vulnerable to this issue -- is that meant to be addressed here?

mperham:
Can't any driver become corrupt or does hiredis already perform similar checks?

djanowski:
@tamird Just updated the PR description with links to the 'test cases' that I used (originally posted on #501). I'll try to make them proper tests in our code base.

Also, I thought from @tarcieri's comment that Hiredis wasn't an issue, but I'll look into it (in fact my first take at this fixed all drivers, but I took another approach in order to make just the Ruby driver pay the penalty).

djanowski:
@mperham ^^^

tarcieri:
@djanowski it's a problem if you use Timeout with hiredis. I was saying that hiredis implements its own timeouts using asynchronous I/O and those are fine.

djanowski:
@tarcieri Yes I understood that afterwards. Since it provides native timeouts, there's usually no need to use external `Timeout.timeout` calls.

OK, I'll go ahead and fix hiredis too. May still have my original implementation in a stash.

djanowski:
@igrigorik If you think the Synchrony driver is affected by this, please chime in.

yaauie:
I think this approach makes a lot of sense, and that the additional overhead is both small & justified. :+1:

fw42:
Thanks a lot for this fix, we ran into the same issue under high load.

fw42:
Any idea when there will be a Rubygems release available that includes this change?

badboy:
As soon as @djanowski pushes it ;)

djanowski:
Will look into this in the next 24 hours

Issue #486, opened 2014-12-10
How to rescue from DNS resolution errors in #quit safely?


tcrayford:
There's a case where you can get an exception calling `#quit` right now:

- if you're using the `hiredis` client
- and you are using DNS names instead of IP addresses to talk to your redis servers
- and dns resolution fails

The tricky part is, I can't work out how to catch this easily without patching `hiredis` - `hiredis` only offers up a `RuntimeError` when it can't resolve the DNS name.

Relevant stacktrace:

```
E, [2014-12-08T11:03:04.431005 #23177] ERROR -- : Name or service not known (RuntimeError)
ruby/2.1.0/gems/redis-3.0.6/lib/redis/connection/hiredis.rb:16:in `connect'
ruby/2.1.0/gems/redis-3.0.6/lib/redis/connection/hiredis.rb:16:in `connect'
ruby/2.1.0/gems/redis-3.0.6/lib/redis/client.rb:273:in `establish_connection'
ruby/2.1.0/gems/redis-3.0.6/lib/redis/client.rb:69:in `connect'
ruby/2.1.0/gems/redis-3.0.6/lib/redis/client.rb:292:in `ensure_connected'
ruby/2.1.0/gems/redis-3.0.6/lib/redis/client.rb:179:in `block in process'
ruby/2.1.0/gems/redis-3.0.6/lib/redis/client.rb:258:in `logging'
ruby/2.1.0/gems/redis-3.0.6/lib/redis/client.rb:178:in `process'
ruby/2.1.0/gems/redis-3.0.6/lib/redis/client.rb:84:in `call'
ruby/2.1.0/gems/redis-3.0.6/lib/redis.rb:103:in `block in quit'
ruby/2.1.0/gems/redis-3.0.6/lib/redis.rb:36:in `block in synchronize'
/opt/rubies/2.1.0/lib/ruby/2.1.0/monitor.rb:211:in `mon_synchronize'
ruby/2.1.0/gems/redis-3.0.6/lib/redis.rb:36:in `synchronize'
ruby/2.1.0/gems/redis-3.0.6/lib/redis.rb:101:in `quit'
```

djanowski:
So #506 should address this as well, right?

djanowski:
Resolved by #506.

PR #506, opened 2015-02-10
Add `Redis#disconnect!` as a public-API way of disconnecting.

djanowski:
This should address the fact that `Redis::Client` will eventually private and `Redis#client` will represent the `CLIENT` command.

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -54,6 +54,11 @@ class Redis
     @original_client.connected?
   end
 
+  # Disconnect the client as quickly and silently as possible.
+  def disconnect!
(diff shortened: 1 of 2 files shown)


djanowski:
@badboy?

tcrayford:
This will still throw DNS connection errors when calling `disconnect` right? In which case it doesn't fix #485 or #486 at all

yaauie:
@tcrayford: I don't believe it will; the test verifies that `quit` is not sent.

tcrayford:
ahh, gotcha. Cool then

Issue #507, opened 2015-02-11
Client possibly leaking state under high load


fw42:
We are running into some problems that look very related to https://github.com/redis/redis-rb/pull/502, https://github.com/redis/redis-rb/issues/501.

We have code that looks like this:
```ruby
result = redis.pipelined do
  redis.incrby(key_name, 1)
  redis.expire(key_name, 1.day)
end
```
and we occasionally get an exception that says:
```NoMethodError: undefined method `each_slice' for nil:NilClass```

The backtrace looks as follows:
```
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis.rb:2579 → block in _hashify
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/pipeline.rb:120 → call
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/pipeline.rb:120 → _set
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/pipeline.rb:67 → block in finish
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/pipeline.rb:66 → each
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/pipeline.rb:66 → each_with_index
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/pipeline.rb:66 → each
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/pipeline.rb:66 → map
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/pipeline.rb:66 → finish
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/client.rb:149 → block in call_pipeline
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/client.rb:279 → with_reconnect
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis/client.rb:147 → call_pipeline
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis.rb:2131 → block in pipelined
[GEM_ROOT]/bundler/gems/redis-rb-26a4309f30b8/lib/redis.rb:37 → block in synchronize
[...]

PR #509, opened 2015-02-12
Fix possible race condition with pipelines

byroot:
Fixes: https://github.com/redis/redis-rb/issues/507

So the reason behind #507, is that the expiry was passed as `1.month`, and that the evaluation of that expression by the Redis client was actually triggering a call to Redis to fetch the translation data.

It's obviously a huge corner case, but I think it would have been prevented by that PR.

@fw42 @djanowski @badboy for review please.

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -2130,13 +2130,14 @@ class Redis
 
   def pipelined
     synchronize do |client|
+      pipeline = Pipeline.new
       begin
-        original, @client = @client, Pipeline.new
+        original, @client = @client, pipeline
         yield(self)
-        original.call_pipeline(@client)
       ensure
         @client = original
       end
+      @client.call_pipeline(pipeline)
     end
   end
 
@@ -2175,14 +2176,14 @@ class Redis
       if !block_given?
         client.call([:multi])
       else
+        pipeline = Pipeline::Multi.new
         begin
-          pipeline = Pipeline::Multi.new
           original, @client = @client, pipeline
           yield(self)
-          original.call_pipeline(pipeline)
         ensure
           @client = original
         end
+        @client.call_pipeline(pipeline)
       end
     end
   end
--- a/lib/redis/pipeline.rb
+++ b/lib/redis/pipeline.rb
@@ -122,7 +122,7 @@ class Redis
     end
 
     def _command
-      @command
+      @command.map(&:to_s)
     end
 
     def value
--- a/test/pipelining_commands_test.rb
+++ b/test/pipelining_commands_test.rb
@@ -239,4 +239,28 @@ class TestPipeliningCommands < Test::Unit::TestCase
 
     assert_equal 3, r.client.db
   end
+
+  class LazyString
+    def initialize(redis)
+      @redis = redis
+    end
+
+    def to_s
+      @redis.hgetall('foo').inspect
+    end
+  end
+
+  def test_pipeline_race_condition
+    ttl = LazyString.new(r)
+    r.pipelined do
+      r.expire('bar', ttl)
+    end
+  end
+
+  def test_multi_race_condition
+    ttl = LazyString.new(r)
+    r.multi do
+      r.expire('bar', ttl)
+    end
+  end
 end


lib/redis.rb
```diff
@@ -2130,13 +2130,14 @@ def unwatch
...
+      pipeline = Pipeline.new
       begin
-        original, @client = @client, Pipeline.new
+        original, @client = @client, pipeline
         yield(self)
-        original.call_pipeline(@client)
       ensure
         @client = original
       end
+      original.call_pipeline(pipeline)
```
byroot: The key of the fix is to reassign the original client before evaluating the pipeline.

I'll work on a test case to demo that.

byroot:
I just added a test case reproducing what we had in production. If you run it against master you'll get:

```
TestPipeliningCommands#test_pipeline_race_condition:
NoMethodError: undefined method `bytesize' for 0:Fixnum
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/connection/command_helper.rb:19:in `block in build_command'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/connection/command_helper.rb:10:in `each'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/connection/command_helper.rb:10:in `build_command'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/connection/ruby.rb:269:in `write'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:257:in `block in write'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:236:in `io'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:255:in `write'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:214:in `block (3 levels) in process'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:208:in `each'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:208:in `block (2 levels) in process'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:353:in `ensure_connected'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:207:in `block in process'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:310:in `logging'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:206:in `process'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:177:in `call_pipelined'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:149:in `block in call_pipeline'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:279:in `with_reconnect'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis/client.rb:147:in `call_pipeline'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis.rb:2136:in `block in pipelined'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis.rb:37:in `block in synchronize'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/monitor.rb:211:in `mon_synchronize'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis.rb:37:in `synchronize'
    /Users/byroot/workspace/ruby/redis-rb/lib/redis.rb:2132:in `pipelined'
    /Users/byroot/workspace/ruby/redis-rb/test/pipelining_commands_test.rb:271:in `test_pipeline_race_condition'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/minitest/unit.rb:1265:in `run'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/test/unit/testcase.rb:17:in `run'
    /Users/byroot/workspace/ruby/redis-rb/test/helper.rb:70:in `run'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/minitest/unit.rb:940:in `block in _run_suite'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/minitest/unit.rb:933:in `map'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/minitest/unit.rb:933:in `_run_suite'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/test/unit.rb:663:in `block in _run_suites'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/test/unit.rb:661:in `each'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/test/unit.rb:661:in `_run_suites'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/minitest/unit.rb:884:in `_run_anything'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/minitest/unit.rb:1092:in `run_tests'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/minitest/unit.rb:1079:in `block in _run'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/minitest/unit.rb:1078:in `each'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/minitest/unit.rb:1078:in `_run'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/minitest/unit.rb:1066:in `run'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/test/unit.rb:27:in `run'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/test/unit.rb:780:in `run'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/test/unit.rb:372:in `block (2 levels) in autorun'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/test/unit.rb:33:in `run_once'
    /Users/byroot/.rvm/rubies/ruby-2.1.3/lib/ruby/2.1.0/test/unit.rb:371:in `block in autorun'
```

byroot:
Oops. I pasted the wrong backtrace sorry. I just updated it.

lib/redis.rb
badboy: How would it have prevented that? Commands in the block are already executed on the above `yield`

byroot: By reseting `@client` before `call_pipeline`, so that if another command is issued inside `call_pipeline` it isn't appended to the pipeline being executed.

byroot:
So actually my fix seems to not be enough, as shown by the failing test case. It appear `LazyString#to_s` end up closing the stream somehow.

lib/redis.rb
davidcornu: @byroot would `@client.call_pipeline(pipeline)` be clearer here (as `original` is defined in the `begin`)?

badboy: Ah, yes, I understand it now.
I would have liked if instead of modifying the client directly, we would have an API like:

```ruby
redis.pipelined do |pipe|
  pipe.hgetall(:key)
end
```
without affecting the outer Redis object, sadly because of backwards-compatibility we need to keep the current one (maybe I can think of a good way to extend it)

badboy:
Hm, gonna take a look at that tomorrow.

lib/redis.rb
byroot: > I would have liked if instead of modifying the client directly, we would have an API like:

We were talking about that with @davidcornu, and definitely that would be a huge improvement. Maybe that can be worked out in a backward compatible way with `Proc#arity`:

```ruby
def pipelined(&block)
  if block.arity == 1
    yield pipeline
  else
    @client = pipeline
    yield
    ....
  end
end
```

davidcornu:
@byroot the `multi` method probably has the same issue https://github.com/redis/redis-rb/blob/be4326e7f2d399069831ba05ef53ba51c9a76c7e/lib/redis.rb#L2173-L2188.

byroot:
@davidcornu yes we initially had the problem with `multi` in #507. Whatever the definitive fix for this is, it should also be applied to `multi`.

byroot:
I reworked the fix a bit and added a test for `multi`. It should be good now.

badboy:
Checking

lib/redis/pipeline.rb
```diff
@@ -122,7 +122,7 @@ def _set(object)
     end
 
     def _command
-      @command
+      @command.map(&:to_s)
```
byroot: This is a bit clever I recognize, the idea it to cast all the commands to string eagerly so that `call_pipelined` is sure not to trigger any side effect.

davidcornu: @byroot doesn't restoring the original client solve this though?

byroot: I though so too but no. See how it was still failing: https://travis-ci.org/redis/redis-rb/jobs/50558668

```
test_pipeline_race_condition(TestPipeliningCommands):
IOError: closed stream
    /home/travis/build/redis/redis-rb/lib/redis/connection/ruby.rb:269:in `write'
    /home/travis/build/redis/redis-rb/lib/redis/connection/ruby.rb:269:in `write'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:257:in `write'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:236:in `io'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:255:in `write'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:214:in `process'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:208:in `each'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:208:in `process'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:353:in `ensure_connected'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:207:in `process'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:310:in `logging'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:206:in `process'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:177:in `call_pipelined'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:149:in `call_pipeline'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:279:in `with_reconnect'
    /home/travis/build/redis/redis-rb/lib/redis/client.rb:147:in `call_pipeline'
    /home/travis/build/redis/redis-rb/lib/redis.rb:2140:in `pipelined'
    /home/travis/build/redis/redis-rb/lib/redis.rb:37:in `synchronize'
    /home/travis/.rvm/rubies/ruby-1.8.7-p374/lib/ruby/1.8/monitor.rb:242:in `mon_synchronize'
    /home/travis/build/redis/redis-rb/lib/redis.rb:37:in `synchronize'
    /home/travis/build/redis/redis-rb/lib/redis.rb:2132:in `pipelined'
    /home/travis/build/redis/redis-rb/test/pipelining_commands_test.rb:255:in `test_pipeline_race_condition'
    /home/travis/build/redis/redis-rb/test/helper.rb:70:in `run'
```

So restoring the initial client does solve the issue with the command being enqueued in the pipeline, but it was still writing the command in the connection at the wrong time, messing with the currently executed pipeline.

davidcornu: O_o

djanowski:
Thank you for your pull request and apologies for not looking at this before.

The fix is clever, the test looks good. Still, I'm not too eager to merge this. Pipelining is used for performance reasons, by issuing lots of commands, and mapping all commands adds overhead. And we'd be doing that just to fix the case where calling `#to_s` causes a side-effect hitting Redis. As you say, "a huge corner case".

By the way, I'm not even sure this change is good for the corner case you encountered. If we map the commands triggering the side effects before the pipeline, you may not notice that you're sending `n` commands to Redis, not pipelined, before sending your `n` pipelined commands. The fact that this broke probably lead you to a better solution to your problem.

byroot:
I understand, and thanks for looking into it.

One concern I have though, is that we were very lucky it actually raised. HGETALL is pretty much the only command that doesn't deal well with `nil`.

This race condition could have silently returned corrupted data. Would you be open to an alternative fix that would simply ensure there is as many responses than there was requests and explictly raise?

djanowski:
Either that, or making sure that while we're writing commands nothing else gets written. Let's explore what's cleaner.

PR #512, opened 2015-02-18
Prevent Client#options from crashing when it contains objects that can't be dumped

byroot:
If the options contains an object that can't be dumped e.g. a File, then the `Marshal.load(Marshal.dump(object))` trick to deep dup will fail.

Since Redis takes a logger as options it is relatively common for the options to contains a file descriptor or sockets.

The `_deep_dup` implementation is heavily inspired from `ActiveSupport`'s.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -23,7 +23,7 @@ class Redis
     }
 
     def options
-      Marshal.load(Marshal.dump(@options))
+      _deep_dup(@options)
     end
 
     def scheme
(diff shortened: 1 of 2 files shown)


djanowski:
Thank you for your pull request and apologies for not looking at this before.

Instead of adding more code to deal with cloning options, I'd like to revisit the underlying cause of why we need to do that.

byroot:
You suggest not to dup at all?

djanowski:
If possible, of course :)

byroot:
Closing in favor of https://github.com/redis/redis-rb/pull/567

PR #517, opened 2015-02-26
Adding warning about redis-sentinel gem on README

Draiken:
We've had some problems after upgrading the redis gem while still using the redis-sentinel. We only had errors in production when the redis client was incorrectly attempting to connect to 127.0.0.1.

This warning should help others avoid this in the future


djanowski:
Thank you for your PR.

Unfortunately, the README can't become a log of all the quirks anybody ever found (because that would hurt the main users of the README -- newcomers).

Again, I appreciate that you took the time to add this note.

PR #521, opened 2015-03-30
Support for URLs/Sentinels with underscores in name

msaffitz:
The built-in URI class for ruby doesn't properly handle hostnames with underscores in them.  See bug at:

https://bugs.ruby-lang.org/issues/8241

This updates to use the addressable gem, which provides better support for current URI standards.

Sentinel tests have also been updated to make sure of an underscore for validation.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -385,9 +385,9 @@ class Redis
 
       # Override defaults from URL if given
       if url
-        require "uri"
+        require "addressable/uri"
 
-        uri = URI(url)
+        uri = Addressable::URI.parse(url)
 
         if uri.scheme == "unix"
           defaults[:path]   = uri.path
--- a/redis.gemspec
+++ b/redis.gemspec
@@ -39,5 +39,6 @@ Gem::Specification.new do |s|
   s.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
   s.executables   = `git ls-files -- bin/*`.split("\n").map{ |f| File.basename(f) }
 
+  s.add_dependency('addressable')
   s.add_development_dependency("rake")
 end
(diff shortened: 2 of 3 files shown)


msaffitz:
Looks like that failure was transient port conflicts :(

jemc:
I'm running into this issue as well, and want to see a fix, but I'd really prefer one that doesn't add an additional gem dependency.

jemc:
I should note that this is fixed in the builtin URI library on more recent versions of MRI - it's only a problem for users who cannot upgrade to 2.2.2.

msaffitz:
If I get a chance and there was interest from the maintainers, I could look into changing this to only pull the dependency in for MRI < 2.2.x (?).  @djanowski Thoughts on whether this would be accepted w/ or w/o that change?

yaauie:
Perhaps a better place for this would be the backports gem? Since the feature is present in stdlib >= 2.2, providing the backports there would make it opt-in, without explicitly adding dependencies to the many, many projects that rely on `URI` from stdlib; users who need stdlib 2.2 features could require those features piecemeal.

Another option could be to use `Addressible` iff it is present (although that would double the work needed to validate every commit in Travis CI).

msaffitz:
@yaauie That's an interesting idea; I wasn't aware of the backports gem.  That would additional address the issue for sidekiq, since mperham was equally unexcited about taking in the dependency :).  This isn't super high priority for us at the moment, but if we need to touch the code again we'll look at moving it to backports.

djanowski:
Apologies for the delay in answering.

Yes, I agree with Mike, this is the first time I hear anybody uses underscores for host names.

I'm not changing code, let alone add a dependency, for some questionable decision by Chef developers.

If you need this so badly, you can:

- Submit a PR to Chef
- Upgrade Ruby
- Monkey patch `Kernel#URI` ¯\_(ツ)_/¯ 

Thank you.

Issue #437, opened 2014-05-22
Provide separate connection timeout and response timeout options


mfischer-zd:
Currently the `:timeout` option to the Redis constructor applies a latency ceiling both to the initial connection to the Redis server and to any responses to requests.  This means you can't easily fail a connection within 1ms but fail a response (assuming the client successfully connected) that takes longer than 300ms.

Also, the error message is always `Redis::TimeoutError: Connection timed out`, even if the client successfully connected and the response timed out.  This is highly misleading.

plashchynski:
If there is `:connect_timeout`, so not obvious the meaning of `:timeout`, isn't it?

mperham:
I'd love to see:

1. timeout deprecated and replaced with `read_timeout`, the same term as used by Net::HTTP.
2. the README updated to document the timeout options.

I will do this soon unless a committer objects.

byroot:
This has been solved for a while.

PR #528, opened 2015-05-07
Rename timeout option to read_timeout

mperham:
Associated with issue #437, this renames the `timeout` option and adds to README so the options are documented.  Existing usage of `timeout` is supported for backwards compatibility.

There's a few test failures, I'm unclear how to fix them.


badboy:
#514 has some initial documentation for the options, we definitely need full documentation of all parameters.

PR #529, opened 2015-05-08
Check redis connection without Redis::CannotConnectError

letz:
I've notice an odd behaviour, if you kill redis-server the `connected?` method will continue to respond `true` if you don't make any operation. Example:

```ruby
[START redis-server]
redis = Redis.new
redis.connected? # => false (as expected lazy connection)
redis.get 'some_key'
redis.connected? # => true (as expected)
[KILL redis-server]
redis.connected? # => true (BUT isn't connected)
redis.get 'some_key' # => raise Redis::CannotConnectError
redis.connected? # => false (as expected)
```

Signed-off-by: Ricardo Leitão <rleitao@onliquid.com>

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -54,6 +54,14 @@ class Redis
     @original_client.connected?
   end
 
+  # Test whether or not the redis server is up
+  def connection_alive?
+    ping
+    connected?
+  rescue Redis::CannotConnectError
+    false
+  end
+
   # Disconnect the client as quickly and silently as possible.
   def disconnect!
     @original_client.disconnect
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -56,7 +56,7 @@ class Redis
 
   # Test whether or not the redis server is up
   def connection_alive?
-    ping
+    without_reconnect { ping }
     connected?
   rescue Redis::CannotConnectError
     false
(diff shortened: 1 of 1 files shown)


yaauie:
Since sending a ping will attempt to reconnect if the connection is in fact severed, it might be a good idea to wrap it in a `without_reconnect` so as not to introduce connection churn.

letz:
@yaauie like this? 

Any ideia why the spec is failing?

badboy:
Restarted the failing tests, some shortcomings in the test suite make this sometimes fail.

badboy:
There you go

natematykiewicz:
Can this be merged? This would be very helpful for me.

djanowski:
You mean kill as in SIGKILL, right? Otherwise, Redis would close its sockets gracefully and the connected status should be correct.

letz:
@djanowski yes, with a SIGKILL

djanowski:
Redis doesn't usually just die like that, and sending SIGKILL is done under very rare circumstances, so I'm inclined not to merge this. People needing this can write the small snippet to probe the connection themselves.

letz:
Ok, no problem.

djanowski:
Thank you!

PR #540, opened 2015-07-14
docs: add installation step.

duykhoa:


yaauie:
We'd rather not tie our users to bundler specifically (believe it or not, some of us don't use it on all projects :smile:), so how about something more flexible:

> 
> To install redis-rb, run the following command:
> ~~~ sh
> gem install redis
> ~~~
> Or, if you're using bundler, add `gem 'redis', '~>3.2'` to your `Gemfile` and run:
> ~~~ sh
> bundle install
> ~~~

duykhoa:
Hehe, didn't know about this.

I think your instruction is better than mine. Let me update the PR.
:+1:

djanowski:
Thank you.

PR #556, opened 2015-10-26
Add support for running sentinel API commands

dsolsona:
### What?
This pull request adds support for the redis sentinel API. The redis-rb gem already has support for redis sentinel but it can't be used (afaik) to connect to a redis sentinel and to run any of the sentinel API commands like `sentinel masters`, `sentinel master`, `sentinel monitor`, etc...

### Why? 
I was trying to build some tooling around the redis sentinel API to add new redis instances to a redis sentinel servers and I wanted to use this redis gem. 

### How?
Added a new def for the sentinel command and all it's subcommands. Some subcommands return a single array that gets transformed to a hash, some other subcommands return an array of arrays which get transformed to an array of hashes for ease of use.

I also had to change the redis_mock to be able to handle replies consisting of multiple arrays like the reply you get when you run `sentinel masters` which returns an array with all the redis instances the sentinel is monitoring.

I've also added some basic tests that cover some sentinel commands.

This is my first contribution to the project, if there's anything missing or something is terribly wrong just let me know, I'm just starting with ruby so please forgive my terrible code :)

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -2539,6 +2539,34 @@ class Redis
     end
   end
 
+  # Interact with the sentinel command (masters, master, slaves, failover)
+  #
+  # @param [String] subcommand e.g. `masters`, `master`, `slaves`
+  # @param [Array<String>] args depends on subcommand
+  # @return [Array<String>, Hash<String, String>, String] depends on subcommand
+  def sentinel(subcommand, *args)
+    subcommand = subcommand.to_s.downcase
+    synchronize do |client|
+      if subcommand == 'master'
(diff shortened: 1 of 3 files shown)


badboy:
Build failures are still the same old float rounding errors -_-

dsolsona:
I can see the same tests failing on my laptop when running them against master.

djanowski:
It's a great first contribution. Thank you for detailing the what, why and how. I'll merge soon.

djanowski:
This is available in 3.2.2.

davewongillies:
@dsolsona you broke our shit :stuck_out_tongue_closed_eyes:

PR #559, opened 2015-11-02
Support sentinels as a string

quixoten:
This allows the `sentinels` option to be passed as a string as well as a
symbol.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -527,8 +527,8 @@ class Redis
         def sentinel_detect
           @sentinels.each do |sentinel|
             client = Client.new(@options.merge({
-              :host => sentinel[:host],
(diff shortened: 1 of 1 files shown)


quixoten:
Anything I can do to get a review of this? Looks like all the tests are passing after merging the latest master.

fjyaniez:
Hi @quixoten, 

Was checking on my PR status and saw this one you created weeks before mine to solve the same issue. 

I also added support for string params recursively in the sentinels settings and not only in the sentinels key, something similar to `HashWithIndifferentAccess` as I saw they do the same for the rest of params:

https://github.com/redis/redis-rb/pull/571

If you think it's not necessary, I'll close mine.

Regards,

PR #560, opened 2015-11-03
Use easily representable floats to prevent formatting issues

badboy:


jcmfernandes:
This solves the troubles with redis unstable.

Are redis devs aware of this problem? It doesn't seem to be related with this gem.

badboy:
They are and it's actually fixed now: https://github.com/antirez/redis/commit/71aa9b75f2c4dd6faa20a4849ff732536bdfceab

So we can keep the old code.

jcmfernandes:
Thanks @badboy! :+1:

PR #561, opened 2015-11-04
3.3

djanowski:


README.md
```diff
@@ -223,6 +223,38 @@ end
 See lib/redis/errors.rb for information about what exceptions are possible.
 
 
+## Timeouts
+
+The client allows you to configure connect, read, and write timeouts.
+Passing a single `timeout` option will set all three values:
```
badboy: What units are they in? Probably seconds, but this should just be documented here.

djanowski:
@igrigorik We got a test hanging in the Synchrony adapter, looks like we're not setting any read timeout. What do you think about this patch? https://gist.github.com/badboy/9ce54ef1c87523649b4b6eeb18040bf2

igrigorik:
@djanowski the patch looks reasonable, but does that address the actual problem? As in, if we're patching the hanging request with a timeout.. that doesn't seem like the right solution?

badboy:
We're adding the timeout only if it is set. The hanging test was a test that set a explicit timeout. All other tests are still green with the above patch.

igrigorik:
Ah, in which case. :+1: 

@djanowski open a PR?

djanowski:
@igrigorik Thank you for looking at this.  Sorry, I didn't understand, PR where?

We'll probably patch it here and merge this whole thing as the 3.3 release.

igrigorik:
nm, you're shipping synchrony adapter as part of the gem. I was thinking you wanted to add this to em-synchrony repo. Apologies about the confusion.. looks good and make it so! :)

djanowski:
@igrigorik BTW our plan is to move the Synchrony adapter to a separate gem. Feel free to jump to `#redis-rb` on Freenode or follow back on Twitter if you want to discuss. Thanks!

djanowski:
Our build found a bug in Redis unstable: https://github.com/antirez/redis/pull/3174

PR #567, opened 2015-11-16
Remove the deep dup in Client#options

byroot:
Replaces https://github.com/redis/redis-rb/pull/512

`redis-rb` itself doesn't use that accessor at all. And the deep dup was simply a precaution.

IMO it's a worthless one.

@djanowski for review please.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -22,9 +22,7 @@ class Redis
       :inherit_socket => false
     }
 
-    def options
-      Marshal.load(Marshal.dump(@options))
-    end
+    attr_reader :options
 
     def scheme
       @options[:scheme]
--- a/test/internals_test.rb
+++ b/test/internals_test.rb
@@ -349,19 +349,6 @@ class TestInternals < Test::Unit::TestCase
     assert_equal "foo", redis.client.options[:scheme]
   end
 
-  def test_does_not_change_self_client_options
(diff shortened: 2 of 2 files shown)


fw42:
@badboy can you take a look here please?

badboy:
Is this currently causing any problems in libs/apps depending on redis-rb?

cc @djanowski

fw42:
It's causing some problems for us in Shopify (see https://github.com/redis/redis-rb/pull/512 for details). We've been running with this patch (in our fork) in production for months and haven't seen any problems.

djanowski:
I favor removing this cloning of objects. But this was added for a reason. Maybe a no longer valid one. So this PR would be much easier to review if it tracked down the issues and PRs that led to this behavior and argue why it is no longer needed.

fw42:
Looks like it was added in https://github.com/redis/redis-rb/pull/386, which also added two tests (one of which is being removed by this PR here).

fw42:
@Napolskih, maybe you can comment on whether or not this change here is okay?

fw42:
@pietern commented on that PR, saying:
> Looks good to me. One minor thing: can you make the reader method for the options hash create a deep copy of the options hash before returning it? I don't want anything else being able to modify the options of a live client.

fw42:
Personally, I feel like this seems safe to remove (and the new behaviour is actually what I would expect).

rafaelfranca:
Ping

djanowski:
This is implemented in 4.0.

PR #568, opened 2015-11-16
Explicitly raise InvalidResponse if commands and responses counts don't match

byroot:
As discussed in https://github.com/redis/redis-rb/pull/509

The idea here is to fail loudly if such unlikely race condition happened.

@djanowski is this more acceptable to you?

### DIFF
--- a/lib/redis/errors.rb
+++ b/lib/redis/errors.rb
@@ -5,8 +5,8 @@ class Redis
 
   # Raised by the connection when a protocol error occurs.
   class ProtocolError < BaseError
-    def initialize(reply_type)
-      super(<<-EOS.gsub(/(?:^|\n)\s*/, " "))
+    def self.invalid_reply_type(reply_type)
+      new(<<-EOS.gsub(/(?:^|\n)\s*/, " "))
         Got '#{reply_type}' as initial reply byte.
         If you're in a forking environment, such as Unicorn, you need to
         connect to Redis after forking.
--- a/lib/redis/pipeline.rb
+++ b/lib/redis/pipeline.rb
@@ -56,6 +56,10 @@ class Redis
     end
 
     def finish(replies, &blk)
+      unless replies.size == futures.size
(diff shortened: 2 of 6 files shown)


lib/redis/errors.rb
```diff
@@ -14,6 +14,10 @@ def initialize(reply_type)
     end
   end
 
+  # Raised when the response from the server is unparsable
+  class InvalidResponse < BaseError
+  end
+
```
djanowski: `ProtocolError`?

lib/redis/pipeline.rb
```diff
@@ -58,6 +58,10 @@ def without_reconnect(&blk)
     end
 
     def finish(replies, &blk)
+      unless replies.size == futures.size
+        raise InvalidResponse, "#{futures.size} commands were enqueued, but only #{replies.size} responses were recieved"
+      end
```
djanowski: Typo.

lib/redis/errors.rb
byroot: I wanted to use ProtocolError initially, but it's constructor is customized. I can refactor it though.

lib/redis/pipeline.rb
```diff
@@ -71,7 +75,7 @@ def finish(replies, &blk)
 
     class Multi < self
       def finish(replies)
-        exec = replies.last
+        exec = replies.pop
```
djanowski: If possible, let's not mutate an argument.

byroot: I can copy the array instead if you think it's best.

byroot:
I fixed the typo, and refactored `ProtocolError` so I can reuse it.

I honestly think the `pop` is simpler and better than the copy. I mean I get your point about not mutating arguments, but in this case it's a method responsible for parsing a response. So I really don't expect any further usage of that object.

That being said I don't think super strongly about it. So I can totally do a `slice` instead. Just let me know.

byroot:
We no longer need this because pipelines are serialized eagerly.

PR #569, opened 2015-11-22
Reduce `Proc` object allocations.

marshall-lee:
It's also  a kind of refactoring: it removes private methods that actually are not methods of `Redis` instance since they don't depend on its state.


djanowski:
Been wanting to make this change for a long time. Will merge soon.

marshall-lee:
Let's merge it!

davydovanton:
:+1:

marshall-lee:
@djanowski  Bump!

Issue #570, opened 2015-11-23
sentinels option must be symbolized


fjyaniez:
Hi,

Loading the configuration from a YML file I noticed the following inconsistent behaviour while initializing `Redis`, depending on passing the sentinels parameters as a key or as a symbol:

``` ruby
irb(main):004:0> redis = Redis.new("url"=>"redis://mymaster", "sentinels"=>[{"host"=>"slave1", "port"=>6381}, {"host"=>"slave2", "port"=>6381}])
=> #<Redis client v3.2.1 for redis://mymaster:6379/0>
irb(main):005:0> redis.set 'key', 'key'
SocketError: getaddrinfo: Name or service not known
```

``` ruby
irb(main):006:0> redis = Redis.new("url"=>"redis://mymaster", :sentinels=>[{"host"=>"slave1", "port"=>6381}, {"host"=>"slave2", "port"=>6381}])
=> #<Redis client v3.2.1 for redis://mymaster:6379/0>
irb(main):007:0> redis.set 'key', 'key'
TypeError: can't convert nil into Integer
```

``` ruby
irb(main):010:0> redis = Redis.new("url"=>"redis://mymaster", :sentinels=>[{"host"=>"slave1", :port=>6381}, {"host"=>"slave2", :port=>6381}])
=> #<Redis client v3.2.1 for redis://mymaster:6379/0>
irb(main):011:0> redis.set 'key', 'key'
SocketError: getaddrinfo: Name or service not known
```

``` ruby
irb(main):012:0> redis = Redis.new("url"=>"redis://mymaster", :sentinels=>[{:host=>"slave1", :port=>6381}, {:host=>"slave2", :port=>6381}])
=> #<Redis client v3.2.1 for redis://mymaster:6379/0>
irb(main):013:0> redis.set 'key', 'key'
=> "OK"
irb(main):014:0>
```

``` ruby
irb(main):012:0> redis = Redis.new("url"=>"redis://mymaster", "sentinels"=>[{:host=>"slave1", :port=>6381}, {:host=>"slave2", :port=>6381}])
=> #<Redis client v3.2.1 for redis://mymaster:6379/0>
irb(main):013:0> redis.set 'key', 'key'
SocketError: getaddrinfo: Name or service not known
```

Would be safe to use an equivalent of Rail's `deep_symblize_keys` method on `Redis::Client`'s `initialize` method to recursively symbolize the `options` parameter?

[...]

PR #571, opened 2015-11-23
Symbolize sentinels option to align with other options

fjyaniez:
Fixes: #570 

If we can indifferently use `"url"` or `:url` as an option, we should be able to do the same with `:sentinels` option.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -87,7 +87,7 @@ class Redis
 
       @pending_reads = 0
 
-      if options.include?(:sentinels) && options[:sentinels].any?
+      if @options.include?(:sentinels) && @options[:sentinels].any?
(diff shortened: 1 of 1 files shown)


lib/redis/client.rb
```diff
@@ -86,7 +87,7 @@ def initialize(options = {})
 
       @pending_reads = 0
 
-      if options.include?(:sentinels)
+      if options.include?(:sentinels) && options[:sentinels].any?
```
quixoten: I remember having issues before i changes this to `@options`

lucaong:
Duplicate of #599 (both PRs are open since years), but #599 also take care of the `:role` option, and avoids symbolizing arbitrary keys.

PR #577, opened 2015-12-10
Fix various test suite warnings relating to variable ambiguity and unused variables

RochesterinNYC:
The following warnings appear between the starting of the Redis server and the loading of the test suite when running `rake`:

```
redis-rb/test/lint/sorted_sets.rb:42: warning: ambiguous first argument; put parentheses or a space even after `-' operator
redis-rb/test/lint/sorted_sets.rb:43: warning: ambiguous first argument; put parentheses or a space even after `+' operator
redis-rb/test/lint/sorted_sets.rb:104: warning: ambiguous first argument; put parentheses or a space even after `-' operator
redis-rb/test/lint/sorted_sets.rb:105: warning: ambiguous first argument; put parentheses or a space even after `+' operator
redis-rb/test/distributed_internals_test.rb:75: warning: assigned but unused variable - redis
```

This commit fixes those warnings and makes the test suite output a little cleaner.

PR #576, opened 2015-12-10
Create Redis::Distributed#remove_node

RochesterinNYC:
- Create `remove_node` method for `Redis::Distributed` which utilizes `Redis::HashRing#remove_node`, which was previously not utilized. 
- Refactor `add_node` and `remove_node` in `Redis::Distributed` by pulling out some duplicated code

PR #579, opened 2015-12-10
Fix CircleCI build inability to connect to Redis

### LINKED ISSUE #578: CircleCI breaking with `Cannot connect to Redis.` on PR builds
CircleCI for this repo is breaking on builds for pull requests that have code changes that would not cause the `Cannot connect to Redis.` error currently on CircleCI builds.

According to the CircleCI documentation at https://circleci.com/docs/environment, they set up redis automatically as a service.

Hence, it's possible that our test suite tries to connect to Redis on the wrong port? As 6379 is default, I imagine that the CircleCI's Redis instance is connected on that port? (Not sure if there's a way to configure that). Our test suite tries to connect on 6381 so this could result in the `Cannot connect to Redis.` error.

RochesterinNYC:
Attempts to fix #578

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,5 @@
+##Start Redis on 6381
+dependencies:
+  pre:
+    - rake start
+    - sleep 5


RochesterinNYC:
This commit would fix CircleCI builds by ensuring that `rake start` is ran as a pre-dependency for running the test suite on their VM. It would slow the CircleCI builds a bit since it seems to re-install Redis on the VM, but it is still much faster than the Travis CI build so this is negligible. 

The following error seems to occur in running the test suite in the CircleCI VM:

```
Error running mock server: Connection reset by peer
/home/ubuntu/redis-rb/test/support/redis_mock.rb:76:in `gets'
/home/ubuntu/redis-rb/test/support/redis_mock.rb:76:in `block in start'
/home/ubuntu/redis-rb/test/support/redis_mock.rb:28:in `block in run'
/home/ubuntu/redis-rb/test/support/redis_mock.rb:24:in `loop'
/home/ubuntu/redis-rb/test/support/redis_mock.rb:24:in `run'
/home/ubuntu/redis-rb/test/support/redis_mock.rb:15:in `block in start'
```

However, it doesn't break the test suite and passes successfully. If this is merged, #576 and #577 can be rebased off of this and they also pass the CircleCI builds.

badboy:
Circle is deactivated again.

PR #580, opened 2015-12-31
Fixed variable name in the example usage

digininja:
I think the second variable should be key2 and not key1 but could be wrong.


badboy:
LGTM

PR #585, opened 2016-02-01
Always return integer from ZADD

pik:
Hello, I can't find an explanation in the commit-message for why this line is here - but I believe the current use of _boolify disagrees with the Redis documentation: http://redis.io/commands/zadd

For example: `cli.zadd('foo', [2, 'bar'], [3, 'delta'])` will return true or false, even though 0, 1, 2 are all possible return values.

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -1485,8 +1485,8 @@ class Redis
         # Variadic: return float if INCR, integer if !INCR
         client.call([:zadd, key] + zadd_options + args[0], &(incr ? _floatify : _identity))
       elsif args.size == 2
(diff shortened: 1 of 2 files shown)


badboy:
Possibly a relict of old times. In Redis <2.4 it was only possible to add a single item

badboy:
This is a breaking change. /cc @djanowski

djanowski:
Thanks @pik for your contribution. I agree the code would be simpler with your proposed changes.

I'd make the change if 0 in Ruby were falsy. Breaking this can lead to all sorts of subtle bugs, so we'll leave it as is for now.

PR #586, opened 2016-02-08
Updated README.md to indicate how to use renamed commands

danlo:
Renamed commands are a security feature (accident preventive) in redis: http://redis.io/topics/security  .  I renamed my command and then wasn't able to execute it!  This adds to the code how to enable these renamed commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,25 @@ end
 # => ["OK", 1]
 ```
 
+### Redis: rename-command
+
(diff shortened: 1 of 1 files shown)


danlo:
Humm, I didn't change any code.  Only documentation.

badboy:
Not sure if the README is the right place for this.

vassilevsky:
README is the right place for such information. I needed a solution for renamed commands. I opened the project page and searched for "rename" on the page. This is a natural thing to do.

However, I'm not sure if the solution is good for everybody. It doesn't cover all cases. For example, I need to issue a renamed command when my app sends the original command.

rutaka-n:
Actually, instead redefining  methods for Redis class you can use command_map for redefine command for particular instance of redis. 
For example: 
```
Redis.new.client.command_map[:command_name] = :new_command_name
```

PR #588, opened 2016-02-27
Add the :nearest_slave role for Sentinel mode

cheald:
This will cause the Sentinel client, upon connection, to measure roundtrip latency to each slave and select the slave with the lowest latency. The intent for this is to enable sentinel-managed clusters of servers for which eventually-consistent reads are acceptable, but to maintain minimum latencies between any individual client-slave pair.

The case I did this for is is shared web application caching across multiple datacenters, where you would not want Redis to read from a slave in another datacenter, but you want to be able to share the cache across the datacenter.

(If this would be more appropriate as a separate gem or something, I'm happy to do that, but there didn't seem to be enough exposed API to accomplish it without a PR.)


cheald:
The 1.8.7 failures appear to be present in master, as well, running them locally with `BUNDLE_GEMFILE=$PWD/.travis/Gemfile rake`. Running just `rake` passes the suite as expected. I'm reasonably confident that these changes haven't caused that break.

cheald:
It looks like the issue is the difference between test_unit 3.1.5 and 3.1.7. Locking test_unit to 3.1.5 causes the suite to pass.

badboy:
Would you mind submitting the version pinning separately? Then we can already merge that regardless of whether this new feature is accepted.

cheald:
Certainly! You'll find it at #593

ggmichaelgo:
@cheald Can you update your PR please?
This PR would be amazing to be merged....

One of our redis replica is in US Eastern, and one of our Rails instance in US Western region...

The delay between two instances are 60ms~, but ideally our Rails application should only connect to its region's redis

cheald:
Looks like a simple merge. I've rebased it to master. I don't currently have a production use case for this, but let me know if it works for you.

byroot:
Rubocop has a few complaints. 

The code looks sane to me, but @supercaracal if you have time and are willing to give this a second look, it would be much appreciated.

cheald:
I'll see what I can do to make Rubocop happy.

PR #590, opened 2016-03-02
Change for frozen string literal

twalpole:
Minor change to work with frozen string literals


badboy:
CI failure unrelated. Gonna check that in the coming days.

PR #591, opened 2016-03-03
Add option to use a different Connector.

tgxworld:


badboy:
What would be the use case of this?

tgxworld:
I'm trying to run a custom connector without having to monkey patch. Basically have the ability to subclass `Redis::Client::Connector` easily.

tgxworld:
@badboy I would like to clarify if the request is reasonable :)

SamSaffron:
@badboy any news on this? need to know if its acceptable or not, I hate carrying freedom patches around for Discourse :)

SamSaffron:
see: 

https://github.com/discourse/discourse/pull/4056

SamSaffron:
@badboy any news on this issue?

tgxworld:
ping @djanowski Any thoughts about this?

tgxworld:
@badboy  @djanowski Don't mean to ping multiple times but we would love to get some feedback on this.

tgxworld:
@byroot Was wondering if you might have any thoughts on this.

lib/redis/client.rb
```diff
@@ -79,11 +79,14 @@ def initialize(options = {})
...
 
-      if options.include?(:sentinels)
-        @connector = Connector::Sentinel.new(@options)
-      else
-        @connector = Connector.new(@options)
-      end
+      @connector =
+        if options.include?(:sentinels)
+          Connector::Sentinel.new(@options)
+        elsif options.include?(:connector) && options[:connector].is_a?(Class)
```
byroot: I wouldn't check for `is_a?(Class)`, `respond_to?(:new)` would be better as it would allow for facades etc.

tgxworld:
@byroot Thank you for reviewing. I've updated the PR as per your comments and added tests as well.

tgxworld:
@byroot Thank you for reviewing. Just curious, is Shopify/Rails taking up ownership of this gem now?

byroot:
A few people from Shopify have been added as collaborators on the repo indeed. We should also get Rubygems access some day so we can release new versions as well.

tgxworld:
Awesome :+1: Thanks for reviewing once again!

PR #595, opened 2016-03-04
Add Redis#in_multi?

oggy:
It can be helpful to know when we're in a multi since we can't nest them.


oggy:
Oh, looks like you can nest calls to `multi`. Ignore me.

PR #599, opened 2016-03-16
Accept sentinel options even with string keys

lucaong:
Before, the sentinel options were assumed to have symbol keys, so string keys
were ignored. This is surprising, as other Redis options can be indifferently
passed as symbols or strings. Now string keys are allowed too.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -22,7 +22,9 @@ class Redis
       :reconnect_attempts => 1,
       :reconnect_delay => 0,
       :reconnect_delay_max => 0.5,
-      :inherit_socket => false
+      :inherit_socket => false,
+      :sentinels => nil,
+      :role => nil
     }
 
     attr_reader :options
@@ -89,7 +91,7 @@ class Redis
       @pending_reads = 0
 
       @connector =
-        if options.include?(:sentinels)
+        if !@options[:sentinels].nil?
           Connector::Sentinel.new(@options)
         elsif options.include?(:connector) && options[:connector].respond_to?(:new)
           options.delete(:connector).new(@options)
@@ -539,7 +541,7 @@ class Redis
           @options[:db] = DEFAULTS.fetch(:db)
 
           @sentinels = @options.delete(:sentinels).dup
-          @role = @options.fetch(:role, "master").to_s
+          @role = (@options[:role] || "master").to_s
           @master = @options[:host]
         end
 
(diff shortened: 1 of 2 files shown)


yagilm:
+1

lucaong:
Travis specs fail on Ruby `1.8.7` due to:

```
Gem::InstallError: rake requires Ruby version >= 1.9.3.
An error occurred while installing rake (11.1.1), and Bundler cannot continue.
Make sure that `gem install rake -v '11.1.1'` succeeds before bundling.
```

This is not related to the code changes though. Tests pass locally, but I would love to see the build pass on different Ruby versions on Travis. Could this be an issue with the version of `test_unit`?

badboy:
More the fact that we don't restrict the rake version and therefore the latest one is pulled which is now incompatible with Ruby 1.8.7

We will pull the plug on 1.8.7 eventually.

lucaong:
I see ;)

lucaong:
It is possible to fix this by adding, in the gemspec, something like this:

```ruby
if RUBY_VERSION == "1.8.7"
  s.add_development_dependency "rake", "~> 10.0"
else
  s.add_development_dependency "rake"
end
```

That would make the build run. Should I do that as part of this PR, or separately, or skip it completely?

badboy:
I don't like the conditional there. We don't depend on any newer features, so we can stick with 10.0 for all Ruby versions for now (this is only a development dependency and should not cause trouble to other gems)

lucaong:
OK, tests are passing now. I also had to make sure a recent version of bundler is used for CI, to fix failures on jRuby head. And close/reopen PR to force Travis to rebuild.

keeganlow:
just helped someone who got burned by this issue - seems like the concerns about rake version mentioned above should be addressed now, since https://github.com/redis/redis-rb/pull/601 is now in master.

Would this PR be ready to merge if we remove the parts that deal with the rake version? Happy to submit a modified PR myself if that helps.

alexpapworth:
Any chance we can get this PR finished off and approved?

lucaong:
As far as I know, this is finished (tests were passing), although I sent it 4 years ago so things might have changed?

lucaong:
I took care of rebasing on the current master and fixing the conflicts, as a few things changed since this PR was open.

alexpapworth:
Awesome! Can we get a member of the Redis team to review?

cc: @byroot

PR #600, opened 2016-03-17
fix redis connection option example

passion8:


badboy:
The usual URI syntax is: `protocol://<user>:<password>@<host>:<port>`, and as Redis does not have a user but only a password prefixing it with `:` is the right thing to do.

Issue #524, opened 2015-04-08
Connections are not garbage collected


robertgrimm:
I noticed the Redis class makes no attempt to automatically disconnect/quit its client connection when the Redis object is garbage collected and instead appears to force the application to manage that.  Is this by design?  While applications can of course keep track of this themselves and call Redis#quit when appropriate, it appears that the Redis class can be modified to do that automatically so applications don't have to worry about that and thus not leak Redis connections.

For example:

``` ruby
class Redis
  def initialize(options={})
     # existing initialize
     # not sure if finalize should be called with @original_client or just client.
     ObjectSpace.define_finalizer(self, self.class.finalize(@original_client))
  end

  def self.finalize(client)
    proc { client.disconnect }
  end

  # rest of the class definition
end
```

This can be tested by doing something like

``` ruby
redis = Redis.new
# Would be nice if the gem natively supported "client setname"...
redis.client.call(["client", "setname", "test"])
# Run "redis-cli client list | grep test" or something similar to see the connection
redis = nil
GC.start
# Run "redis-cli client list | grep test" again and see the connection is no longer there
```

Another option would be add something like this to the Client class instead.  Thoughts?

fotinakis:
May or not be related, but I'm currently seeing a lot of Ruby segfaults like this:

ruby/2.2.0/gems/redis-3.2.1/lib/redis/connection/ruby.rb:49: [BUG] object allocation during garbage collection phase

fotinakis:
I definitely agree with this issue — I've found that the redis-rb library can easily consume the limited number of connections I'm paying for on a third-party redis service, and so I have to do a lot of nasty application and system configuration to aggressively cleanup the connections.

BBCMarkMcDonnell:
Wow, I'm surprised this hasn't gotten more traction/feedback from the Redis gem authors?

I've been load testing a collection of microservices hosted on AWS and discovered that our connections weren't ever dropping down once opened. I've only just started looking into why, and this issue popped up first on Google :-(

I'll have to take a look at @hale's commit/fix and fork my own version.

badboy:
Hey @BBCMarkMcDonnell.
We maitain redis-rb in our free time and may not face the same problems as heavy users. Looks like this issue just fell through. @hale's fix was never submitted as a PR, so we didn't really see that either.
I now pulled it in (#603) and will wait for the CI to report back.

Thanks for bringing this up again.

Integralist:
[...]

PR #603, opened 2016-04-06
Ensure connections are closed when client is GC'd

badboy:
DO NOT MERGE!

See https://github.com/redis/redis-rb/issues/524

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -24,6 +24,10 @@ class Redis
     @current = redis
   end
 
+  def self.finalize(client)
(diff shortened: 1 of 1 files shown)


badboy:
This _does_ close the connections, but it seems we're leaking internal objects somehow, otherwise the tcp connection would be closed on GC anyway. Investigating further.

kaspergrubbe:
@badboy Did your investigations yield anything?

badboy:
I don't think so. That was 4 years ago. I don't maintain this project anymore.

PR #609, opened 2016-04-14
Add Hash examples to Readme

fpaula:
I've been using Redis Hashes a lot in my Rails Application but I had to open the gem's source to check if it had the `hset`, `hmset` and `hget` commands. In fact, it has but It was not in the Readme.

To add this information in the Readme is the objective of this pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ use it for the next requests.
 
 ## Storing objects
 
-Redis only stores strings as values. If you want to store an object, you
+Redis is well known to store strings as values. If you want to store an object, you
 can use a serialization mechanism such as JSON:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


badboy:
We should probably just change the sentence and refer to the many different values Redis can now take. Listing each and every value type is a duplicated effort IMO.

fpaula:
Hi @badboy.

I think we should have at least a simple usage for the most common data types such as:

- String
- Hash
- List
- Set
- Ordered set

What do you think?

arturcp:
I liked this change. It is useful to know what we can do with the gem without having to dig into the code, at least with the most common data types. I vote 👍  to merge this PR.

PR #615, opened 2016-05-04
Write nonblock

badboy:
This should fix the latest leaks due to the way the timeout module spawns threads.

I used the following test to verify, but I'm not 100% sure it is reliable.
I'm still sure the code is actually correct ([tcp-timeout](https://github.com/lann/tcp-timeout-ruby) uses it as well). It's definitely better than using Timeout.

```ruby
driver(:ruby) do
  def test_write_timeout
    server = TCPServer.new("127.0.0.1", 0)
    port   = server.addr[1]

    # Hacky, but we need the buffer size
    val = TCPSocket.new("127.0.0.1", port).getsockopt(Socket::SOL_SOCKET, Socket::SO_SNDBUF).unpack("i")[0]

    assert_raise(Redis::TimeoutError) do
      Timeout.timeout(1) do
        redis = Redis.new(:port => port, :timeout => 5, :write_timeout => 0.1)
        redis.set("foo", "1" * val*2)
      end
    end
  end
end
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,27 +38,48 @@ matrix:
     # hiredis
     - rvm: jruby-18mode
       gemfile: .travis/Gemfile
-      env: conn=hiredis REDIS_BRANCH=2.8
+      env: conn=hiredis REDIS_BRANCH=3.0
+    - rvm: jruby-18mode
+      gemfile: .travis/Gemfile
+      env: conn=hiredis REDIS_BRANCH=3.2
     - rvm: jruby-19mode
       gemfile: .travis/Gemfile
-      env: conn=hiredis REDIS_BRANCH=2.8
+      env: conn=hiredis REDIS_BRANCH=3.0
+    - rvm: jruby-19mode
+      gemfile: .travis/Gemfile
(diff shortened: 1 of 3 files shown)


badboy:
I added the test code, so it at least runs on Travis once. I can still take it out before merging if it is too unrealiable.

djanowski:
Great! I'll try to review tomorrow, along with checking of benchmarks.

badboy:
Test failures due to problems in Ruby 1.8.7 and flaky tests in jruby (plus JRuby in 1.8 mode).
I'm in favor of dropping 1.8 support immediately, though I understand that we might want to still make the next patch release work on 1.8.7 (that's why I added a monkey-patch to have a somewhat similar `byteslice` in Ruby <1.9

badboy:
Can we pull this in?

badboy:
This change seems to not work on jruby (using the above test). Everything else is green though.
I'm disabling the test for now.

Old results: https://travis-ci.org/redis/redis-rb/builds/143895633

badboy:
cc @djanowski

PR #616, opened 2016-05-04
Pin em-synchrony version

badboy:
Seems 1.0.5 does not work with Ruby 1.8.7 anymore.


badboy:
Somehow this didn't work as expected.

PR #617, opened 2016-05-12
Add count param to spop interface

zebogen:
The count param was added to SPOP in Redis 3.2.


djanowski:
Implemented by https://github.com/redis/redis-rb/pull/628. Thanks.

PR #623, opened 2016-06-12
Keeps default url even with empty url string

diogobenica:
When passing a empty url string, it breaks with `ArgumentError: invalid uri scheme ''`. In my case, Sidekiq was passing a empty string, and I took a time to debug to redis-rb to find the problem, since it was a indirect dependency.

Hope it is ok!

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -398,7 +398,7 @@ class Redis
       url = options[:url] || defaults[:url]
 
       # Override defaults from URL if given
-      if url
+      if url && !url.empty?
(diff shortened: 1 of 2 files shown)


diogobenica:
Hey, I think some tests are broken in CI but has not relation with my changes. There is a way to run again?

diogobenica:
I know it's been a time. This PR makes sense? Don't know who to talk to @badboy

badboy:
If you are explicitely passing an empty string I expect the library to tell me this and not fall back to defaults.

PR #627, opened 2016-07-22
allow for configuration keys with nil values to fall back onto defaults

cmw:
In our effort to turn our deployment into something 12-factor-compliant, we tried to base our configuration solely on ENV variables like so:

```
url: <%= ENV['REDIS_URL'].presence || nil %>
sentinels:
  - host: <%= ENV['REDIS_SENTINEL_HOST'].presence || nil %>
    port: <%= ENV['REDIS_SENTINEL_PORT'].presence || nil %>
role: <%= ENV['REDIS_ROLE'].presence || nil %>
timeout: <%= ENV['REDIS_TIMEOUT'].presence || 0.1 %>
db: <%= ENV['REDIS_DB'].presence || 0 %>
host: <%= ENV['REDIS_HOST'].presence || '127.0.0.1' %>
port: <%= ENV['REDIS_PORT'].presence || 6379 %>
```

This however results in the configuration hash holding several unnecessary keys with `nil` values. Thus these changes are needed so in case of a key existing but being nil, the default is still used.

I assume this was intended all along, but did only work for some of the keys, not all.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -86,7 +86,7 @@ class Redis
 
       @pending_reads = 0
 
-      if options.include?(:sentinels)
+      if options.include?(:sentinels) && options[:sentinels].any? { |conf| !conf[:host].nil? && !conf[:port].nil? }
         @connector = Connector::Sentinel.new(@options)
       else
         @connector = Connector.new(@options)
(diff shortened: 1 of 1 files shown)


cmw:
@badboy is there a chance to get this merged?

PR #534, opened 2015-06-11
Fix error "No sentinels available" when the master name is different from the host name

nguyenductung:
Associated with issue #531.
This allows `master_name` in options.

PR #630, opened 2016-08-04
Allow master_name to be specified

martin308:
This is a continuation of #534, I hope you don't mind @nguyenductung!

I have allowed specifying a master_name where it differs from the host and added a test for this.


martin308:
Ah I see now why this is unnecessary! The url that you specify when using sentinels is not a url to a redis instance in the collection.

PR #631, opened 2016-08-04
handling SocketError

hokuts:
When connecting to the Sentinel, the standard socket library may raise `SocketError`.
(e.g. `getaddrinfo: Name or service not known` )
But current redis-rb doesn't rescue it, so the process can not try to connect to the next Sentinel.
This patch will rescue it.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -336,6 +336,7 @@ class Redis
       @connection = @options[:driver].connect(@options)
       @pending_reads = 0
     rescue TimeoutError,
+           SocketError,
            Errno::ECONNREFUSED,
            Errno::EHOSTDOWN,
            Errno::EHOSTUNREACH,


Argoday:
+1

tanaylakhani:
+1

blmundie:
+1

casperisfine:
We've just had an issue with this in production. Rails's cache store rescue `Redis::BaseConnectionError`, but doesn't expect a `SocketError`.

@badboy @djanowski any chance we could get this merged ? 🙏

badboy:
I opted for just merging it, as I figure it should not break too much stuff... but, I am not maintaining redis-rb anymore, so I won't push a release.

fw42:
@badboy: Thanks! Who should we ping for a release?

PR #632, opened 2016-08-16
Adds IO::WaitWritable to nonblocking exceptions to support use with SSL

ryancalhoun:
When used with SSL, any read or write can trigger a handshake; this means we need to handle both nonblocking read and write errors.


djanowski:
#644, thanks.

PR #636, opened 2016-09-16
Retry "write would block" for SSL connections

Connorhd:
OpenSSL throws a different exception for EWOULDBLOCK.


djanowski:
@nerdrew @mdayaram @tarcieri This is superseded by #644?

tarcieri:
@djanowski yes, #644 is the correct solution

djanowski:
@tarcieri Thank you.

PR #628, opened 2016-07-23
Add spop count support

ceritium:
Copied from https://github.com/redis/redis-rb/pull/621 but passing tests.

All merits to @arashbm

PR #646, opened 2016-11-01
Support for `count` arg on `spop`

theorygeek:
Redis 3.2 added an optional `count` argument for the [SPOP command](http://redis.io/commands/spop).


theorygeek:
Nevermind, just realized that this is a duplicate of #628.

PR #647, opened 2016-11-09
Add optional message argument to ping

matrinox:
According to Redis, you can pass an optional argument to PING. To be consistent with their API, I've added that optional argument

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -143,10 +143,12 @@ class Redis
 
   # Ping the server.
   #
+  # @param [optional, String] message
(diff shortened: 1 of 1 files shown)


matrinox:
Can someone help me with the travis fails? I can't see how what I did would really cause so many failures, and randomly it feels like. First one was https://travis-ci.org/redis/redis-rb/builds/174376222 and only that config. Second commit is running now and already has 3 failures with seemingly cryptic errors. Help would be much appreciated

badboy:
Test failures are unrelated.

PR #649, opened 2016-11-18
Fix #_write_to_socket in connection

quixoten:
By placing the NBIO_READ_EXCEPTIONS above the NBIO_WRITE_EXCEPTIONS, an
Errno::EWOULDBLOCK or Errno::EAGAIN would trigger a wait for the socket
to be readable. This is causing timeout errors in our testing.

When calling #write_nonblock, the appropriate action is to wait for the
socket to be writable, not readable. Ensuring that we handle the write
exceptions before the read exceptions accomplishes this.

Unless I'm missing something?

### DIFF
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -306,9 +306,8 @@ class Redis
         if config[:scheme] == "unix"
           raise ArgumentError, "SSL incompatible with unix sockets" if config[:ssl]
           sock = UNIXSocket.connect(config[:path], config[:connect_timeout])
-        elsif config[:ssl] && RUBY_VERSION < "1.9.3"
-          raise ArgumentError, "This library does not support SSL on Ruby < 1.9"
         elsif config[:scheme] == "rediss" || config[:ssl]
+          raise ArgumentError, "This library does not support SSL on Ruby < 1.9" if RUBY_VERSION < "1.9.3"
           sock = SSLSocket.connect(config[:host], config[:port], config[:connect_timeout], config[:ssl_params])
         else
           sock = TCPSocket.connect(config[:host], config[:port], config[:connect_timeout])
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -27,7 +27,7 @@ class Redis
       CRLF = "\r\n".freeze
 
       # Exceptions raised during non-blocking I/O ops that require retrying the op
-      if RUBY_VERSION >= "1.9.0"
+      if RUBY_VERSION >= "1.9.3"
         NBIO_READ_EXCEPTIONS = [IO::WaitReadable]
         NBIO_WRITE_EXCEPTIONS = [IO::WaitWritable]
       else
@@ -105,14 +105,14 @@ class Redis
         begin
           write_nonblock(data)
 
-        rescue *NBIO_READ_EXCEPTIONS
-          if IO.select([self], nil, nil, @write_timeout)
+        rescue *NBIO_WRITE_EXCEPTIONS
+          if IO.select(nil, [self], nil, @write_timeout)
             retry
           else
             raise Redis::TimeoutError
           end
-        rescue *NBIO_WRITE_EXCEPTIONS
-          if IO.select(nil, [self], nil, @write_timeout)
+        rescue *NBIO_READ_EXCEPTIONS
+          if IO.select([self], nil, nil, @write_timeout)
             retry
           else
             raise Redis::TimeoutError
@@ -306,7 +306,7 @@ class Redis
         if config[:scheme] == "unix"
           raise ArgumentError, "SSL incompatible with unix sockets" if config[:ssl]
           sock = UNIXSocket.connect(config[:path], config[:connect_timeout])
-        elsif config[:ssl] && RUBY_VERSION < "1.9.0"
+        elsif config[:ssl] && RUBY_VERSION < "1.9.3"
           raise ArgumentError, "This library does not support SSL on Ruby < 1.9"
         elsif config[:scheme] == "rediss" || config[:ssl]
           sock = SSLSocket.connect(config[:host], config[:port], config[:connect_timeout], config[:ssl_params])
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -104,14 +104,14 @@ class Redis
         begin
           write_nonblock(data)
 
-        rescue *NBIO_READ_EXCEPTIONS
-          if IO.select([self], nil, nil, @write_timeout)
+        rescue *NBIO_WRITE_EXCEPTIONS
+          if IO.select(nil, [self], nil, @write_timeout)
             retry
           else
             raise Redis::TimeoutError
           end
-        rescue *NBIO_WRITE_EXCEPTIONS
-          if IO.select(nil, [self], nil, @write_timeout)
+        rescue *NBIO_READ_EXCEPTIONS
+          if IO.select([self], nil, nil, @write_timeout)
             retry
           else
             raise Redis::TimeoutError
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -27,11 +27,12 @@ class Redis
       CRLF = "\r\n".freeze
 
       # Exceptions raised during non-blocking I/O ops that require retrying the op
-      NBIO_READ_EXCEPTIONS = [Errno::EWOULDBLOCK, Errno::EAGAIN]
-      NBIO_WRITE_EXCEPTIONS = [Errno::EWOULDBLOCK, Errno::EAGAIN]
-      if RUBY_VERSION >= "1.9.3"
-        NBIO_READ_EXCEPTIONS << IO::WaitReadable
-        NBIO_WRITE_EXCEPTIONS << IO::WaitWritable
+      if RUBY_VERSION >= "1.9.0"
+        NBIO_READ_EXCEPTIONS = [IO::WaitReadable]
+        NBIO_WRITE_EXCEPTIONS = [IO::WaitWritable]
+      else
+        NBIO_READ_EXCEPTIONS = [Errno::EWOULDBLOCK, Errno::EAGAIN]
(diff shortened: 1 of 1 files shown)


quixoten:
To be more explicit, here's the _previous_ version of `_write_to_socket` with NBIO_READ_EXCEPTIONS and NBIO_WRITE_EXCEPTIONS expanded (comments mine):

```ruby
def _write_to_socket(data)
  begin
    write_nonblock(data)

  # This first rescue is the issue. when calling #write_nonblock,
  # Errno::EWOULDBLOCK and Errno::EAGAIN should be handled by waiting for the
  # socket to be writable, but we're waiting for it to be readable here, which will
  # Timeout in many cases
  rescue Errno::EWOULDBLOCK, Errno::EAGAIN, IO::WaitReadable
    if IO.select([self], nil, nil, @write_timeout)
      retry
    else
      raise Redis::TimeoutError
    end

  # Errno::EWOULDBLOCK and Errno::EAGAIN are ignored in this rescue because
  # they are also included in the rescue above (Ruby only runs the first
  # match).
  rescue Errno::EWOULDBLOCK, Errno::EAGAIN, IO::WaitWritable
    if IO.select(nil, [self], nil, @write_timeout)
      retry
    else
      raise Redis::TimeoutError
    end
  end

rescue EOFError
  raise Errno::ECONNRESET
end
```

mdayaram:
We've encountered this problem as well in a different app actually.  Though I wonder if the better solution is to just *not* include Errno::EWOULDBLOCK, Errno::EAGAIN in NBIO_READ_EXCEPTIONS.

mdayaram:
Actually, nvm, this makes sense -- the order is dependent on which operation you're doing (rescue write exceptions first on write operations and read exceptions first on read operations).  lgtm, @djanowski could you take a look?

tarcieri:
@quixoten 

> When calling #write_nonblock, the appropriate action is to wait for the
> socket to be writable, not readable. Ensuring that we handle the write
> exceptions before the read exceptions accomplishes this.
>
> Unless I'm missing something?

Yes, you are missing something: SSL.

SSL introduces cases where you might have to read to complete a write or write to complete a read, such as handshaking or renegotiation.

To properly support SSL, both read and write interests need to be handled for both reading and writing.

I think the proper solution is to make `IO::WaitReadable` and `IO::WaitWritable` the only exceptions rescued for `RUBY_VERSION >= "1.9.0"`. The other ones are only there for 1.8 support.

nerdrew:
@tarcieri like so?

Exceptions:
```
      # Exceptions raised during non-blocking I/O ops that require retrying the op
      NBIO_EXCEPTIONS = [IO::WaitReadable, IO::WaitWritable]
      if RUBY_VERSION < "1.9.3"
        NBIO_EXCEPTIONS << [Errno::EWOULDBLOCK, Errno::EAGAIN]
      end
```

e.g. read:
```
      def _read_from_socket(nbytes)

        begin
          read_nonblock(nbytes)

        rescue *NBIO_EXCEPTIONS
          if IO.select([self], [self], nil, @timeout)
            retry
          else
            raise Redis::TimeoutError
          end
        end

      rescue EOFError
        raise Errno::ECONNRESET
      end
```

nerdrew:
Also, question about `timeout` and `write_timeout`. They look like socket timeouts, not the overall timeouts. Is that true?

quixoten:
@tarcieri: I'm not following. This does not remove support for rescuing `IO::WaitReadable` when calling `write_nonblock` on the socket. It just changes what happens when `Errno::EWOULDBLOCK` or `Errno::EAGAIN` are raised.

nerdrew:
@quixoten Do you have a test case that triggers the error you were seeing?

quixoten:
@nerdrew: I do not have a test case. We rolled `3.3.2` out to a few nodes of our application and started seeing the Redis::TimeoutError being raised. Rolling back to `3.3.1` resolved the issue.

tarcieri:
@nerdrew yes. Also note that nonblocking SSL won't work at all on 1.8 as the OpenSSL module didn't support nonblocking operations at all.

quixoten:
@nerdrew: I think using `IO.select([self], [self], nil @timeout)` will cause short bursts of rapid retry loops because a connected socket is going to be almost always writable. If the code is waiting for the socket to be readable, it'll be retrying over and over again.

tarcieri:
> I'm not following. This does not remove support for rescuing IO::WaitReadable when calling write_nonblock on the socket.

@quixoten again, you MUST handle `IO::WaitReadable` when calling `write_nonblock` on an SSL socket. As I explained before: "SSL introduces cases where you might have to read to complete a write or write to complete a read, such as handshaking or renegotiation."

quixoten:
@tarcieri: I'm in complete agreement with you. 100 percent. What I'm unsure of is how this change fails to account for that scenario. Here's the effective code after this change:

```ruby
def _write_to_socket(data)
  begin
    write_nonblock(data)

  rescue Errno::EWOULDBLOCK, Errno::EAGAIN, IO::WaitWritable
    if IO.select(nil, [self], nil, @write_timeout)
      retry
    else
      raise Redis::TimeoutError
    end
  rescue IO::WaitReadable
    if IO.select([self], nil, nil, @write_timeout)
      retry
    else
      raise Redis::TimeoutError
    end
  end

rescue EOFError
  raise Errno::ECONNRESET
end
```

and this is the effective code before the change:

```ruby
def _write_to_socket(data)
  begin
    write_nonblock(data)

  rescue Errno::EWOULDBLOCK, Errno::EAGAIN, IO::WaitReadable
    if IO.select([self], nil, nil, @write_timeout)
      retry
    else
      raise Redis::TimeoutError
    end
  rescue IO::WaitWritable
    if IO.select(nil, [self], nil, @write_timeout)
      retry
    else
      raise Redis::TimeoutError
    end
  end

rescue EOFError
  raise Errno::ECONNRESET
end
```

tarcieri:
On 1.9+, it should look like this:

```ruby
def _write_to_socket(data)
  begin
    write_nonblock(data)

  rescue IO::WaitReadable
    if IO.select([self], nil, nil, @write_timeout)
      retry
    else
      raise Redis::TimeoutError
    end
  rescue IO::WaitWritable
    if IO.select(nil, [self], nil, @write_timeout)
      retry
    else
      raise Redis::TimeoutError
    end
  end

rescue EOFError
  raise Errno::ECONNRESET
end
```

tarcieri:
@quixoten I *think* that should do the trick.

For comparison, here's similar logic from http.rb:

https://github.com/httprb/http/blob/master/lib/http/timeout/global.rb#L82

quixoten:
@tarcieri: yeah, I think I know where my assumptions were breaking down. Because Ruby is extending the original error with `IO::WaitReadable` or `IO::WaitWritable` rather than raising them directly, rescuing `Errno::EWOULDBLOCK` or `Errno::EAGAIN` is always going to lead to ambiguity with an SSL socket.

tarcieri:
@quixoten yeah, by rescuing both you're short circuiting the exception handing, and potentially handling the wrong one

quixoten:
Does that mean SSL will be broken on Ruby < 1.9?

tarcieri:
@quixoten yes. As I [noted earlier](https://github.com/redis/redis-rb/pull/649#issuecomment-261668897):

> Also note that nonblocking SSL won't work at all on 1.8 as the OpenSSL module didn't support nonblocking operations at all.

Attempting to enable SSL on 1.8 should probably raise an exception.

quixoten:
Would this be acceptable ?

```ruby
def self.connect(config)
  if config[:scheme] == "unix"
    raise ArgumentError, "SSL incompatible with unix sockets" if config[:ssl]
    sock = UNIXSocket.connect(config[:path], config[:connect_timeout])
  elsif config[:ssl] && RUBY_VERSION < "1.9.0"
    raise ArgumentError, "This library does not support SSL on Ruby < 1.9"
  ...
  end
...
end
```

tarcieri:
LGTM

tarcieri:
Err, actually, on second thought, I think it should be this...

```ruby
elsif config[:scheme] == "rediss" || config[:ssl]
  raise ArgumentError, "This library does not support SSL on Ruby < 1.9" if RUBY_VERSION < "1.9.3"
```

I believe nonblocking SSL wasn't added until 1.9.3.

quixoten:
I still need to change the order of the rescues right? Otherwise Ruby < 1.9 is going to handle Errno::EWOULDBLOCK and Errno::EAGAIN by waiting for the socket to be readable.

tarcieri:
@quixoten yes, make sure rescue for the op you're performing comes first

nerdrew:
Is the ruby OpenSSL documentation incorrect? https://ruby-doc.org/stdlib-2.3.1/libdoc/openssl/rdoc/OpenSSL/Buffering.html#method-i-write_nonblock

It has an example for `write_nonblock` that rescues `IO::WaitReadable` first...

tarcieri:
@nerdrew the problem is Ruby 1.8 doesn't have `IO::WaitReadable` or `IO::WaitWritable` (or nonblocking SSL support at all)... on those platforms we're subbing a different set of exceptions (e.g. `Errno::EAGAIN`), so we need to make sure it has the right behavior there.

badboy:
I didn't follow the whole discussion, so one question: will the actual correct implementation be easier without support for Ruby 1.8?

quixoten:
@badboy: in its current state, the implementation is correct (for Ruby 1.8 as well). Removing support for 1.8 would not meaningfully impact the difficulty of a correct implementation, in my opinion.

djanowski:
But it wouldn't be correct to officially drop support for SSL in 1.8 in a bugfix release.

Honestly, I'm tired of wasting our time with 1.8.

Would y'all test a 4.0 release candidate with this fix?

quixoten:
I could test a 4.0 release candidate.

djanowski:
4.0.0.rc1 is out. Can you please give it a try?

NIHERASE:
Hello guys this problem affects us as well, wouldn't it be reasonable to remove the bugged version from rubygems? Or it's troublesome?

djanowski:
@NIHERASE Does 4.0.0-rc1 work for you?

djanowski:
Released 3.3.3 which includes this fix.

PR #653, opened 2016-11-23
4.0

djanowski:


.travis.yml
```diff
@@ -1,16 +1,13 @@
...
   - 2.1
   - 2.2
-  - 2.3.0
-  - jruby-18mode
-  - jruby-19mode
-  - jruby-9.0.5.0
-  - rbx-2
+  - 2.3.3
+  - jruby-9
+  - rbx-3
```
badboy: Should we drop rbx-3 testing/support as well? We allow failures anyway and usage of it is probably low anyway

djanowski: This was added quite recently. I think it doesn't hurt that a Rubinius user can quickly figure out whether redis-rb works on Rubinius or not...

badboy: Well, given that sometimes the tests work and sometimes they don't doesn't really give anyone any idea about the state of Rubinius support

jgoulah:
Just curious about the decision to remove Redis::Distributed (I can't find any PR or discussion).  When I attempted to use the 4.0.0-rc1 I am getting an error from redis-store `cannot load such file -- redis/distributed` and this version is also required by redis-rails which is at the newest version.  While I understand its not your problem to support all these wrappers, it would be good to know why you dropped support in a totally backward incompatible way and some documentation on how to get around it. Is it as simple as not requiring redis/distributed any longer?

PR #657, opened 2016-11-30
Wrap exceptions for missing socket with CannotConnectError

### LINKED ISSUE #656: Error Handling Around Missing Socket
We have a number of hosts deployed with local redis instances that are accessed over a unix socket.  The documentation is straightforward and this all works great.  However, occasionally when hosts are brought down the socket may be removed, which breaks redis communication.  This is also not a problem, but our error handling logic only gracefully resolves all redis related errors that stem from Redis::BaseError.  

The connection error for a socket that does not exist is not wrapped with Redis::CannotConnectError.  However, all sorts of other connection errors (e.g. unreachable hosts, timeouts, etc.) are.  For example,

```
redis = Redis.new(:path => "/tmp/redis.sock")
redis.set("mykey", "hello world")
```
raises `Errno::ENOENT: No such file or directory - connect(2) for /tmp/redis.sock`.  However,

```
redis = Redis.new(:url => "redis://missing.host")
redis.set("mykey", "hello world")
```
raises `Redis::CannotConnectError: Error connecting to Redis on missing.host:6379 (Errno::ECONNREFUSED)` essentially wrapping the underlying error.

Is this intentional or can I submit pull request that handles Errno::ENOENT alongside the other connection errors?  Thanks.

sallustfire:
Resolves #656.  Test is added as well, though may not be wanted.  Did not see tests checking the other types of connection errors except for timeout.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -340,6 +340,7 @@ class Redis
            Errno::EHOSTDOWN,
            Errno::EHOSTUNREACH,
            Errno::ENETUNREACH,
+           Errno::ENOENT,
            Errno::ETIMEDOUT
 
       raise CannotConnectError, "Error connecting to Redis on #{location} (#{$!.class})"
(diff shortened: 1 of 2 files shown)


djanowski:
Marking this for 4.0 because we'd be changing the exception class.

djanowski:
Fixed in 4.0.

PR #658, opened 2016-12-02
Add support for EAGAINWait{Readable, Writable}

igorwwwwwwwwwwwwwwwwwwww:
Beginning in ruby 2.1.0 there is an `EAGAINWaitReadable` in addition to `WaitReadable` that also needs to be trapped and select+retried.

### DIFF
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -33,6 +33,10 @@ class Redis
         NBIO_READ_EXCEPTIONS << IO::WaitReadable
         NBIO_WRITE_EXCEPTIONS << IO::WaitWritable
       end
+      if RUBY_VERSION >= "2.1.0"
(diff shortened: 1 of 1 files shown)


igorwwwwwwwwwwwwwwwwwwww:
This is causing us to run into `Redis::TimeoutError` every now and then.

igorwwwwwwwwwwwwwwwwwwww:
Ah, I just found that we are running v2.3.1, which does not include this fix: https://github.com/redis/redis-rb/commit/cf9d7b376a2949759638ba51619a6d825a04ac59.

And from a quick look, it seems like `EAGAINWaitReadable` includes `WaitReadable` -- so this patch might not be necessary after all. :)

igorwwwwwwwwwwwwwwwwwwww:
confirm, closing. 🎉

PR #660, opened 2016-12-17
Avoid Fixnum deprecation warning on Ruby trunk

utilum:


irobayna:
Any chance we can merge this item?

badboy:
Let me grab 2.4-rc1 and see for myself. Otherwise I don't see why not.

badboy:
Thanks @utilum for the patch, thanks @irobayna for the reminder.

matthewrudy:
Any chance of getting a release of this?
Would be great to have ruby 2.4.0 running without any deprecations

badboy:
Hm, we should probably get [4.0](https://github.com/redis/redis-rb/pull/653) out of the door.

manuelmeurer:
Would love to see v3.3.3 with this change released! 😄

vincentwoo:
Just to echo the sentiments here in case this got forgotten in the holiday shuffle, I think a tiny patch release just so we can silence the deprecation warnings would be great!

georgemillo:
Just so everyone knows, it looks like 3.3.3 contains the above commit. It's not reflected in the  changelog for 3.3.3, but I just submitted a PR to add it ;)

matthewrudy:
@georgemillo yep
I switched to 3.3.3 a while ago,
works for me!

PR #663, opened 2016-12-30
Allow master name to be set

bschaeffer:


bschaeffer:
Sorry, kept opening this on redis-rb instead of my own fork

PR #677, opened 2017-02-06
add note to CHANGELOG about Ruby 2.4 deprecation warnings

georgemillo:
I noticed that as of 3.3.3, redis-rb no longer raises deprecation warnings on
Ruby 2.4 related to Fixnum etc (see #660), but the CHANGELOG doesn't reflect
this.

Updated the changelog with the new info ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 # 3.3.3
 
+* Remove deprecation warnings in Ruby 2.4 related to `Fixnum` vs `Integer`
+
 * Improved timeout handling after dropping Timeout module.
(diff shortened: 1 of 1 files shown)


djanowski:
Not a feature or bug fix. Thanks.

sandstrom:
@djanowski I think this should have been pulled in.

I was looking for exactly this line in the changelog and didn't find it, so I started searching through the issues after `2.4` to see if there was a pending issue or if I should open a PR.

Changelogs are written for humans, sometimes information that isn't directly related to a feature or bug can still be included because it's useful.

PR #678, opened 2017-02-07
SSL params: allow skipping hostname verification 

zerobearing2:
**Use case:** 

Using _heroku-redis_, its possible to connect directly to its _stunnel_ port to secure the data transport -- however, the SSL certificate hostname from heroku _(aka AWS)_ is an internal address, not the public DNS record used to initiate the connection to heroku-redis instance. Therefore, it required **skipping hostname checks** to make a successful connect.

Given the unsafe nature of this change, is there merit to adding this feature and allowing others to use _skip hostname verification_ if they need too?

Also, I'm unsure the proper way to test this setting, any suggestion would be appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,17 @@ redis = Redis.new(
 )
 ```
 
+Turn off hostname verification using `OpenSSL::SSL::SSLContext#verify_mode`, ie:
+
+```ruby
+redis = Redis.new(
+  :url        => "rediss://:p4ssw0rd@10.0.1.1:6381/15",
+  :ssl_params => {
+    :verify_mode => OpenSSL::SSL::VERIFY_NONE
+  }
+)
+```
+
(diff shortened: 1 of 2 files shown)


atsheehan:
I'm also running into this issue using the Heroku Redis Ruby connection instructions: https://devcenter.heroku.com/articles/securing-heroku-redis#using-ruby

It looks like a few other libraries skip the `post_connection_check` if `OpenSSL::SSL::VERIFY_NONE` is set, including Net::HTTP:

* https://github.com/ruby/ruby/blob/v2_4_0/lib/net/http.rb#L949-L951
* https://github.com/excon/excon/pull/56

meenie:
Any update on this? For now, I'm going to extend the Redis gem.

byroot:
Superseeded by https://github.com/redis/redis-rb/pull/745

PR #683, opened 2017-03-15
Adding support for the ZLEXCOUNT command on sorted set.

lscspirit:
Adding support for `ZLEXCOUNT` command. This command has been available since Redis `2.8.9`, along with the other lexicographical commands on sorted set - `ZRANGEBYLEX`, `ZREVRANGEBYLEX` and `ZREMRANGEBYLEX`

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -1701,6 +1701,30 @@ class Redis
     end
   end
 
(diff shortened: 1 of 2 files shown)


badboy:
There we go \o/

Issue #12, opened 2010-01-04
DistRedis#mget Quesion/Issue


levicook:
I don't think DistRedis#mget maps to Redis#mgets correctly.

I created a gist to highlight the issue. http://gist.github.com/268284

Do you think DistRedis should map mgets this way?

Issue #282, opened 2012-10-13
Pipelining and non-existent keys


iamjoshua:
I'm unable to get any false returns from pipelining with redis-rb as I should via redis.
Take the following example:

```
keys = [1, 2, 3]
res = $redis.pipelined do
    keys.each do | key |
        $redis.exists(key)  
    end
end
```

Assuming key 2 doesn't exist, res will return:
[true, true] 
instead of
[true, false, true]

and If I try to use futures like so:

```
keys = [1, 2, 3]
results = Hash.new
$redis.pipelined do
    keys.each do | id |
        results[id] = $redis.exists(id)
    end
end
```

I'll get the following error:

```
undefined method `is_a?' for <Redis::Future [:exists, "1"]>:Redis::Future
```

Am I missing something or is this a bug?

PR #687, opened 2017-04-13
Distributed: support mget and mapped_mget

jeremy:
Upstreaming a patch we use to support distributed cache lookups. Map the keys to their nodes, mget them on each node, and stitch the results together.

Apparently this component is frozen and unmaintained. That's a shame, considering caching is a common use case for Redis, and will become more so with upcoming support for LFU expiry. Sharing for posterity, in any case.

Related: #673, #554, #478, #12.

### DIFF
--- a/lib/redis/distributed.rb
+++ b/lib/redis/distributed.rb
@@ -277,13 +277,16 @@ class Redis
       node_for(key).get(key)
     end
 
-    # Get the values of all the given keys.
+    # Get the values of all the given keys as an Array.
     def mget(*keys)
-      raise CannotDistribute, :mget
+      mapped_mget(*keys).values_at(*keys)
     end
 
+    # Get the values of all the given keys as a Hash.
     def mapped_mget(*keys)
-      raise CannotDistribute, :mapped_mget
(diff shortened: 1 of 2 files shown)


lib/redis/distributed.rb
```diff
@@ -277,13 +277,18 @@ def get(key)
...
     def mget(*keys)
-      raise CannotDistribute, :mget
+      mapped_mget(*keys).values_at(*keys)
     end
 
+    # Get the values of all the given keys as a Hash.
     def mapped_mget(*keys)
-      raise CannotDistribute, :mapped_mget
+      results = {}
+      keys.group_by { |k| node_for k }.each do |node, subkeys|
```
tubbo: this could also be `each_with_object({})`.

tubbo:
**redis-store** is waiting to merge #282 until this one gets into **redis.rb**.

jeremy:
https://github.com/redis-store/redis-store/pull/282 😊

djanowski:
Should be out soon in the form of ~4.0.1.

PR #689, opened 2017-04-20
Add missing defaults

aeroastro:
I would like to add some missing keys in DEFAULTS so that options with string keys are accepted.

Also, I added a comment in order not to forget to add keys.

Since Ruby 2.2 and after has symbol GC functionality, it might be about time to convert all string keys to symbols.

c.f. #256 #259

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -6,12 +6,16 @@ require "cgi"
 
 class Redis
   class Client
+    # Defaults are also used for converting string keys to symbols.
     DEFAULTS = {
       url: -> { ENV["REDIS_URL"] },
       scheme: "redis",
       host: "127.0.0.1",
       port: 6379,
       path: nil,
+      read_timeout: nil,
+      write_timeout: nil,
+      connect_timeout: nil,
       timeout: 5.0,
       password: nil,
       db: 0,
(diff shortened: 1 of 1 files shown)


aeroastro:
Hello.

It's been a while since I submitted this PR, and I'm wondering if there is anything I could do on my end. I totally understand you are busy and require more time to consider, it is much appreciated if you could give me a response.

Also, IMHO, memory leak pointed out here https://github.com/redis/redis-rb/pull/256#issuecomment-7211535 is becoming a very minor issue and I prefer converting all the keys from String to Symbol.

If it is O.K. I will fix my patch!

Here are the reasons.

1. Rational developers won't pass hashes containing a large number of unsupported keys to external libraries as their options, especially on environments where memory leak is a problem.
2. In most cases, unintentionally inserted keys in options will not change dynamically until the program ends, which means unintentionally used memory size by converted Symbols is limited (fixed).
3. Even if some developers pass the options containing a large number of dynamically-changing unsupported keys, currently (September, 2017) supported ruby versions, which is 2.2 and later, have Symbol GC functionality and memory leak is restrained to some extent.

aeroastro:
@pietern 

Could you take a short look into this PR?
I think it is about time to convert all keys from String to Symbol.

aeroastro:
@byroot 
I have completed adjusting the commit to the latest master branch.
Could you take a look into this?

aeroastro:
Thank you ⭐

PR #690, opened 2017-04-29
Allow hostname checks to be disabled

mattheworiordan:
When using mutual certificate authentication, it is often unnecessary (and a blocker) to have the certificate match the host domain. The security model is provided with the client certificates.

`verify_hostname` is firstly a valid option of [SSLContext](https://docs.ruby-lang.org/en/2.4.0/OpenSSL/SSL/SSLContext.html#method-i-set_params), so this option and intention should be carried through in this client when using TLS certificate authentication.

### DIFF
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -287,7 +287,7 @@ class Redis
           ssl_sock.hostname = host
           ssl_sock.connect
 
-          unless ssl_params.has_key?(:verify_hostname) && ssl_params[:verify_hostname] == false
+          unless ssl_params && ssl_params[:verify_hostname] == false
             ssl_sock.post_connection_check(host)
           end
 
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -286,7 +286,10 @@ class Redis
           ssl_sock = new(tcp_sock, ctx)
           ssl_sock.hostname = host
           ssl_sock.connect
-          ssl_sock.post_connection_check(host)
+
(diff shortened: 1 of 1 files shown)


lib/redis/connection/ruby.rb
```diff
@@ -286,7 +286,10 @@ def self.connect(host, port, timeout, ssl_params)
           ssl_sock = new(tcp_sock, ctx)
           ssl_sock.hostname = host
           ssl_sock.connect
-          ssl_sock.post_connection_check(host)
+
+          unless ssl_params.has_key?(:verify_hostname) && ssl_params[:verify_hostname] == false
```
olleolleolle: The last half of the check is benign, and the first part (`ssl_params.has_key?(:verify_hostname)`) _could_ be left out.

Example:

```
irb(main):009:0> ssl_params = { }
{}
irb(main):010:0> ssl_params[:verify_hostname] == false
false #   which is the value we are after
```

mattheworiordan: True, not sure I recall why I checked for the key in the first place, will issue a fixup.

mattheworiordan: See 4bafef2. Note for consistency with the code for that method, a truthy check for `ssl_params` is done I assume because it's expected that `ssl_params` could be null

olleolleolle: Thanks for putting in this effort!

sstarcher:
Anything else needed to get this merged in

mattheworiordan:
Not that I am aware of. It’s been a while though.

byroot:
Superseeded by https://github.com/redis/redis-rb/pull/745

PR #692, opened 2017-05-03
Travis: Remove rbx-2 from the CI matrix

olleolleolle:
This PR removes rbx-2 from the CI matrix.

Rubinius expressed like that errors on installation:

```
Requested binary installation but no rubies are available to download, consider skipping --binary flag.
Gemset '' does not exist, 'rvm rbx-2 do rvm gemset create ' first, or append '--create'.
The command "rvm use rbx-2 --install --binary --fuzzy" failed and exited with 2 during .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ rvm:
   - jruby-18mode
   - jruby-19mode
   - jruby-9.0.5.0
-  - rbx-2
 
 gemfile: ".travis/Gemfile"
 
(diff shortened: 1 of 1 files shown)


olleolleolle:
I wanted to make this change simple, so I kept the existing versions of Rubies in there.

I want to make another PR with "update CI matrix to current releases of each ruby".

djanowski:
We're doing rbx-3 only now, thanks.

PR #701, opened 2017-06-21
Get tests passing with frozen-string-literals enabled.

pat:
This one simple change ensure that all string literals can be frozen (as per the optional feature in MRI 2.3 and onwards). @twalpole has (again) beaten me to such a patch (in #590), though mine (again) does not add the pragma comment to all files. Getting their or my PRs merged in would be excellent :)

As an alternative to the pragma comment, I would recommend adding the following to your .travis.yml file to ensure regressions aren't introduced:

```yml
before_script:
- if (ruby -e "exit RUBY_VERSION.to_f >= 2.4"); then export RUBYOPT="--enable-frozen-string-literal"; fi; echo $RUBYOPT
```

This will add the flag when the tests are run on MRI 2.4 or newer (while the feature was introduced in 2.3, it doesn't seem to work reliably until 2.4). Please note: tests will currently fail when this flag is set unless test-unit is also updated (as noted in https://github.com/test-unit/test-unit/pull/149).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ rvm:
   - 2.1
   - 2.2
   - 2.3.0
+  - 2.4.1
   - jruby-18mode
   - jruby-19mode
   - jruby-9.0.5.0
@@ -15,6 +16,9 @@ rvm:
 gemfile: ".travis/Gemfile"
 
 sudo: false
+
+before_script:
+- if (ruby -e "exit RUBY_VERSION.to_f >= 2.4"); then export RUBYOPT="--enable-frozen-string-literal"; fi; echo $RUBYOPT
 
 env:
   global:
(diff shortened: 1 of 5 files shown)


pat:
For what it's worth, I've added MRI 2.4.1 and the RUBYOPT flag to the test suite, along with the latest release of test-unit, and everything (beyond rbx-2) is green.

badboy:
Great! 

bors r+

pat:
bors really isn't having much luck :(

badboy:
Yeah, turns out I'm basically a beta tester for bors and discover all the bugs :D

badboy:
another try
bors r+

badboy:
There you go.

pat:
Great, thanks :)

badboy:
I'll try to get a release out by the end of the week

PR #709, opened 2017-07-25
Use :read_timeout instead of :timeout for read timeout

### LINKED ISSUE #708: :read_timeout config value is not honored for 'ruby' driver
When I run the following script:
```ruby
require 'redis'
require 'hiredis'

# we will use this as a long running redis call. It takes a little over 1 second on my 
# machine, alter the number of iterations so it runs in one second again.
def long_call(redis)
    redis.eval("local a = 0; while a < 80000000 do a = a + 1 end", [], [])
    puts "No timeout"
  rescue Redis::TimeoutError
    puts "Caught TimeoutError"
end

# no problem
r1 = Redis.new
long_call(r1)

# no problem
r2 = Redis.new(timeout: 3)
long_call(r2)

# This raises an error, correctly
r3 = Redis.new(timeout: 0.1)
long_call(r3)

# This should raise an error, but doesn't
r4 = Redis.new(timeout: 3, read_timeout: 0.1)
long_call(r4)

# This should raise an error, but doesn't
r5 = Redis.new(read_timeout: 0.1)
long_call(r5)

# The hiredis driver raises an error, correctly
r6 = Redis.new(driver: :hiredis, read_timeout: 0.1) 
long_call(r6) # should time out, and does
```

So what seems to happen is that the `:read_timeout` is ignored in the normal driver and that the `:timeout` config value is used instead. This is contrary to what the README says, the `:read_timeout` should overwrite the `:timeout` value and it should also be able to be used alone (ie without `:connect_timeout` and `:write_timeout`). The `hiredis` driver seems to work as expected.

WJWH:
This will fix #708. After this fix all the tests still run on my machine and my test script from the issue gives the expected results (run with `bundle exec <scriptname>` with the hiredis bits commented out and see the results). 

I scanned the code but could not see any other renamings that could be done without refactoring lots of code. Also, I thought about writing a separate test for this but decided against it because a. I don't think I know enough about the testing system of this library to make it idiomatic, b. This test might fail on other (faster/slower) machines than mine and c. testing timeouts is inherently messy and I dislike waiting in my test suite.

I welcome comments and/or thoughts for improvements of course :)

### DIFF
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -314,7 +314,7 @@ class Redis
         end
 
         instance = new(sock)
-        instance.timeout = config[:timeout]
+        instance.timeout = config[:read_timeout]
         instance.write_timeout = config[:write_timeout]
         instance.set_tcp_keepalive config[:tcp_keepalive]
         instance


pandurang90:
@pietern @djanowski any updates on this?

mperham:
Not backwards compatible.  This would break everyone depending on the :timeout option, no?

WJWH:
Hi @mperham! It's been a while since I looked at this, but I believe that existing code should not break because [if you don't set `read_timeout` it will default to the value of `timeout](https://github.com/redis/redis-rb/blob/929f83a413aa92835e9ac2892c7949646c556c42/lib/redis/client.rb#L440).

The code for other drivers already acts this way ([synchrony driver](https://github.com/redis/redis-rb/blob/60971a1723c70609974535a3cdbbcf92ae136629/lib/redis/connection/synchrony.rb#L99), [hiredis driver](https://github.com/redis/redis-rb/blob/60971a1723c70609974535a3cdbbcf92ae136629/lib/redis/connection/hiredis.rb#L23)), and it is also the behavior proscribed by the `README.md`.

Issue #546, opened 2015-08-20
Support for Redis Cluster (3.0)


iandyh:
Hi

Surprisingly, I haven't found too much information on the status of cluster support in Ruby client. I am aware of `redis-rb-cluster` from @antirez . But apparently, it's not production ready yet.

So any plans in `redis-rb`? Or when `redis-rb-cluster` can be merged into `redis-rb`?

spuder:
I'm also finding that there is no documented way to follow ASK and MOVED redirections (`redis-cli -c`)

```
2.2.0 :014 > redis = Redis.new(:host => "10.10.10.10")
 => #<Redis client v3.2.1 for redis://10.10.10.10:6379/0> 
2.2.0 :015 > redis.get("foo")
Redis::CommandError: MOVED 12365 10.10.10.10:6379
	from /Users/sowen/.rvm/gems/ruby-2.2.0/gems/redis-3.2.1/lib/redis/client.rb:113:in `call'
	from /Users/sowen/.rvm/gems/ruby-2.2.0/gems/redis-3.2.1/lib/redis.rb:789:in `block in get'
	from /Users/sowen/.rvm/gems/ruby-2.2.0/gems/redis-3.2.1/lib/redis.rb:37:in `block in synchronize'
	from /Users/sowen/.rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/monitor.rb:211:in `mon_synchronize'
	from /Users/sowen/.rvm/gems/ruby-2.2.0/gems/redis-3.2.1/lib/redis.rb:37:in `synchronize'
	from /Users/sowen/.rvm/gems/ruby-2.2.0/gems/redis-3.2.1/lib/redis.rb:788:in `get'
	from (irb):15
	from /Users/sowen/.rvm/rubies/ruby-2.2.0/bin/irb:11:in `<main>'
```

iandyh:
@spuder The doc is here: http://redis.io/topics/cluster-spec. It explains how to handle the redirections.

spuder:
Thanks. I've read that document twice and I'm not able to grok what needs to change. Is this a client side or server side configuration change?

badboy:
The redirect needs to be handled on the client side.
redis-rb currently does no handling of Cluster, and thus also don't follows the redirect.
redis-rb-cluster is a reference implementation on top of the existing redis-rb, that you can use. 
[...]

PR #716, opened 2017-09-26
Add Redis Cluster support

supercaracal:
We want to use Redis Cluster in session store such as.

There are some gems such as:
https://github.com/redis-store/redis-rails
https://github.com/redis-store/redis-store

These gems depend to redis-rb. But redis-rb doesn't support Redis Cluster.
https://github.com/redis-store/redis-rails/issues/72
https://github.com/redis-store/redis-activesupport/issues/89

So, we add a client to redis-rb for Redis Cluster. We can check it at this sample.
https://github.com/supercaracal/redis-cluster-playground

```ruby
nodes = (7000..7005).map { |port| "redis://127.0.0.1:#{port}" }
redis = Redis.new(cluster: nodes)
redis.set('hogehoge', 1)
redis.get('hogehoge')

# @see https://redis.io/commands/readonly
Redis.new(cluster: nodes, replica: true)
```

ref: #546
ref: https://github.com/antirez/redis-rb-cluster/pull/6
ref: https://github.com/antirez/redis-rb-cluster/issues/8
ref: https://github.com/NodeRedis/node_redis/issues/574
ref: https://github.com/andymccurdy/redis-py/issues/931
ref: https://github.com/andymccurdy/redis-py/pull/604

ref: https://github.com/go-redis/redis
ref: https://github.com/luin/ioredis
ref: https://github.com/xetorthio/jedis
ref: https://github.com/redisson/redisson
ref: https://github.com/phpredis/phpredis
ref: https://github.com/nrk/predis


supercaracal:
- [ ] ci failed

supercaracal:
- [ ] ci failed

```
Failure:
  Exception raised:
  Redis::CannotConnectError(<Error connecting to Redis on tmp/redis-unstable/redis.sock (Errno::ECONNREFUSED)>)
test_connecting_to_unix_domain_socket(TestInternals)
/home/travis/build/redis/redis-rb/test/internals_test.rb:328:in `test_connecting_to_unix_domain_socket'
     325:   end
     326: 
     327:   def test_connecting_to_unix_domain_socket
  => 328:     assert_nothing_raised do
     329:       Redis.new(OPTIONS.merge(:path => ENV.fetch("SOCKET_PATH"))).ping
     330:     end
     331:   end
```

supercaracal:
- [ ] ci failure

```
$ rvm use jruby-9 --install --binary --fuzzy
Unknown ruby string (do not know how to handle): jruby-9.1.13.0200.
jruby-9.1.13.0200 is not installed - installing.
Unknown ruby string (do not know how to handle): jruby-9.1.13.0200.
Searching for binary rubies, this might take some time.
Unknown ruby string (do not know how to handle): jruby-9.1.13.0200.
Requested binary installation but no rubies are available to download, consider skipping --binary flag.
Gemset '' does not exist, 'rvm jruby-9.1.13.0200 do rvm gemset create ' first, or append '--create'.
The command "rvm use jruby-9 --install --binary --fuzzy" failed and exited with 2 during .
```

badboy:
Yes, these test failures look unrelated. I currently don't have time to look at the PR so don't expect a decision any time soon

supercaracal:
- [ ] Fix hash tags bug

supercaracal:
Is `ASKING` implicitly defined command? We could not find `ASKING` in https://redis.io/commands

lib/redis/cluster.rb
```diff
@@ -0,0 +1,197 @@
...
+      when 'slots' then cluster_slots(response)
+      when 'nodes' then cluster_nodes(response)
+      when 'slaves' then cluster_slaves(response)
+      when 'info' then cluster_info(response)
+      else response
+      end
+    end
+
+    def asking
+      try_cmd(find_node, :synchronize) { |client| client.call(%i[asking]) }
```
supercaracal: https://github.com/antirez/redis-rb-cluster/issues/4#issuecomment-139825363

supercaracal:
- [ ] `Redis::CommandError(<ERR This instance has cluster support disabled>)`

antirez:
Long term... I would love to see redis-rb to get full support for Redis Cluster, perhaps starting from the POC I wrote here: https://github.com/antirez/redis-rb-cluster. I'm sure the same implementation, a bit polished and documented, would receive far more PRs/attention if part of Redis-rb.

supercaracal:
doc is here http://www.rubydoc.info/github/supercaracal/redis-rb/add_redis_cluster_support/Redis/Cluster

supercaracal:
```
Finished in 43.487909485 seconds.
------
620 tests, 5423 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
------
14.26 tests/s, 124.70 assertions/s

make stop
make[1]: Entering directory `/home/travis/build/redis/redis-rb'
(test -f tmp/redis.pid && (kill $(cat tmp/redis.pid) || true) && rm -f tmp/redis.pid) || true
make[1]: Leaving directory `/home/travis/build/redis/redis-rb'
make stop_cluster
make[1]: Entering directory `/home/travis/build/redis/redis-rb'
wget https://github.com/antirez/redis/archive/unstable.tar.gz -O tmp/redis-unstable.tar.gz
--2017-10-02 05:10:58--  https://github.com/antirez/redis/archive/unstable.tar.gz
Resolving github.com (github.com)... 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```
Test cases were successful.

supercaracal:
MEMO:
https://github.com/go-redis/redis/blob/v6.7.1/cluster.go#L637

samuelebistoletti:
Hi,
any plans to merge this? Is it production ready? I would like to use this in our Redis cluster.

Thanks

supercaracal:
@samuelebistoletti I think `Redis::Distributed` is original implementation. It looks like it is different from the Redis Cluster and Sentinel.

samuelebistoletti:
thanks. Do you think it's safe using your implementation of redis cluster in production? Or you are still testing it?

supercaracal:
@samuelebistoletti I think it's safe. I'm sure the same implementation as [POC](https://github.com/antirez/redis-rb-cluster). But owners seem [busy](https://github.com/redis/redis-rb/pull/716#issuecomment-332478557).

badboy:
I won't review it as I currently have neither the time or energy to review or maintain this.

supercaracal:
@badboy I understand.

supercaracal:
@djanowski @pietern @soveran @yaauie Excuse me. Could anyone start review when you're available? I am sorry to bother you while you are busy. Is there anything that I can do?

synth:
If Redis-rb maintainers are not down to support cluster (understandably), what about creating a fork?

filiptepper:
Perhaps this could be shipped as a separate gem?

supercaracal:
> would receive far more PRs/attention if part of Redis-rb.

I think it would be better a part of redis-rb than a separate gem for the reasons above.

byroot:
I can manage some time for it, the problem is more than I don't have experience with Redis cluster, so I'll have to gather lots of context before I can even start to review.

byroot:
But yes, at first sight I'd like this merged.

lib/redis/cluster.rb
```diff
@@ -0,0 +1,467 @@
...
+    # @param node [Redis] the instance of node client
+    # @param command [String, Symbol] the command of redis
+    # @param ttl [Integer] the limit of count for retry or redirection
+    #
+    # @return [Object] depends on the command
+    def try_cmd(node, command, *args, ttl: RETRY_COUNT, &block)
+      ttl -= 1
+      node.send(command, *args, &block)
+    rescue TimeoutError, CannotConnectError, Errno::ECONNREFUSED, Errno::EACCES => err
+      raise err if ttl <= 0
```
byroot: Better to raise without argument here, so that the original backtrace is kept.

lib/redis/cluster.rb
```diff
@@ -0,0 +1,467 @@
...
+      sleep(RETRY_WAIT_SEC)
+      node = find_node || node
+      retry
+    rescue CommandError => err
+      if err.message.start_with?('MOVED')
+        redirection_node(err.message).send(command, *args, &block)
+      elsif err.message.start_with?('ASK')
+        raise err if ttl <= 0
```
byroot: Same raise issue here.

lib/redis/cluster.rb
```diff
@@ -0,0 +1,467 @@
...
+        asking
+        retry
+      else
+        raise err
```
byroot: Same raise issue here.

lib/redis/cluster.rb
```diff
@@ -0,0 +1,467 @@
...
+    #   #=> { host: '127.0.0.1', port: 6379 }
+    #
+    # @param addr [String, Hash] the node address
+    #   e.g. `'redis://127.0.0.1:6379'`, `{ host: '127.0.0.1', port: 6379 }`
+    #
+    # @return [Hash{Symbol => String, Integer}] converted options
+    #
+    # @raise [ArgumentError] if addr is not a `String` or `Hash`
+    def to_client_option(addr)
+      if addr.is_a?(String)
```
byroot: nitpick: You can use a case here:

```ruby
case addr
when String
when Hash
else
end
```

lib/redis/cluster.rb
```diff
@@ -0,0 +1,467 @@
...
+    #   #=> { cluster_state: 'ok', cluster_size: '3' }
+    #
+    # @param response [String] the raw data of cluster info
+    #
+    # @return [Hash{Symbol => String}] parsed data
+    def cluster_info(response)
+      response
+        .split(/[\r\n]+/)
+        .map { |str| str.split(':') }
+        .map { |arr| [arr.first.to_sym, arr[1]] }
```
byroot: What's the rationale for converting to a Symbol?

byroot:
@jrmhaig can you be more specific? What's the version of your redis server & the actual code your are using?

jrmhaig:
I am using the current release candidate of Redis 5.0 from https://redis.io/download (version 4.9.101).

I can do:

```bash
irb(main):001:0> require 'redis'
=> true
irb(main):002:0> r = Redis.new host: 'localhost'
=> #<Redis client v4.0.1 for redis://localhost:6379/0>
irb(main):003:0> r.xadd 'test', '*', 'key', 'value'
=> "1529082632840-0"
```

but if I try this with `Redis::Cluster` instead I get 'unknown command'.

lib/redis/cluster.rb
supercaracal: Because I thought that we are use to using Symbol as a Hash key in Ruby. But I don't have rationales. Should we remove it?

supercaracal:
@jrmhaig I try to learn Redis Streams API on the weekend. Thank you for the information.

supercaracal:
@jrmhaig I fixed at 5fc21cf

supercaracal:
Failed in jruby. 😕 

```
...........Error running mock server: OpenSSL::SSL::SSLError: Received fatal alert: certificate_unknown
org/jruby/ext/openssl/SSLSocket.java:283:in `accept'
/home/travis/.rvm/rubies/jruby-9.1.17.0/lib/ruby/stdlib/jopenssl23/openssl/ssl.rb:405:in `accept'
/home/travis/build/redis/redis-rb/test/support/redis_mock.rb:36:in `block in run'
org/jruby/RubyKernel.java:1316:in `loop'
/home/travis/build/redis/redis-rb/test/support/redis_mock.rb:35:in `run'
/home/travis/build/redis/redis-rb/test/support/redis_mock.rb:26:in `block in start'
```

supercaracal:
@byroot Thank you for your reviewing. I fixed but CI was failure. It seems not related issues. Could you retry build?

lib/redis/cluster.rb
```diff
@@ -0,0 +1,463 @@
...
+    #
+    # @param node [Redis] the instance of node client
+    # @param command [String, Symbol] the command of redis
+    # @param ttl [Integer] the limit of count for retry or redirection
+    #
+    # @return [Object] depends on the command
+    def try_cmd(node, command, *args, ttl: RETRY_COUNT, &block)
+      ttl -= 1
+      node.send(command, *args, &block)
+    rescue TimeoutError, CannotConnectError, Errno::ECONNREFUSED, Errno::EACCES => err
```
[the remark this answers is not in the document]
supercaracal: I ported from [POC](https://github.com/antirez/redis-rb-cluster/blob/master/cluster.rb#L242) . But I couldn't realize those problems. I would like to fix it.

supercaracal:
```
tmp/cache/redis-4.0/src/redis-trib.rb:127:in `load_info': undefined method `split' for #<Array:0x0000000001455488> (NoMethodError)
```

- [ ] fix

supercaracal:
@byroot  Thank you for reviewing. I fixed the call order and others.

lib/redis/cluster.rb
```diff
@@ -0,0 +1,162 @@
...
+      @slot_node_key_maps = build_slot_node_key_maps(available_slots)
+    end
+
+    private
+
+    def respond_to_missing?(method_name, include_private = false)
+      find_node.respond_to?(method_name, include_private)
+    end
+
+    def method_missing(method_name, *args, &block)
```
byroot: Can't we do without `method_missing`? It seems to me that there is relatively few methods that would need to be implemented. I can see `call`, `call_loop`, `call_pipeline`, maybe a couple others.

Even if it's using a macro to do some standard delegation, I'd feel much more confortable with a whitelist of delegated methods rather than `method_missing`.

byroot:
The new architecture looks MUCH better though.

supercaracal:
- [ ] Add more test cases
- [ ] Use whitelisted delegation

supercaracal:
- [ ] Fix low-frequency timeout error in CI
- [ ] Fix key extraction logic in multiple key includable command
- [ ] Fix node selecting logic in transaction command (related to pipelining)
- [ ] Add test cases of failover and resharding and cluster down pattern
- [ ] Add test cases of `TODO`

supercaracal:
@byroot I added many test cases for cluster mode. And then I found so many bugs. I apologize for this PR of bad quality. I fixed them. Could you start again to review?

test/cluster_commands_on_strings_test.rb
```diff
@@ -0,0 +1,169 @@
+# frozen_string_literal: true
+
+require_relative 'helper'
+
+# ruby -w -Itest test/cluster_commands_on_strings_test.rb
+# @see https://redis.io/commands#string
+class TestClusterCommandsOnStrings < Test::Unit::TestCase
+  include Helper::Cluster
+
+  def test_append
```
byroot: It seems to me that many of these added teste are not really specific to cluster mode.

Some do assert on the `CROSSSLOT Keys` error, but many would work just fine on a regular Redis.

There is already shared suites applied on both `Redis` and `Redis::Distributed`, see `Lint::String` for example. 

Using these shared tests should allow to reduce the number of added tests, as well as reduce the duplication, so less risks of forgetting about clusters when new commands etc are added in the future.

lib/redis/cluster/constants.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class Redis
+  class Cluster
+    DEFAULT_SCHEME = 'redis'
+    SECURE_SCHEME = 'rediss'
```
byroot: It's probably not worth to put these in a file of their own.

byroot:
Also, I think at this point you can squash all the history :)

supercaracal:
- [ ] DRY test suite. Use `LINT::`
- [ ] Fix timeout error on CI
- [ ] Remove constants file

supercaracal:
@byroot Thank you for your reviewing. I fixed them. I think I've learned a lot.

byroot:
I'll do a final round of review, but this looks very good from a quick glance.

@jrmhaig any change you could give this branch another go, some more real world experience would be invaluable.

lib/redis/cluster.rb
```diff
@@ -0,0 +1,272 @@
...
+        @node.call_all(command, &block).reject(&:empty?).map { |e| Hash[*e] }
+             .reduce({}) { |a, e| a.merge(e) { |_, v1, v2| v1 + v2 } }
+      when 'numpat' then @node.call_all(command, &block).reduce(:+)
+      else assign_node(command).call(command, &block)
+      end
+    end
+
+    # @see https://redis.io/topics/cluster-spec#redirection-and-resharding
+    #   Redirection and resharding
+    def try_send(node, method_name, *args, ttl: RETRY_COUNT, &block)
```
byroot: Maybe it's just me but TTL makes me thing of a duration, not a number of retries. It's true that it's used in this sense for TCP packets (TTL as number of hops), but in Redis that term is used for a number of seconds before expiration.

So it's a tiny bit confusing IMO.

lib/redis/cluster.rb
```diff
@@ -0,0 +1,272 @@
...
+      raise if ttl <= 0
+      sleep(RETRY_WAIT_SEC)
+      update_cluster_info!
+      node = @node.sample
+      retry
+    rescue CommandError => err
+      if err.message.start_with?('MOVED')
+        assign_redirection_node(err.message).public_send(method_name, *args, &block)
+      elsif err.message.start_with?('ASK')
+        raise if ttl <= 0
```
byroot: It seems like we don't decrement `ttl` in `rescue CommandError`.

lib/redis/cluster/slot.rb
```diff
@@ -0,0 +1,45 @@
...
+    class Slot
+      def initialize(available_slots, node_flags = {}, with_replica = false)
+        @map = build_slot_node_key_map(available_slots, node_flags, with_replica)
+      end
+
+      def exists?(slot)
+        @map.key?(slot)
+      end
+
+      def not_exists?(slot)
```
byroot: Not sure this is very useful, I'd rather see `unless slot.exist?(...)` than `if slot.not_exist(...)`

lib/redis/cluster/slot.rb
```diff
@@ -0,0 +1,45 @@
...
+        !exists?(slot)
+      end
+
+      def find_node_key(slot)
+        @map[slot].to_a.sample
+      end
+
+      def put(slot, node_key)
+        @map[slot] = Set.new unless @map.key?(slot)
```
byroot: Wouldn't `@map[slot] ||= Set.new` be better?

lib/redis/cluster/slot.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+
+      private
+
+      def build_slot_node_key_map(available_slots, node_flags, with_replica)
+        available_slots.each_with_object({}) do |(node_key, slots), acc|
+          next if !with_replica && node_flags[node_key] == 'slave'
+
+          slots.each do |slot|
+            acc[slot] = Set.new unless acc.key?(slot)
```
byroot: Same here. `||=` would be simpler.

supercaracal:
- [ ] Fix them

supercaracal:
@byroot I fixed them all. I removed connection error handling and changed naming of `ttl` into `retry_count` .

On another subject, I'm not sure yet that whether this client can work speedy on thousand of nodes or not. Some commands send to multiple nodes via `Node#call_all` , `Node#call_master` and `Node#call_slave` . It is O(n).

supercaracal:
https://travis-ci.org/redis/redis-rb/jobs/403385003

A build was terminated. The others were successful. Could you retry CI? I am sorry about that.

byroot:
> Some commands are sent to multiple nodes via Node#call_all , Node#call_master and Node#call_slave . Those take O(n) time.


For the `call_all` commands, from what I see it's mostly administrative ones like flushdb etc. So that's less of a problem.

For the `call_slave`, it's similar. `KEYS` won't ever be fast, and no reasonable people will call it in production.

For `call_master`, well we indeed should have a better way to get the master node than to iterate over all the nodes.

supercaracal:
- [ ] survey

tubbo:
@supercaracal @byroot in redis-store, we have an `:avoid_multi_commands` option that users of Redis Cluster can enable in order to avoid those commands from being used. Instead, we pipeline the query. https://github.com/redis-store/redis-store/blob/master/lib/redis/store/ttl.rb#L36-L39

supercaracal:
MEMO:

* https://godoc.org/github.com/go-redis/redis#Tx
* https://godoc.org/github.com/go-redis/redis#ClusterClient.TxPipeline

supercaracal:
In case of Node:

https://github.com/luin/ioredis/tree/622975d9d454eca6a9c495d35f5638d68e2662f2#transaction-and-pipeline-in-cluster-mode

> You can't use multi without pipeline (aka `cluster.multi({ pipeline: false })` ). This is because when you call `cluster.multi({ pipeline: false })` , ioredis doesn't know which node the `multi` command should be sent to.

supercaracal:
In case of PHP (extension):

https://github.com/phpredis/phpredis/blob/develop/cluster.markdown#transactions

> When you call RedisCluster->multi(), the cluster is put into a MULTI state, but the MULTI command is not delivered to any nodes until a key is requested on that node.

lib/redis/cluster.rb
```diff
@@ -0,0 +1,272 @@
...
+    end
+
+    # @see https://redis.io/topics/cluster-spec#redirection-and-resharding
+    #   Redirection and resharding
+    def try_send(node, method_name, *args, ttl: RETRY_COUNT, &block)
+      ttl -= 1
+      node.public_send(method_name, *args, &block)
+    rescue CannotConnectError, ConnectionError => err
+      raise if ttl <= 0
+      sleep(RETRY_WAIT_SEC)
```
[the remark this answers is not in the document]
supercaracal: I think so too. I deleted connection error handling but client can't reload cluster information when we operate node deletion in production.

supercaracal:
The transaction which executes on multiple nodes is not reliable. I think cluster client should raise `AmbiguousNodeError` when `MULTI` `EXEC` `DISCARD` call without Ruby block (pipelining). May I modify?

byroot:
> May I fix it?

Of course. Don't feel like you should not change things because I approved.

supercaracal:
I fixed as below.

- [x] Raise `AmbiguousNodeError`
- [x] Send `Pub/Sub` command to random node (Add to keyless commands) [spec](https://redis.io/topics/cluster-spec#publishsubscribe)
- [x] Use `redis-cli` instead of `redis-trib.rb`

```
$ make trib_cluster 
yes yes | bundle exec ruby tmp/cache/redis-unstable/src/redis-trib.rb create --replicas 1 127.0.0.1:7000 127.0.0.1:7001 127.0.0.1:7002 127.0.0.1:7003 127.0.0.1:7004 127.0.0.1:7005
WARNING: redis-trib.rb is not longer available!
You should use redis-cli instead.

All commands and features belonging to redis-trib.rb have been moved
to redis-cli.
In order to use them you should call redis-cli with the --cluster
option followed by the subcommand name, arguments and options.

Use the following syntax:
redis-cli --cluster SUBCOMMAND [ARGUMENTS] [OPTIONS]

Example:
redis-cli --cluster create 127.0.0.1:7000 127.0.0.1:7001 127.0.0.1:7002 127.0.0.1:7003 127.0.0.1:7004 127.0.0.1:7005 --cluster-replicas 1

To get help about all subcommands, type:
redis-cli --cluster help

makefile:67: recipe for target 'trib_cluster' failed
make: *** [trib_cluster] Error 1
```

supercaracal:
@byroot Are there any other concerns?

supercaracal:
- [ ] Use https://redis.io/commands/command

supercaracal:
I fixed as below.

- [x] Use [COMMAND](https://redis.io/commands/command) instead of hard coding handling in key extraction.
- [x] Fix a matter that client should send to master/slave exactly each of writable/readonly commands.

supercaracal:
I fixed as below.

- [x] Use [COMMAND](https://redis.io/commands/command) instead of hard coding handling in key extraction.
- [x] Fix a matter that client should send writable/readonly commands to master/slave exactly.

supercaracal:
I fixed as below.

- [x] Use [COMMAND](https://redis.io/commands/command) instead of hard coding handling in key extraction.
- [x] Fix a matter that client should send writable/readonly commands to master/slave exactly.

supercaracal:
I think our Redis Cluster client is ready for shipping.

byroot:
Really looks good. I'd have another nitpick ;)

`Redis::Cluster::CrossSlotPipeliningError` could use a default `message` so that's it's easier for users to understand what's wrong.

After that I'd like to merge and if it requires further improvements it might as well happen in followup PRs.

I'll also see if I can get publish access so that we can release a RC to get some feedback from users.

supercaracal:
- [ ] Add default error message to `Redis::Cluster::CrossSlotPipeliningError`

supercaracal:
@byroot I have fixed it. I am grateful for your support. Thank you so much.

byroot:
This was released as `4.1.0.beta1`, if you are interested in this feature, please try to use it a report any issues you might spot.

xfalcox:
Hey @byroot amd @supercaracal, I'm trying to wrap around my head about using this in production on AWS.

When we setup a new High Avaliable Redis Cluster in AWS we get a single endpoint, which AWS calls `configuration_endpoint_address`, and is documented in Terraform [here](https://www.terraform.io/docs/providers/aws/r/elasticache_replication_group.html#configuration_endpoint_address).

I can connect to this endpoint using `redis-cli -h $configuration_endpoint_address -c` and issue commands like `SET` and `GET` just fine, and I'll be routed to a shard.

That said, by reading this PR code, I'm supposed to pass an array of Redis hosts instead of a single one when establishing a connection. Are AWS and redis-rb approachs different, incompatible or I'm reading something wrong?

supercaracal:
We're able to specify single node to this gem.

```
[1] pry(main)> cli = Redis.new(cluster: %w[redis://127.0.0.1:7000])
=> #<Redis client v4.1.0.beta1 for redis://127.0.0.1:7000/0 redis://127.0.0.1:7001/0 redis://127.0.0.1:7002/0>
[2] pry(main)> cli.get :key1
=> nil
[3] pry(main)> cli.get :key2
=> nil
[4] pry(main)> cli.get :key3
=> nil
```

It seems the `redis-cli` redirect to each nodes internally too.
https://github.com/antirez/redis/blob/5.0-rc5/src/redis-cli.c#L1025-L1053

xfalcox:
That's amazing @supercaracal. I just tried pointing the new gem to an AWS Elasticache configuration_endpoint_address and looks like the redirect is handled automatically. Thanks.

kitsirota:
@byroot when do you think the 4.1.0.beta1 release can get cut?  I didnt see it as a milestone or branch/tag in this repo yet.

Thanks!

byroot:
It's already been cut over a month ago: https://rubygems.org/gems/redis/versions/4.1.0.beta1

PR #725, opened 2017-10-18
Improve unix urls

ccutrer:


ccutrer:
Thinking more about this, I don’t like the query being the db. It should be regular key/value pairs instead. That would mean being able to specify other params (like the id, for using in a ring), and we could add that to regular redis URIs too. I’ll work on that tomorrow.

ccutrer:
I've update the PR to be more generic, and to _not_ change the behavior of `Connection#id` (realizing now that that could break the layout of existing hash rings).

byroot:
Closing as stale.

Issue #698, opened 2017-05-30
Redis::exists should no longer use Boolify


scottyschup:
In the [Redis documentation for `exists`](https://redis.io/commands/exists), as of Redis 3.0.3 `exists` may take multiple key names and returns a count of the number of keys that exist.

Using the `redis-rb` gem v3.3.3, the result of [`Redis::exists`](https://github.com/redis/redis-rb/blob/19bbbc8dc1b75cfbaf3dd7e0a92a9ff72f474fb5/lib/redis.rb#L504) with an array of keys returns `true` *if and only if* just one of the keys exist. `false` is returned if none of the keys exist **or if multiple keys do** because `Boolify` looks for a non-nil value equal to 1.

To reproduce:
```ruby
r = Redis.new
r.mset 'test1', 1, 'test2', 2
r.exists ['test1', 'nonexistant']
#=> true
r.exists ['test1', 'test2']
#=> false
```
So either `Boolify` could be adjusted to look for values `>= 1`, or (what makes more sense to me) `Boolify` should not be used in this function so that the return value is the count of the number of keys found as is described in the Redis docs.

kumekay:
I've created a new pull request address to fix this issue. 
It differs from #728 that it doesn't change return type of this method. It always return boolean.
true if all passed keys exist in database, and false if at least one is missing.

https://github.com/redis/redis-rb/pull/736

kumekay:
If there is a real demand for number of existed keys, it can be implemented as separate method, `exists_count`, for example

vgmoose:
Any update on the status of this? It looks like there are three in-progress PRs for providing functionality to accept multiple keys:

https://github.com/redis/redis-rb/pull/728 by @ybinzu - uses `exists`, returns the count if multiple keys are provided, nil if 0 are found
https://github.com/redis/redis-rb/pull/736 by @kumekay - uses `exists`, returns true if all provided keys exist, false if one doesn't
https://github.com/redis/redis-rb/pull/741 by @ericproulx - adds `mexists`, returns the count of how many provided keys exist (and never nil)

Thanks to all involved and sorry for any unwanted pings

micahbf:
Just ran into this issue as well.
Can I do anything to move this forward?

scottyschup:
Because this issue was not resolved by the time I needed it to be, I just ended up writing a helper method in my `RedisClient` class to use in place of `Redis#exists`. It was something like:
```rb
class RedisClient
...
  @client = Redis.new
...
[...]

PR #728, opened 2017-12-01
Change #exist behaviour for multiple keys #698

ybinzu:
This is attempt to solve #698.

For single key argument method `#exist` works as usual - it returns true or false.

But when it has multiple key arguments then it works almost like it described in [redis documentation](https://redis.io/commands/exists). Except for 0 value then method returns nil. Consider this tests:

``` ruby
assert_equal false, r.exists("foo")

r.set("foo", "s1")

assert_equal true,  r.exists("foo")

r.set("foo1", "s2")

assert_equal 2,  r.exists(["foo", "foo1"])
assert_equal 1,  r.exists(["foo", "foo5"])
assert_equal nil, r.exists(["foo7", "foo5"])
```

0 to nil conversion added just in case of backward compatibility with exist codebases (nil as false comparison and not nil as true).

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -530,6 +530,10 @@ class Redis
   # @return [Boolean]
   def exists(key)
     synchronize do |client|
+      if key.is_a?(Enumerable)
+        return client.call([:exists, key], &BoolifyExistEnum)
+      end
+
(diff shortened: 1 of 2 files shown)


byroot:

PR #731, opened 2017-12-20
B*pop read timeout

Yurokle:
Attempt to solve this issue: https://github.com/redis/redis-rb/issues/732

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,8 +224,8 @@ See lib/redis/errors.rb for information about what exceptions are possible.
 
 ## Timeouts
 
-The client allows you to configure connect, read, and write timeouts.
-Passing a single `timeout` option will set all three values:
+The client allows you to configure connect, read, _`bpop`_ read, and write timeouts. 
+Passing a single `timeout` option will set all four values:
 
 ```ruby
 Redis.new(:timeout => 1)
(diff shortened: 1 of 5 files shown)


Yurokle:
Attempt to solve this issue: https://github.com/redis/redis-rb/issues/732

badboy:
We already have far too many confusing timeouts. Why do we need another again? I don’t fully get the reasoning and it sounds more like an issue with what the server does.

Yurokle:
@badboy it's totally a server-side issue, yeah. But it's a known problem and we can make client work smoothly with it, instead of throwing timeout exceptions.
I gave more details here: https://github.com/redis/redis-rb/issues/732

In short, on older redis servers `redis.brpop('asdf', timeout: 1)` and similar blocking operations might take more than 2 seconds, which is a problem if you're using `1.second` timeout.
In order to solve the problem, one can either update to newer redis server versions, which might take months in big companies, or we can make redis client work with this known problem.
Btw, another approach of solving this would be to use something like `bpop_timeout_delta` param, which would default to `0` – the delta would be added to `b*pop` timeouts in addition to other timeouts.

Both approaches are invisible to those who don't need to use them.

badboy:
So just so I understand it correctly, setting a timeout of 2 seconds on the connection and 1 second on the `brpop` would then also correctly work?

Yurokle:
Yep, 2 seconds timeout on the connection works fine, except for some people might want 1.second timeout, because Redis is usually super fast and it's better to fail fast when something goes wrong.

Here is an example:
I have a high-loaded web app that processes many QPS (many web workers, one redis server), and within the HTTP call I'm using Redis for scheduling some Sidekiq jobs (no `b*pop` operations happen here). If something happens to Redis and my requests start to timeout within the lifecycle of my HTTP request, I don't want my HTTP queue to grow, because the HTTP latency jumps from 50ms to 2050ms, since it needs to wait for a redis timeout. If I can timeout in 1.second vs 2.seconds, it means I can process 2X requests per unit of time if I get a spike of requests (which you usually get when something goes wrong).

On the Sidekiq server side, where we make `b*pop` calls to read from sidekiq queue, we don't usually care too much, and also `b*pop` technically happens outside of sidekiq job lifecycle. So it should be fine to have a separate bigger timeout for those calls, that doesn't affect behavior of the rest of application.

badboy:
So if you have separate clients anyway, this can be mitigated by different timeouts for different clients.
I really don't like adding a new special timeout for one now-fixed server issue in an old version of this library.

Yurokle:
@badboy yeah that makes sense, I totally understand you. And actually it's a good idea, I probably just gonna bump Sidekiq server redis client timeouts until we migrate to newer Redis server version.
Thanks!

PR #735, opened 2018-01-12
Fix tests

TheSmartnik:
The problem was caused by [this line](https://github.com/rubygems/rubygems/blob/v2.7.4/lib/rubygems/bundler_version_finder.rb#L105) introduced in rubygems 2.7.0. The problem is fixed in master, but not in any release yet.

Downgrading rubygems to version prior to 2.7.0 fixes the problem. I've tested it locally and on [travis-ci](https://travis-ci.org/TheSmartnik/redis-rb/builds/328176592).

--- 

Also trying to fix JRuby specs with fix discussed in https://github.com/travis-ci/travis-ci/issues/9049

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,30 +42,30 @@ branches:
 matrix:
   exclude:
     # hiredis
-    - rvm: jruby-9
+    - rvm: jruby-9.1.15.0
       gemfile: .travis/Gemfile
       env: DRIVER=hiredis REDIS_BRANCH=3.0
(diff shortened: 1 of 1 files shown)


badboy:
Great, expect me to merge as soon as it’s run through

badboy:
Hmpf, jruby tests fail, probably because of some weird corner case of the ipv6 setup. I cannot reproduce it locally. gonna merge it anyway.

PR #736, opened 2018-01-13
Update non-distributed `exists` command to accept multiple keys. Fix #698

### LINKED ISSUE #698: Redis::exists should no longer use Boolify

kumekay:
Since Redis 3.0.3 supports EXISTS command may accept multiple keys.

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -524,13 +524,13 @@ class Redis
     end
   end
 
-  # Determine if a key exists.
+  # Determine if all keys exists.
   #
-  # @param [String] key
+  # @param [String, Array<String>] key
   # @return [Boolean]
-  def exists(key)
+  def exists(*keys)
     synchronize do |client|
-      client.call([:exists, key], &Boolify)
+      client.call([:exists] + keys) == keys.size
     end
   end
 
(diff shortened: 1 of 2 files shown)


byroot:

PR #739, opened 2018-01-23
improve configurability of HashRing

ccutrer:
allow different digests, and pass through replicas and digest config
when creating a Distributed redis client


ccutrer:
Ping?

byroot:
Closing as stale. Either way I think the `ring:` parameter is enough, users can pass their own ring implementation, which is good enough for adavanced usages and much more flexible than passing tons of parameters.

PR #741, opened 2018-02-14
Adding exists_count

ericproulx:
Since Redis 3.0.3 it is possible to specify multiple keys for the command `exists` instead of a single one. In such a case, it returns the total number of keys existing

My work is based on issue #698 comments


mperham:
I would name it `mexists` to match `mget`.

byroot:

PR #744, opened 2018-03-13
Add Errno::EADDRNOTAVAIL to system call errors

spk:
Rescue low-level exception `Errno::EADDRNOTAVAIL` (Address not available) to have a `CannotConnectError` instead

Cheers

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -341,6 +341,7 @@ class Redis
       @pending_reads = 0
     rescue TimeoutError,
            SocketError,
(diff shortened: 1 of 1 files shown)


hugopl:
I get this when running redis inside docker, instead of an usual redis exception.

byroot:
Would you mind to share a backtrace and / or some minimal repro script?

byroot:
Actually nevermind. That PR makes perfect sense.

spk:
Thanks I don't find the trace anymore I will try to remember that for next time, cheers

PR #747, opened 2018-03-26
Optimize empty transactions

chanks:
A minor optimization - I run a bunch of logic inside a `multi` block that can conditionally execute commands, but if it doesn't, there's no point in spending time sending the empty transaction to Redis.

I realize the test is a bit silly, but I'm not sure offhand how to improve it - happy to take suggestions.

Thanks!

### DIFF
--- a/lib/redis/pipeline.rb
+++ b/lib/redis/pipeline.rb
@@ -22,6 +22,10 @@ class Redis
       @shutdown
     end
 
+    def empty?
+      @futures.empty?
+    end
+
     def call(command, &block)
       # A pipeline that contains a shutdown should not raise ECONNRESET when
       # the connection is gone.
@@ -86,7 +90,11 @@ class Redis
       end
 
       def commands
-        [[:multi]] + super + [[:exec]]
+        if empty?
+          []
+        else
+          [[:multi]] + super + [[:exec]]
+        end
       end
     end
   end
--- a/test/transactions_test.rb
+++ b/test/transactions_test.rb
@@ -114,6 +114,16 @@ class TestTransactions < Test::Unit::TestCase
     assert_equal "s1", r.get("foo")
   end
 
+  def test_empty_multi_exec
+    result = nil
+
+    redis_mock(:exec => lambda { |*_| "-ERROR" }) do |redis|
+      result = redis.multi {}
+    end
+
+    assert_equal [], result
+  end
+
   def test_raise_command_errors_when_accessing_futures_after_multi_exec
     begin
       r.multi do |m|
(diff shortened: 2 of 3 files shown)


test/transactions_test.rb
```diff
@@ -114,6 +114,11 @@ def test_raise_command_errors_in_multi_exec
     assert_equal "s1", r.get("foo")
   end
 
+  def test_empty_multi_exec
+    r.multi do |m|
+    end
+  end
+
```
jeremy: Perhaps we could mock the multi / exec to return an error, causing the test to fail if they're actually invoked, as with `test_raise_command_error_when_exec_fails`

lib/redis/pipeline.rb
```diff
@@ -86,7 +86,13 @@ def finish(replies)
       end
 
       def commands
-        [[:multi]] + super + [[:exec]]
+        a = super
```
jeremy: If there are no `@futures`, we needn't `super` either, huh! Would be nice to have e.g. `Redis::Pipeline#empty?` which delegates to `@futures.empty?` (or `none?`/`any?`…), leaving us with a concise
```ruby
def commands
  if empty?
    []
  else
    [[:multi]] + super + [[:exec]]
  end
end
```

chanks:
Changes made!

chanks:
Pushed a fix for the return value of `multi`.

The test is implemented the same way as `test_raise_command_error_when_exec_fails`, so it seems consistent with the current codebase.

test/transactions_test.rb
```diff
@@ -114,6 +114,18 @@ def test_raise_command_errors_in_multi_exec
     assert_equal "s1", r.get("foo")
   end
 
+  def test_empty_multi_exec
+    result = nil
+
+    redis_mock(:exec => lambda { |*_| "-ERROR" }) do |redis|
+      result =
+        redis.multi do |m|
```
byroot: np: the formatting here is very weird. A one-liner would be better IMO `result = redis.multi {}`

chanks:
Formatting changed and commits squashed.

byroot:
@rafaelfranca you're good with this?

byroot:
I'll merge on green (which might take a while since Travis is backlogged, I'm trying to speedup CI, but that's another story).

rafaelfranca:
You can use bors.

bors r+

rafaelfranca:
Oh. Looks like we are not in the reviewers list for bors.

byroot:
JRuby is failing as usual, but the rest of the jobs are green.

PR #754, opened 2018-05-21
Don't coerce Redis command errors

### LINKED ISSUE #602: `multi` with a block does not return errors in array
Without block, errors are found in the return value (array) from `exec`:

```
> redis.set('akey', 'a string')
=> "OK"
> redis.multi
=> "OK"
> redis.sadd('akey', 'another')
=> false
> redis.incr('akey')
=> "QUEUED"
> redis.exec
=> [#<Redis::CommandError: WRONGTYPE Operation against a key holding the wrong kind of value>, #<Redis::CommandError: ERR value is not an integer or out of range>]
```

When used with a block, the error does not appear in the return value...

```
> redis.multi do |m|
>   m.sadd('akey', 'another')
> end  
=> [false]
```

...or it is just raised (i.e., not returned):

```
> redis.multi do |m|
>   m.incr('akey')
> end  
Redis::CommandError: ERR value is not an integer or out of range
```

The former behavior (without block) looks more consistent, except that `sadd` returns false ("QUEUED" expected); I found some other commands like `setnx` and `expire` also return false when queued.

soylent:
This PR addresses the following issues:

1. Commands that are executed in a transaction and use the `Boolify` or `BoolifySet` transformations coerce Redis command errors to false instead of raising an exception:

```ruby
redis.multi do |multi|
  multi.set('foo', 'bar')
  multi.hexists('foo', 'x') # foo is not a hash
end
# Returns ["OK", false]
# insread of raising
#   `Redis::CommandError: WRONGTYPE Operation against a key holding the wrong kind of value'
```

2. Commands that are executed in a transaction and use the `Hashify` and `Floatify` transformations don't handle Redis command errors.

```ruby
redis.multi do |multi|
  multi.set('foo', 'bar')
  multi.hgetall('foo') # foo is not a hash
end
# NoMethodError: undefined method `each_slice' for #<Redis::CommandError:0x00007fbc73d8f1d0>
```

Resolves #602

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -192,11 +192,7 @@ class Redis
   def config(action, *args)
     synchronize do |client|
       client.call([:config, action] + args) do |reply|
-        if reply.kind_of?(Array) && action == :get
-          Hashify.call(reply)
-        else
-          reply
-        end
+        action == :get ? Hashify.call(reply) : reply
       end
     end
   end
@@ -2852,46 +2848,49 @@ private
   # returning false.
   Boolify =
     lambda { |value|
-      value == 1 if value
+      case value
+      when 1 then true
+      when 0 then false
+      else value
+      end
     }
 
   BoolifySet =
     lambda { |value|
-      if value && "OK" == value
-        true
-      else
-        false
+      case value
+      when "OK" then true
+      when nil then false
+      else value
       end
     }
 
   Hashify =
-    lambda { |array|
+    lambda { |value|
+      return value unless value.respond_to?(:each_slice)
+
       hash = Hash.new
-      array.each_slice(2) do |field, value|
-        hash[field] = value
+      value.each_slice(2) do |key, val|
+        hash[key] = val
       end
       hash
     }
 
   Floatify =
-    lambda { |str|
-      if str
-        if (inf = str.match(/^(-)?inf/i))
-          (inf[1] ? -1.0 : 1.0) / 0.0
-        else
-          Float(str)
-        end
+    lambda { |value|
+      case value
+      when "inf" then Float::INFINITY
+      when "-inf" then -Float::INFINITY
+      when String then Float(value)
+      else value
       end
     }
 
   FloatifyPairs =
-    lambda { |result|
-      if result.respond_to?(:each_slice)
-        result.each_slice(2).map do |member, score|
-          [member, Floatify.call(score)]
-        end
-      else
-        result
+    lambda { |value|
+      return value unless value.respond_to?(:each_slice)
+
+      value.each_slice(2).map do |member, score|
+        [member, Floatify.call(score)]
       end
     }
 
--- a/test/transactions_test.rb
+++ b/test/transactions_test.rb
@@ -102,7 +102,7 @@ class TestTransactions < Test::Unit::TestCase
     assert @info.value.kind_of?(Hash)
   end
 
-  def test_raise_command_errors_in_multi_exec
+  def test_raise_command_errors_when_reply_is_not_transformed
     assert_raise(Redis::CommandError) do
       r.multi do |m|
         m.set("foo", "s1")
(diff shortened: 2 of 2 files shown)


soylent:
CI seems to be failing for unrelated reason

lib/redis.rb
```diff
@@ -2852,46 +2848,49 @@ def method_missing(command, *args)
...
-      if str
-        if (inf = str.match(/^(-)?inf/i))
-          (inf[1] ? -1.0 : 1.0) / 0.0
-        else
-          Float(str)
-        end
+    lambda { |value|
+      case value
+      when "inf" then 1.0 / 0.0
+      when "-inf" then -1.0 / 0.0
```
javierhonduco: Maybe we could use `Float::INFINITY` instead?

soylent: Thanks! Updated

soylent:
@javierhonduco Yes, #733 did not resolve all problems of this kind

chanks:
Here's a bug I just ran into that seems to fall under your second issue. I'm not immediately sure if this PR fixes it, but if so, maybe this makes sense as an additional test case:

```
2.6.0 :001 > require "redis"
 => true 
2.6.0 :002 > redis = Redis.new
 => #<Redis client v4.1.0 for redis://127.0.0.1:6379/0> 
2.6.0 :003 > 
2.6.0 :004 > redis.xgroup :create, "my_stream", "my_group", 0, mkstream: true
 => "OK" 
2.6.0 :005 > 
2.6.0 :006 > redis.multi do |r|
2.6.0 :007 >       r.xpending "my_stream", "my_group", "-", "+", 10
2.6.0 :008?>     r.xpending "my_stream", "not_a_group", "-", "+", 10
2.6.0 :009?>   end
       16: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis.rb:50:in `synchronize'
       15: from /home/chris/.rvm/rubies/ruby-2.6.0/lib/ruby/2.6.0/monitor.rb:230:in `mon_synchronize'
       14: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis.rb:50:in `block in synchronize'
       13: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis.rb:2454:in `block in multi'
       12: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/client.rb:161:in `call_pipeline'
       11: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/client.rb:299:in `with_reconnect'
       10: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/client.rb:163:in `block in call_pipeline'
        9: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/pipeline.rb:84:in `finish'
        8: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/pipeline.rb:60:in `finish'
        7: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/pipeline.rb:60:in `map'
        6: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/pipeline.rb:60:in `each'
        5: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/pipeline.rb:60:in `each_with_index'
        4: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/pipeline.rb:60:in `each'
        3: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/pipeline.rb:61:in `block in finish'
        2: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis/pipeline.rb:122:in `_set'
        1: from /home/chris/.rvm/gems/ruby-2.6.0/gems/redis-4.1.0/lib/redis.rb:3405:in `block in <class:Redis>'
NoMethodError (undefined method `map' for #<Redis::CommandError:0x000055cdf6a95e10>)
Did you mean?  tap
```

byroot:
Thank you.

Issue #626, opened 2016-07-12
HDEL should use the splat (*fields) operator


bradley:
Currently, setting multiple key values for a redis hash using this library, requires the user pass in a comma separated key/value pair for each key and value the user wishes to set in their hash, as follows:

```
$redis.hmset("myhash", "field1", "Hello", "field2", "World")
```

Which closely matches the redis CLI [spec](http://redis.io/commands/hmset):

```
HMSET myhash field1 "Hello" field2 "World"
```

This works because this library's implementation of `hmset` [uses the splat operator](https://github.com/redis/redis-rb/blob/3ae7fbda312caba5668bedd2ecd1ab15ff0d975e/lib/redis/distributed.rb#L659):

```
def hmset(key, *attrs)
  node_for(key).hmset(key, *attrs)
end
```

I believe `hdel` should work the same way.

Currently, this library's implementation [does not use the splat operator](https://github.com/redis/redis-rb/blob/3ae7fbda312caba5668bedd2ecd1ab15ff0d975e/lib/redis/distributed.rb#L682):

```
def hdel(key, field)
  node_for(key).hdel(key, field)
end
```

Which means that requesting the deletion of more than one key in one's hash requires passing in the keys as an array, like so:

```
$redis.hdel("myhash", ["field1", "field2"])
```

When it should, for the sake of consistency, work more like `hmset`:

```
$redis.hdel("myhash", "field1", "field2")
```

My argument for this is that the redis CLI spec for [hmset](http://redis.io/commands/hmset) and for [hdel](http://redis.io/commands/hdel) have consistent behaviour:

```
HMSET myhash field1 "Hello" field2 "World"
HDEL myhash field1 field2
> 2
```

bradley:
I understand this is probably a completely opinion-based thing and that Im just butthurt because I was reading the [Redis guides for hashes](http://redis.io/commands#hash) rather than the redis-rb guides.

PR #755, opened 2018-05-22
Super splat args for HDEL

byroot:
Fix: https://github.com/redis/redis-rb/issues/626

It already supported passing an array of keys, but splat args are much more consistent with other commands like `del`.


byroot:
CI is failing because of https://github.com/redis/redis-rb/issues/756

Issue #610, opened 2016-04-18
Speed up Travis builds


djanowski:
My first idea would be to cache our Redis binaries across versions. Probably need to use Git to pull the branches and rely on `make` to know if anything needs to be rebuilt.

aripollak:
Could building redis at least take advantage of multiple CPUs with `make -j2`? That seems to work well for me locally.

PR #757, opened 2018-05-22
Cache redis server compilation on CI

byroot:
Fix: https://github.com/redis/redis-rb/issues/610

Or at least most of it.

My Makefile foo isn't great, but from what I've seen it's not easy at all to do checksum based caching with it. So instead I moved the build script to a simple ruby script, and do the conditional build there.

Ultimately there is not much left in the makefile, so this PR is a bit questionable, but let's see if it works first.

The cache should be scoped by branch & by job (each combination of the matrix has it's own cache). The cache should invalidated everytime a Redis server release change.I don't know how often it happens for `unstable`, but for others it should be fairly rare.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: ruby
-
+cache:
+    directories:
+      - tmp/cache
 before_install:
   - gem update --system 2.6.14
   - gem --version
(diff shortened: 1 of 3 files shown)


byroot:
The results seem good

Job with a cache hit (`1m36s`): https://travis-ci.org/redis/redis-rb/jobs/382384621
Job with a cache miss (`2m50s`): https://travis-ci.org/redis/redis-rb/jobs/382384622

@rafaelfranca the script is a bit ugly and could be perfected, but I think it's a nice win already. WDYT ?

Issue #756, opened 2018-05-22
Flaky test in JRuby test suite


byroot:
The JRUby test suite seem to fail on a regular basis: https://travis-ci.org/redis/redis-rb/jobs/382308731

```ruby
     16:     def test_unverified_ssl_connection
  => 17:       assert_raise(OpenSSL::SSL::SSLError) do
     18:         RedisMock.start({ :ping => proc { "+PONG" } }, ssl_server_opts("untrusted")) do |port|
     19:           redis = Redis.new(:port => port, :ssl => true, :ssl_params => { :ca_file => ssl_ca_file })
     20:           redis.ping
/home/travis/build/redis/redis-rb/test/ssl_test.rb:17:in `test_unverified_ssl_connection'
<OpenSSL::SSL::SSLError> expected but was
<java.lang.NullPointerException>
diff:
? O                        pe  nSSL::SSL::SSLError
? java.lang.NullPointerExce tio                   
Failure: test_unverified_ssl_connection(SslTest)
```

I know very little about JRuby so I don't quite know where to start to fix this, but this flaky test is causing havoc on CI.

cc @rafaelfranca @mperham maybe one of you have some knowledge of JRuby?

rafaelfranca:
I don't, but that looks like a bug in JRuby. Maybe upgrading to 9.1.17 could get it more stable?

PR #759, opened 2018-05-22
Update JRuby to 9.1.17

byroot:
Ref: https://github.com/redis/redis-rb/issues/756

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ rvm:
   - 2.3.3
   - 2.4.1
   - 2.5.0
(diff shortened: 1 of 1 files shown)


rafaelfranca:
No luck. Maybe @headius have an idea?

byroot:
No luck indeed. However I went a bit further and fixed the `exclude` directive in `travis.yml`.

There was duplication, and also some directives were not matching because of alignement issues.

So JRuby is still flaky, but at least now there is only 3 JRuby jobs like it was intended instead of 7.

PR #760, opened 2018-05-24
Fix testing of redis branches

byroot:
While trying to figure out further optimizations, I noticed that jobs with `REDIS_BRANCH=3.0` were actually building `unstable`.

I though I broke it in #757, but actually looking at older builds the issue was already here, e.g.: https://travis-ci.org/redis/redis-rb/jobs/280892409

```bash
export REDIS_BRANCH=3.0
#....
wget https://github.com/antirez/redis/archive/unstable.tar.gz -O tmp/redis-unstable.tar.gz
```

Again my makefile foo isn't great, but this change seem to work.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,5 +1,5 @@
 TEST_FILES   := $(shell find test -name *_test.rb -type f)
-REDIS_BRANCH := unstable
+REDIS_BRANCH ?= unstable
 TMP          := tmp
 BUILD_DIR    := ${TMP}/cache/redis-${REDIS_BRANCH}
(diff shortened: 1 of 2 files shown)


byroot:
Hum. It seems like there is 12 broken tests with `REDIS_BRANCH=3.0`, I'll see if I can fix them.

byroot:
I marked the tests as skipped for Redis < 3.2.0.

rafaelfranca:
Oh. Looks like unsable is now failing.

ajlai:
Perhaps we should avoid building the unstable branch or move it to allowed failures?
It looks like unstable failing is also happening in redisci: http://ci.redis.io/run_66863.html

PR #761, opened 2018-05-25
Fix failing jruby tests, change default REDIS_BRANCH to 4.0

ajlai:
Fix: https://github.com/redis/redis-rb/issues/756

~~jruby tests fail when using jruby_openssl 0.10.0 but not when using 0.9.21
This change adds 0.9.21 to the travis build matrix and sets 0.10.0 as an allowed failure.~~

~~This change also switches the default REDIS_BRANCH to stable since
the unstable branch of redis isn't always in a buildable state.~~

This change adds a Gemfile dependency for jruby_openssl <0.10.0

This change also switches the default REDIS_BRANCH to 4.0 since
the unstable branch of redis isn't always in a buildable state.

Also adds minor tweaks to how the tests are run due to minor issues with Gemfile / bundle usage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ env:
     - DRIVER=hiredis   REDIS_BRANCH=3.2
     - DRIVER=synchrony REDIS_BRANCH=3.0
     - DRIVER=synchrony REDIS_BRANCH=3.2
-    - DRIVER=ruby      REDIS_BRANCH=unstable
+    - DRIVER=ruby      REDIS_BRANCH=4.0
 
 branches:
   only:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -42,21 +42,33 @@ branches:
...
+    - rvm: jruby-9.1.17.0
+      env: DRIVER=ruby REDIS_BRANCH=3.2 JRUBY_OPENSSL=0.9.21
+    - rvm: jruby-9.1.17.0
+      env: DRIVER=ruby REDIS_BRANCH=stable JRUBY_OPENSSL=0.9.21
+  allow_failures:
+      # Using jruby-openssl 0.10.0, we get NPEs in jruby tests: https://github.com/redis/redis-rb/issues/756
+      # jruby-openssl 0.9.21 required tests have been added above in the include section.
+    - rvm: jruby-9.1.17.0
+      env: DRIVER=ruby      REDIS_BRANCH=3.0
+    - rvm: jruby-9.1.17.0
```
byroot: I don't think it's worth running test we know will fail. If the fix for the jruby test suite is to use `jruby-openssl < 0.10` then let's add this to the Gemfile / gemspec and that's it.

.travis.yml
```diff
@@ -33,7 +33,7 @@ env:
     - DRIVER=hiredis   REDIS_BRANCH=3.2
     - DRIVER=synchrony REDIS_BRANCH=3.0
     - DRIVER=synchrony REDIS_BRANCH=3.2
-    - DRIVER=ruby      REDIS_BRANCH=unstable
+    - DRIVER=ruby      REDIS_BRANCH=stable
```
byroot: In think the intent of building `unstable` was to test `4.0`. The build matrix could simply be `3.0 3.2 4.0`.

PR #763, opened 2018-05-28
Fix client ID scheme

Looooong:
This fix `Redis::Client#id` to display the scheme being used correctly. It used to show **redis** scheme regardless of any other scheme we use (rediss, unix...)

### DIFF
--- a/test/connection_test.rb
+++ b/test/connection_test.rb
@@ -22,8 +22,8 @@ class TestConnection < Test::Unit::TestCase
   end
 
   def test_default_id_with_host_and_port_and_explicit_scheme
-    redis = Redis.new(OPTIONS.merge(:host => "host", :port => "1234", :db => 0, :scheme => "foo"))
-    assert_equal "redis://host:1234/0", redis.connection.fetch(:id)
+    redis = Redis.new(OPTIONS.merge(:host => "host", :port => "1234", :db => 0, :scheme => "rediss"))
+    assert_equal "rediss://host:1234/0", redis.connection.fetch(:id)
   end
 
   def test_default_id_with_path
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -107,7 +107,7 @@ class Redis
     end
 
     def id
-      @options[:id] || "redis://#{location}/#{db}"
+      @options[:id] || "#{@options[:path] ? 'redis' : scheme}://#{location}/#{db}"
     end
 
     def location


test/connection_test.rb
```diff
@@ -23,17 +23,17 @@ def test_default_id_with_host_and_port
...
   def test_default_id_with_host_and_port_and_explicit_scheme
     redis = Redis.new(OPTIONS.merge(:host => "host", :port => "1234", :db => 0, :scheme => "foo"))
-    assert_equal "redis://host:1234/0", redis.connection.fetch(:id)
+    assert_equal "foo://host:1234/0", redis.connection.fetch(:id)
   end
 
   def test_default_id_with_path
     redis = Redis.new(OPTIONS.merge(:path => "/tmp/redis.sock", :db => 0))
-    assert_equal "redis:///tmp/redis.sock/0", redis.connection.fetch(:id)
+    assert_equal "unix:///tmp/redis.sock/0", redis.connection.fetch(:id)
```
byroot: Actually I'm not sure this is correct:

```bash
$ redis-cli -u redis:///tmp/redis.sock GET foo
(nil)
$ redis-cli -u unix:///tmp/redis.sock GET foo
Invalid URI scheme
```

I'd need to double check, but I'm pretty sure the `scheme` define the protocol used, not the type of socket etc.

Can we step back a bit? You are trying to improve support for `rediss://` ?

In this case we should probably coerce the scheme better first. 

I also can't find anything about the `unix://` scheme: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

Looooong: Yes, I am trying to improve support for `rediss`.

And, I'm not sure about the `unix` part. You can tell me more on what to do about it.

byroot: My point was that `unix://` doesn't make sense because the `scheme` part of the URL (also known as `protocol`), is here to tell the client how to communicate with the resource.

There is no such thing as a `unix://` protocol, you are actually using the `redis://` or `rediss://` protocol over a unix socket.

So IMO: `Redis.new(path: '/tmp/redis.sock').id` should be `"redis:///tmp/redis.sock"`.

Does that makes sense to you?

Looooong: I understand. Then, isn't it a semantic problem because `unix` is stored in instance variable named `@scheme`?

byroot: It's not semantic no.

It's that `unix://` is simply not a thing (AFAICT), so I believe we should coerce it to `reids://` here: https://github.com/redis/redis-rb/blob/bc80cb14fbaa06165cb646ca31afb2f33770a48c/lib/redis/client.rb#L409

Looooong: I think that we should coerce it here:

https://github.com/redis/redis-rb/blob/bc80cb14fbaa06165cb646ca31afb2f33770a48c/lib/redis/client.rb#L432

Looooong:
@byroot I updated the commit. You can check it out.

byroot:
`scheme` is no longer a thing in 5.0, closing.

PR #766, opened 2018-07-05
Support UNLINK command interface for redis v4

ksss:
I implemented interface for UNLINK command available since 4.0.0.
https://redis.io/commands/unlink


byroot:
Thank you.

PR #710, opened 2017-07-31
Back off mechanism for reconnection attempts.

wpp:
This is a first take on a simple back off algorithm for reconnection
attempts. I've previously mentioned this in #598
but didn't get any feedback. Therefore I haven't changed much to the
solution proposed there.
A possible introduction of a back off mechanism has also been mentioned
in #347.

The patch changes the reconnection behaviour to be "less aggressive"
by sleeping for a period of time. The sleep period increases
exponentially, starting at 0.2 seconds and ending at 12.8 seconds
(this maximum can be configured by supplying a `reconnect_max_timeout`
option to the client.)

I've recorded a quick video of how this looks like when
starting/stopping a redis container: https://youtu.be/Z4HuZmwYCZM

Hope I didn't overlook any major points.

Looking forward to your feedback,
thanks for a great library.

PR #778, opened 2018-08-31
redis::client: Add :reconnect_delay option

eavgerinos:
`:reconnect_delay` option can be used alongside `:reconnect_attempts`
option to provide a try->fail->backoff->retry mechanism for connecting
to Redis.

By using the `:reconnect_delay` option we can avoid having busy-loops
that try to reconnect and also add a time limit to reconnect retries.


eavgerinos:
I took it a step further and added [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) and a `:reconnect_delay_max` to avoid an unbounded sleep time.

kaspergrubbe:
Could we add some of the documentation from @wpp that was added in #710 ?

Thank you for merging this.

byroot:
Sure. Feel free to PR it. Otherwise I'll add that to the changelog etc when I cut the next release.

eavgerinos:
@byroot Thanks for merging :)

@kaspergrubbe By documentation we mean rdoc or an entry in README?

kaspergrubbe:
@byroot Do you know when you are going to make a release that contains this code?

byroot:
I didn't planned any. But looking at master a 4.03 might be in order. I'll look at it tomorrow morning.

byroot:
Here you go: https://rubygems.org/gems/redis/versions/4.0.3

kaspergrubbe:
We need to add the paragraph about reconnections from https://github.com/redis/redis-rb/pull/710 to the `README.md`, it is currently missing.

byroot:
Mind PRing it?

kaspergrubbe:
@byroot Not at all, here it is: https://github.com/redis/redis-rb/pull/790

PR #780, opened 2018-09-07
Increase timeout seconds for the CI

supercaracal:
https://travis-ci.org/redis/redis-rb/jobs/423903215

I apologize for several timeout errors. It seems Redis 3.0 Cluster often needs long timeout seconds.


byroot:
Yeah, I rather have  the CI be slightly slower than randomly failing.

Thanks for the PR.

Issue #782, opened 2018-09-15
Passing an ActiveSupport::Duration into brpop or blpop blocks indefinitely


GuyPaddock:
If you make the mistake of using Rails convenience methods on `Numeric` for `seconds`, `minutes`, etc, when passing a timeout into `brpop` or `blpop`, you will end up with no timeout at all. This is because `_bpop` does a switch on the exact class of the last argument.

Compare this:
```Ruby
@redis.blpop("my_key", 2)
```
To this:
```Ruby
@redis.blpop("my_key", 2.seconds)
```
The former case works because `2` ends up passing the type check, but `2.seconds` does not. Here's the offending code:
```Ruby
def _bpop(cmd, args)
    options = {}

    case args.last
    when Hash
      options = args.pop
    when Integer # <-- This line ignores durations
      # Issue deprecation notice in obnoxious mode...
      options[:timeout] = args.pop
    end
# ...
end
```

I will note that if the timeout is passed as a named argument, a duration is accepted properly. For example:
```Ruby
@redis.blpop("my_key", timeout: 2.seconds)
```

PR #783, opened 2018-09-15
Handle integer like objects as timeout of blpop and brpop

byroot:
Fix: https://github.com/redis/redis-rb/issues/782

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -1162,9 +1162,9 @@ class Redis
     case args.last
     when Hash
       options = args.pop
-    when Integer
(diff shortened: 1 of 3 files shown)


byroot:
Thanks for the review but the CI issues are legit, I need to find time to finalize this.

supercaracal:
How is this?

```ruby
if args.last.is_a?(Hash)
  options = args.pop
elsif args.last.respond_to?(:to_int)
  # Issue deprecation notice in obnoxious mode...
  options[:timeout] = args.pop.to_int
end
```

Issue #752, opened 2018-05-18
Can I help?


mperham:
Looks like this repo is struggling with support.  I would be happy to provide some maintenance / support for issues/PRs.  Interested?

djanowski:
I understand @badboy added more collaborators recently. What's the status of that, Jan-Erik?

badboy:
I haven't done so! Will do that on the weekend.

badboy:
I just added a bunch of Shopify folks as admins to this repo (through the redis-rb team):

@fw42 
@sirupsen
@rafaelfranca
@dylanahsmith
@csfrancis
@byroot

I will also add them as collaborators on hiredis-rb and to Rubygems.

Please coordinate ongoing work, don't remove folks and keep hacking!

fw42:
Danke Jan-Erik!

byroot:
Thanks @badboy, we'll do our best.

However if @mperham propose to step in, I think he would be very helpful person for this repo (if not more than us shopifolks).

tubbo:
I can also help, currently maintaining the redis-store gems and I've seen more than one place where some collaboration between us could really clean up some ugly code. will continue to monitor issues and contribute where I can.

whatl3y:
In addition to better support, I would love the latest updates be published to rubygems. I struggled for a while on trying to get the SSLContext verify_mode == `VERIFY_NONE` since I was connecting via the redis server's IP address, but realized after hours of trying, asking a stack overflow question, and opening a forum discussion in AWS with little help that the commit I needed wasn't up yet: https://github.com/redis/redis-rb/commit/25f1f705e0afb9942b2303799f4da156173e05ab

I am working around this by updating by Gemfile to be the following, but for obvious reasons would like to avoid this if possible in the future:

`gem "redis", :git => 'git://github.com/redis/redis-rb'`

byroot:
> I would love the latest updates be published to rubygems.

I'd love to as well, unfortunately I still don't have publish access on rubygems. I was traveling the last few week, but I'll try to get them.

@badboy if you ear us I'd love to push a new version out

Also cc @fw42

badboy:
And it's done (added the list of emails I got a while back from @fw42). Thanks for maintaining this and sorry for the delay.

PR #788, opened 2018-10-17
Fix raising command error for first command in pipeline

dylanahsmith:
## Problem

When the first command in a redis pipeline has invalid arguments, then the Redis::CommandError object is passed to the transform block which can cause the exception to be transformed and ignored.  For example,

```ruby
set_result, * = redis.pipelined do
  redis.set('mykey', 0, nx: true, px: 900.0) # px value is invalid if a float is given
end
puts set_result.inspect # => false
```

This problem was mentioned before for zrange in https://github.com/redis/redis-rb/pull/733 but the fix was specific to the FloatifyPairs transform block so didn't also fix the problem for the `set` command that uses `BoolifySet` for the transform block when using the `nx: true` option.

## Solution

Remove the special case in Redis::Client#call_pipelined to `read` the first command reply which was there only to set `@reconnect = false`, since it is fine to call that in the loop after each call to `read`.

I've removed the workaround that https://github.com/redis/redis-rb/pull/733 added for this problem which isn't needed now that the exception gets raised before the transform block gets called.  That way we can also rely on the test in that PR to make sure this works for any transform block.

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -2936,12 +2936,8 @@ private
 
   FloatifyPairs =
     lambda { |result|
-      if result.respond_to?(:each_slice)
-        result.each_slice(2).map do |member, score|
-          [member, Floatify.call(score)]
-        end
-      else
-        result
+      result.each_slice(2).map do |member, score|
+        [member, Floatify.call(score)]
       end
     }
 
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -191,13 +191,10 @@ class Redis
         exception = nil
 
         process(commands) do
-          result[0] = read
-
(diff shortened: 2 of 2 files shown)


dylanahsmith:
I think the intention was that `@reconnect = false` gets set after the first `read` succeeds so that it doesn't reconnect and try to read replies to commands set on the previous connection.  With the way it was written, it looks like missing the `exception = reply if exception.nil? && reply.is_a?(CommandError)` line after the first `read` was accidental.

> interested to see the maintainer response.

Looks like Shopify took over maintaining the gem (see #752).  Of course, I would appreciate another set of :eyes: on this first.

PR #794, opened 2018-11-05
Decrease timeout sec for the CI

supercaracal:
https://github.com/redis/redis-rb/pull/780#issuecomment-435791157

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_script:
 env:
   global:
     - VERBOSE=true
-    - TIMEOUT=30
+    - TIMEOUT=9
   matrix:
     - DRIVER=ruby      REDIS_BRANCH=3.0
     - DRIVER=ruby      REDIS_BRANCH=3.2
--- a/test/support/cluster/orchestrator.rb
+++ b/test/support/cluster/orchestrator.rb
@@ -7,7 +7,7 @@ class ClusterOrchestrator
 
   def initialize(node_addrs)
     raise 'Redis Cluster requires at least 3 master nodes.' if node_addrs.size < 3
-    timeout_sec = Float(ENV['TIMEOUT'] || 30.0)
+    timeout_sec = Float(ENV['TIMEOUT'] || 9.0)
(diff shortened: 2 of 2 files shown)


byroot:
I'm not sure it's just a problem of max timeout.

CI went from ~1h to ~4h when we merged cluster support, and then ~8h when timeout was bumped to 30s.

The cluster test suite is not small, but it's very surprising it's that slow, I wonder where the time is spent. Maybe we're frequently hitting that timeout?

supercaracal:
- [ ] investigate

supercaracal:
in my local

```
$ time make test

real    2m30.473s
user    1m39.712s
sys     0m2.332s
```

```
$ time TIMEOUT=9 make test

real    7m52.540s
user    1m41.680s
sys     0m4.764s
```

I did `strace`. I saw many `select(2)` calls with blocking.

https://github.com/redis/redis-rb/blob/master/test/lint/blocking_commands.rb#L149-L171

```ruby
mock(:delay => 1 + OPTIONS[:timeout] * 2) do |r|
# test code
end
```

```ruby
def mock(options = {}, &blk)
  commands = build_mock_commands(options)
  redis_mock(commands, &blk)
end
```

```ruby
def build_mock_commands(options = {})
  {
    blpop: lambda do |*args|
      sleep options[:delay] if options.key?(:delay)
      to_protocol([args.first, args.last])
    end,
    brpop: lambda do |*args|
      sleep options[:delay] if options.key?(:delay)
      to_protocol([args.first, args.last])
    end,
    brpoplpush: lambda do |*args|
      sleep options[:delay] if options.key?(:delay)
      to_protocol(args.last)
    end,
    bzpopmax: lambda do |*args|
      sleep options[:delay] if options.key?(:delay)
      to_protocol([args.first, args.last])
    end,
    bzpopmin: lambda do |*args|
      sleep options[:delay] if options.key?(:delay)
      to_protocol([args.first, args.last])
    end
  }
end
```

so many sleep

byroot:
Yeah, no reason for these tests to be as slow. If we're explicitly testing a timeout scenario, the timeout should be as low as possible (e.g 10ms).

PR #789, opened 2018-10-27
Fix resharding test helper method for cluster mode and enhance migrate command method

supercaracal:
I apologize for the [CI failure](https://travis-ci.org/redis/redis-rb/jobs/443189621). It's a resharding related error. It  seems to occur with low frequency even if the timeout is 30 seconds, so I added retry code.

```
/home/travis/build/redis/redis-rb/test/helper.rb:34:in `run'
     87:   def test_redirection_when_slot_is_resharding
     88:     100.times { |i| redis.set("{key}#{i}", i) }
     89: 
  => 90:     redis_cluster_resharding(12539, src: '127.0.0.1:7002', dest: '127.0.0.1:7000') do
     91:       100.times { |i| assert_equal i.to_s, redis.get("{key}#{i}") }
     92:     end
     93:   end
/home/travis/build/redis/redis-rb/test/cluster_client_slots_test.rb:90:in `test_redirection_when_slot_is_resharding'
/home/travis/build/redis/redis-rb/test/helper.rb:319:in `redis_cluster_resharding'
/home/travis/build/redis/redis-rb/test/support/cluster/orchestrator.rb:48:in `start_resharding'
/home/travis/build/redis/redis-rb/test/support/cluster/orchestrator.rb:48:in `loop'
/home/travis/build/redis/redis-rb/test/support/cluster/orchestrator.rb:51:in `block in start_resharding'
/home/travis/build/redis/redis-rb/test/support/cluster/orchestrator.rb:51:in `each'
/home/travis/build/redis/redis-rb/test/support/cluster/orchestrator.rb:51:in `block (2 levels) in start_resharding'
/home/travis/build/redis/redis-rb/lib/redis.rb:515:in `migrate'
/home/travis/build/redis/redis-rb/lib/redis.rb:50:in `synchronize'
/home/travis/.rvm/rubies/ruby-2.5.0/lib/ruby/2.5.0/monitor.rb:226:in `mon_synchronize'
/home/travis/build/redis/redis-rb/lib/redis.rb:50:in `block in synchronize'
/home/travis/build/redis/redis-rb/lib/redis.rb:516:in `block in migrate'
/home/travis/build/redis/redis-rb/lib/redis/client.rb:124:in `call'
Error: test_redirection_when_slot_is_resharding(TestClusterClientSlots): Redis::CommandError: IOERR error or timeout reading to target instance
```

[MIGRATE command reference](https://redis.io/commands/migrate)

[...]

Issue #774, opened 2018-08-13
Flaky test: SentinelTest#test_sentinel_retries


byroot:
This test is plaguing CI runs: 

```ruby
===============================================================================
/home/travis/build/redis/redis-rb/test/helper.rb:60:in `run'
     236: 
     237:     connections.clear
     238: 
  => 239:     ex = assert_raise(Redis::CannotConnectError) do
     240:       RedisMock.start(master) do |master_port|
     241:         RedisMock.start(handler.call(:s1, master_port)) do |s1_port|
     242:           RedisMock.start(handler.call(:s2, master_port)) do |s2_port|
/home/travis/build/redis/redis-rb/test/sentinel_test.rb:239:in `test_sentinel_retries'
<Redis::CannotConnectError> expected but was
<Errno::ECONNRESET(<Connection reset by peer - connect(2) for 127.0.0.1:37455>)>
diff:
? Redis:: Cannot     ConnectE   r           r               or                  
? Errno  E ONNRESET(<       ion  eset by pee  - connect(2) f   127.0.0.1:37455>)
Failure: test_sentinel_retries(SentinelTest)
===============================================================================
```

PR #795, opened 2018-11-05
Fix slow test cases which depend on timeout seconds

supercaracal:
#780
#794

### DIFF
--- a/test/lint/blocking_commands.rb
+++ b/test/lint/blocking_commands.rb
@@ -1,5 +1,7 @@
 module Lint
   module BlockingCommands
+    LOW_TIMEOUT = 0.01
+
     def setup
       super
 
@@ -25,7 +27,7 @@ module Lint
 
     def mock(options = {}, &blk)
       commands = build_mock_commands(options)
-      redis_mock(commands, &blk)
+      redis_mock(commands, { timeout: LOW_TIMEOUT }, &blk)
     end
 
     def build_mock_commands(options = {})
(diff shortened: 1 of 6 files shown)


byroot:
Master: `Finished in 1117.441769271 seconds.`
This branch: `Finished in 149.787086536 seconds.`

Nice 7-8x speedup here. 👏

test/lint/blocking_commands.rb
```diff
@@ -1,5 +1,8 @@
 module Lint
   module BlockingCommands
+    TIMEOUT_SEC = 0.01
```
byroot: I'm not a huge fan of `TIMEOUT_SEC`, as it's not very meaningful. What about `LOW_TIMEOUT` ?

test/lint/blocking_commands.rb
```diff
@@ -1,5 +1,8 @@
...
+    DELAY_SEC = 0.05
```
byroot: Similar issue with the naming here. I'd almost suggest not to constantize this and to use `LOW_TIMEOUT * 5` as an expression.

supercaracal:
Hmm... Several errors occurred on CI but they didn't reproduce on my local environment. I'll try investigate them.

* https://travis-ci.org/redis/redis-rb/jobs/451166986
  * related Redis Sentinel
* https://travis-ci.org/redis/redis-rb/jobs/451167015
  * related #789

supercaracal:
@byroot I have fixed them and CI has been success.

byroot:
Many thanks for handling this!

supercaracal:
https://github.com/redis/redis-rb/issues/774
fixed by 51ee07b

PR #797, opened 2018-11-13
Change REDIS_URL behavior to avoid overwrite by options

### LINKED ISSUE #679: Default URL unexpectedly merges with empty options
Configuring a Redis client, a user is expected to either use several hash options (`:host`, `:port`, etc...) *or* a `:url`. I've seen confusion in environments with multiple Redis clients around the code's behavior here:

Expected behavior:
If REDIS_URL is defined and I create a new client with `Redis.new( host: 'foo', password: 'bar' )`, I would expect the other settings to have normal defaults (i.e. port 6379, db 0, etc)

Actual behavior:
If REDIS_URL is defined, the resulting Redis client merges the REDIS_URL settings with the options passed to Redis.new.

To demonstrate, in an environment with REDIS_URL = redis://x:mypass@dev.server:86379

```
Redis.new.client.port == 86379  # expected

redis = Redis.new( host: "dev2", password: "mypass2" )
redis.client.port == 86379  # I would expect this to be 6379
```

I can submit a patch to fix this minor issue if it would be accepted – wanted to check how that would fit in with some of the other in-flight changes I see in this repo.

phaldan:
I have increased the priority of the connection settings from `REDIS_URL` environment variables to avoid overrides by hash options from `Redis.new`. This should resolve #679.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -6,7 +6,7 @@ class Redis
   class Client
 
     DEFAULTS = {
-      :url => lambda { ENV["REDIS_URL"] },
+      :url => nil,
       :scheme => "redis",
       :host => "127.0.0.1",
       :port => 6379,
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Thank you for the pull request.

I disagree that the environment variable should have precedence over explicitly passed options. It should works in the same way that passing an `:url` option to `Redis.new`. If I get it write this is how it behaves right now.

If anything I'd remove the automatic handling of `REDIS_URL` like commented here https://github.com/redis/redis-rb/issues/679#issuecomment-278497128.

rafaelfranca:
I think this should be up to the object instantiation that you want to do. You as a developer should chose if you want the `REDIS_URL` to apply to your `Redis.new` call or not. And if we give always precedence to `REDIS_URL` it means that all `Redis.new` call in the application will use the same Redis instance and this is not what may want.

I don't think there is a rush to drop support to REDIS_URL. This behavior exists probably since the first version and this was many years ago. If this behavior worked fine for this many years there is no reason to rush removal of it.

PR #800, opened 2018-11-24
Add a Rake task for test

supercaracal:
I'd say that the following test command is easier to type:

```
$ bin/rake test test/something_test.rb
```

than:

```
$ bundle exec ruby -w -Itest test/something_test.rb -v
```


byroot:
Interesting. Overall I think the test suite need some cleanup. Ideally I'd get rid of make & also upgrade to minitest.

I'll try to get on that when I have more time to spent on the gem, hopefully early January.

PR #805, opened 2018-11-28
Fix several issues for CI

supercaracal:
I fixed the following issues in CI.

* `TIMEOUT` error was occasionally occured in JRuby.
  * => I fixed as to increase `LOW_TIMEOUT` seconds from `0.01` to `0.03` seconds.
* `CLUSTERDOWN` error was occasionally occured after failover test cases.
  * => I fixed as to use `WAIT` command instead of uncertainly `sleep`.
    * https://redis.io/commands/wait
    * > However in the context of Sentinel or Redis Cluster failover, WAIT improves the real world data safety.
  * ref: https://github.com/redis/redis-rb/pull/807

### DIFF
--- a/test/support/cluster/orchestrator.rb
+++ b/test/support/cluster/orchestrator.rb
@@ -5,10 +5,10 @@ require_relative '../../../lib/redis'
 class ClusterOrchestrator
   SLOT_SIZE = 16384
 
-  def initialize(node_addrs)
+  def initialize(node_addrs, timeout: 30.0)
     raise 'Redis Cluster requires at least 3 master nodes.' if node_addrs.size < 3
-    timeout_sec = Float(ENV['TIMEOUT'] || 30.0)
-    @clients = node_addrs.map { |addr| Redis.new(url: addr, timeout: timeout_sec) }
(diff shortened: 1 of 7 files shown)


test/support/cluster/orchestrator.rb
```diff
@@ -157,27 +156,35 @@ def save_config(clients)
...
+    end
+  end
 
+  def wait_for(state, clients, max_attempts)
+    attempt_count = 1
     clients.each do |client|
-      loop do
-        flags = hashify_cluster_node_flags(client)
+      attempt_count.step(max_attempts) do |i|
+        p "Reached max attempts at waiting for #{state}" if i >= max_attempts
```
byroot: Is this print a forgotten debug statement?

supercaracal: No, it isn't. It is on purpose. There's a possibility that test failure is occured when reaching max attempts. I think that the print is better than nothing.

Is this print redundant or noisy?

supercaracal: I'm sorry, I misunderstood. I removed the statement and added `break` step.

PR #807, opened 2018-12-01
Exclude test files from bundled gem

greysteil:
Rubygems doesn't use `test_files` anymore ([issue](https://github.com/rubygems/guides/issues/90)), and by removing the tests from the bundled gem size drops by 2/3rds. (It's not a big gem anyway, but since each version is installed millions of times hopefully that's a saving worth having.)


byroot:
Thanks.

PR #799, opened 2018-11-22
Add Streams support

supercaracal:
I'd like to use Redis Streams.

* [Introduction to Redis Streams](https://redis.io/topics/streams-intro)
* [Redis Streams Commands](https://redis.io/commands#stream)

PR #811, opened 2018-12-10
Stream feature polish

byroot:
While playing with https://github.com/redis/redis-rb/pull/799 locally, I realized it uses some modern Ruby feature like `Array#concat` varargs signature.

This PR updates the builds matrix so that it spots more of this kind of issues.

I'll wait for the results and fix what comes up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,9 @@ before_install:
 script: make
 
 rvm:
-  - 2.2.2
-  - 2.3.3
-  - 2.4.1
-  - 2.5.0
+  - 2.3.8
+  - 2.4.5
+  - 2.5.3
   - jruby-9.1.17.0
 
 gemfile: ".travis/Gemfile"
@@ -30,11 +29,10 @@ env:
   matrix:
     - DRIVER=ruby      REDIS_BRANCH=3.0
     - DRIVER=ruby      REDIS_BRANCH=3.2
-    - DRIVER=hiredis   REDIS_BRANCH=3.0
-    - DRIVER=hiredis   REDIS_BRANCH=3.2
-    - DRIVER=synchrony REDIS_BRANCH=3.0
-    - DRIVER=synchrony REDIS_BRANCH=3.2
     - DRIVER=ruby      REDIS_BRANCH=4.0
+    - DRIVER=ruby      REDIS_BRANCH=5.0
+    - DRIVER=hiredis   REDIS_BRANCH=5.0
+    - DRIVER=synchrony REDIS_BRANCH=5.0
 
 branches:
   only:
@@ -52,8 +50,6 @@ matrix:
       env: DRIVER=ruby REDIS_BRANCH=3.2 LOW_TIMEOUT=0.3
     - rvm: jruby-9.1.17.0
       env: DRIVER=ruby REDIS_BRANCH=4.0 LOW_TIMEOUT=0.3
-    - rvm: 2.5.3
-      env: DRIVER=ruby REDIS_BRANCH=5.0
 
 notifications:
   irc:
--- a/bin/console
+++ b/bin/console
@@ -1,0 +1,7 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH.unshift(File.expand_path('../../lib', __FILE__))
+require 'redis'
+
+require 'irb'
+IRB.start
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -3055,8 +3055,8 @@ class Redis
   # @return [Hash{String => Hash{String => Hash}}] the entries
   def xread(keys, ids, count: nil, block: nil)
     args = [:xread]
-    args.concat(['COUNT', count])      if count
-    args.concat(['BLOCK', block.to_i]) if block
+    args << 'COUNT' << count if count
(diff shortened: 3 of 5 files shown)


byroot:
cc @supercaracal just so you know.

.travis.yml
```diff
@@ -9,10 +9,9 @@ before_install:
 script: make
 
 rvm:
-  - 2.2.2
-  - 2.3.3
-  - 2.4.1
-  - 2.5.0
+  - 2.3.8
```
byroot: I updated the rubies to recent point releases, and dropped 2.2.x since it's now EOL. The matrix is already huge so we can't realistically test EOL versions.

.travis.yml
```diff
@@ -30,11 +29,10 @@ env:
...
     - DRIVER=ruby      REDIS_BRANCH=3.0
     - DRIVER=ruby      REDIS_BRANCH=3.2
-    - DRIVER=hiredis   REDIS_BRANCH=3.0
-    - DRIVER=hiredis   REDIS_BRANCH=3.2
-    - DRIVER=synchrony REDIS_BRANCH=3.0
-    - DRIVER=synchrony REDIS_BRANCH=3.2
     - DRIVER=ruby      REDIS_BRANCH=4.0
+    - DRIVER=ruby      REDIS_BRANCH=5.0
+    - DRIVER=hiredis   REDIS_BRANCH=5.0
+    - DRIVER=synchrony REDIS_BRANCH=5.0
```
byroot: The idea here is to test all the versions of the server with the ruby driver, but only the most recent one with the alternative drivers. I believe it's a good compromise between coverage and matrix size.

byroot:
@supercaracal another thing I spotted and I'd like your opinion on. `XREAD` returns a hash of a hash:

```ruby
>> pp r.xread('mystream', '0-0', block: 50_000)
{"mystream"=>
  {"1544447512733-0"=>{"f1"=>"v1", "f2"=>"v2"},
   "1544447564923-0"=>{"f1"=>"v2", "f2"=>"v5"}}}
```

I wonder if it actually make sense, and if arrays wouldn't be more appropriate:

```
{"mystream"=> [
  ["1544447512733-0", {"f1"=>"v1", "f2"=>"v2"}],
  ["1544447564923-0",{"f1"=>"v2", "f2"=>"v5"}]
]}
```

Granted, hashes are ordered in Ruby, but I'm not sure addressing the result by `id` actually make sense. What do you think?

supercaracal:
@byroot I think so too. I'm sorry. Hash is excessive abstraction. Array is more intuitive than Hash.

byroot:
No need to be sorry !

byroot:
@supercaracal I pushed a change so that we use `start` and `end` as params to be consistent with the server docs. Mind taking a look when you have time?

byroot:
Actually I'm now wondering if we should use named arguments at all for `start` and `end`.

```ruby
def xrange(key, start='-', _end='+', count: nil)
def xrevrange(key, _end='+', start='-', count: nil)
```

That's probably as easy to use if not more ^

supercaracal:
:+1: It is consistent. e.g. [#bitpos](https://github.com/redis/redis-rb/blob/v4.0.3/lib/redis.rb#L1033).

supercaracal:
Perhaps, we should use named arguments for the `count` option too. e.g. [#spop](https://github.com/redis/redis-rb/blob/v4.0.3/lib/redis.rb#L1382) [#srandmember](https://github.com/redis/redis-rb/blob/v4.0.3/lib/redis.rb#L1397)

byroot:
I think named arg for count make sense.

I feel like a common use for xrange is "give me the last X events", so `xrange(key, count: X)`

Another likely common use case is "give me events since X, but no more than 50", so `xrange(key, last_id, count: 50)`.

supercaracal:
Certainly. I understand.

supercaracal:
I'd say that the use cases of specifying XRANGE's `end` or XREVRANGE's `start` are rare.

I think that the following argument pattern is easy-to-use and consistent.
* `start` or `end` options are named arguments
* `count` option is keyword argument

byroot:
I'm not sure I understand the distinction you make between keyword and named arguments. Care to explain ?

supercaracal:
I'm sorry. I made a mistake. I chose wrong English word.

x: named argument
o: optional argument

supercaracal:
I apologize for the above my confused English.

If we should adjust to Redis command interface:

```ruby
# xread(key, start, _end, count = nil)
r.xread('s1', '-', '+')
r.xread('s1', '-', '+', 10)
r.xread('s1', '0-9', '+', 10)
```

* Pros
  * simple and intuitive
  * consistent for other methods
* Cons
  * we have to specify `start` and `end` arguments every time

If we have priority to easy-to-use interface in Ruby:

```ruby
# xread(key, start = '-', _end = '+', count: nil)
r.xread('s1')
r.xread('s1', count: 3)
r.xread('s1', '0-1')
r.xread('s1', '0-1', '0-9')
r.xread('s1', '0-1', count: 3)
```

* Pros
  * easy-to-use
* Cons
  * need a few learning cost for the abstraction

I'd say that the following interface looks good.

```ruby
# xread(key, start = '-', _end = '+', count: nil)
```

Issue #813, opened 2018-12-17
password to authenticate to Redis sentinel


mmb:
Is there a way to provide a password for the Redis sentinel (configured in the sentinel with requirepass)?

It looks like the password for the sentinel connection always gets set to nil here:

https://github.com/redis/redis-rb/blob/master/lib/redis/client.rb#L528

Issue #810, opened 2018-12-07
Redis 5.0 support


anero:
Just to double check: latest gem version doesn't support redis 5.0, does it? If not, any idea of timelines for that to happen?

Thanks!

PR #814, opened 2018-12-20
Fix sentinel authentication for Redis5

supercaracal:
Hi,

This is related https://github.com/redis/redis-rb/issues/813.

---
Postscript:
It seems that the feature is a miner change of Redis5.
[Configuring Sentinel instances with authentication](https://redis.io/topics/sentinel#configuring-sentinel-instances-with-authentication)
> this feature is only available starting with Redis 5.0.1
---

The findings:
I think there is no way to be able to authenticate by current implementation. The error related the issue was reproduced on my local real Sentinels like the following.

I ran three Sentinels with a master Redis and a replica Redis.
```
+------------------------+
| M1 <--- slaveof --- R1 |
| ^                    ^ |
| |                    | |
| +----------+---------+ |
|            |           |
|       +----+----+      |
|       |    |    |      |
|      S1   S2   S3      |
+------------------------+
```

I configured `requirepass` and `masterauth` options each instance. `M1` `R1` `S1` `S2` `S3`

```
requirepass hogehoge
masterauth hogehoge
```

I configured `auth-pass` option each Sentinel. `S1` `S2` `S3`

```
sentinel auth-pass mymaster hogehoge
```

I tried to use the client in `bin/console`, and then, the following error was occurred.

```
irb(main):008:0> r = Redis.new(url: 'redis://mymaster', sentinels: [{host:'127.0.0.1',port:6400}], password: 'hogehoge')
=> #<Redis client v4.1.0 for redis://mymaster:6379/0>       
irb(main):009:0> r.get :hoge         
       16: from /path/to/redis-rb/lib/redis/client.rb:123:in `call'
       15: from /path/to/redis-rb/lib/redis/client.rb:223:in `process'
       14: from /path/to/redis-rb/lib/redis/client.rb:312:in `logging'
       13: from /path/to/redis-rb/lib/redis/client.rb:224:in `block in process'
       12: from /path/to/redis-rb/lib/redis/client.rb:372:in `ensure_connected'
       11: from /path/to/redis-rb/lib/redis/client.rb:103:in `connect'
       10: from /path/to/redis-rb/lib/redis/client.rb:299:in `with_reconnect'
        9: from /path/to/redis-rb/lib/redis/client.rb:104:in `block in connect'               
        8: from /path/to/redis-rb/lib/redis/client.rb:337:in `establish_connection'
        7: from /path/to/redis-rb/lib/redis/client.rb:556:in `resolve'
        6: from /path/to/redis-rb/lib/redis/client.rb:592:in `resolve_master'
        5: from /path/to/redis-rb/lib/redis/client.rb:567:in `sentinel_detect'
        4: from /path/to/redis-rb/lib/redis/client.rb:567:in `each'                              
        3: from /path/to/redis-rb/lib/redis/client.rb:575:in `block in sentinel_detect'
        2: from /path/to/redis-rb/lib/redis/client.rb:593:in `block in resolve_master'
        1: from /path/to/redis-rb/lib/redis/client.rb:124:in `call'
Redis::CommandError (NOAUTH Authentication required.)
```

ref: [Sentinel and Redis authentication](https://redis.io/topics/sentinel#sentinel-and-redis-authentication)
ref: https://github.com/redis/redis-rb/issues/810

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require "test/unit"
 require "mocha/test_unit"
 require "logger"
@@ -180,6 +182,40 @@ module Helper
 
     def _new_client(options = {})
       Redis.new(_format_options(options).merge(:driver => ENV["DRIVER"]))
+    end
+  end
+
+  module Sentinel
+    include Generic
+
+    MASTER_PORT = PORT.to_s
+    SLAVE_PORT = '6382'
+    SENTINEL_PORT = '6400'
+    SENTINEL_PORTS = %w[6400 6401 6402].freeze
+    MASTER_NAME = 'master1'
+    LOCALHOST = '127.0.0.1'
+
+    def build_sentinel_client(options = {})
+      opts = { host: LOCALHOST, port: SENTINEL_PORT, timeout: TIMEOUT, logger: ::Logger.new(@log) }
+      Redis.new(opts.merge(options))
+    end
+
+    def build_slave_role_client(options = {})
+      _new_client(options.merge(role: :slave))
+    end
+
+    private
+
+    def _format_options(options = {})
+      {
+        url: "redis://#{MASTER_NAME}",
+        sentinels: [{ host: LOCALHOST, port: SENTINEL_PORT }],
+        role: :master, timeout: TIMEOUT, logger: ::Logger.new(@log)
+      }.merge(options)
+    end
+
+    def _new_client(options = {})
+      Redis.new(_format_options(options).merge(driver: ENV['DRIVER']))
     end
   end
 
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -525,7 +525,6 @@ class Redis
         def initialize(options)
           super(options)
 
-          @options[:password] = DEFAULTS.fetch(:password)
           @options[:db] = DEFAULTS.fetch(:db)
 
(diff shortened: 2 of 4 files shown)


supercaracal:
- [ ] Use real sentinels as possible for test

test/helper.rb
```diff
@@ -183,6 +185,40 @@ def _new_client(options = {})
...
+      Redis.new(opts.merge(options))
+    end
+
+    def build_slave_role_client(options = {})
+      _new_client(options.merge(role: :slave))
+    end
+
+    private
+
+    def _format_options(options = {})
```
byroot: Why are these methods prefixed with an underscore? They are private, do they conflict with other methods in Generic?

supercaracal: No, there are no conflicts. I think the underscore is meaningless too.

The `Generic` call the `#_new_client` method at [#setup](https://github.com/redis/redis-rb/blob/v4.1.0/test/helper.rb#L102). I think the private methods are overall need refactoring but this Pull Request is focused on the Sentinel tests.

* [Client](https://github.com/redis/redis-rb/blob/v4.1.0/test/helper.rb#L177-L183)
* [Distributed](https://github.com/redis/redis-rb/blob/v4.1.0/test/helper.rb#L197-L206)
* [Cluster](https://github.com/redis/redis-rb/blob/v4.1.0/test/helper.rb#L340-L350)

PR #819, opened 2019-01-10
Always return Fixnum (or Float when :incr is specified) for SADD / ZADD

jnoconor:
For someone who is familiar with Redis, returning booleans from SADD and ZADD (and SREM / ZREM) is confusing at first, and I don't see the benefit, so I've removed the boolean casting from these methods.

Further explanation here: #820

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -1338,21 +1338,10 @@ class Redis
   #
   # @param [String] key
   # @param [String, Array<String>] member one member, or array of members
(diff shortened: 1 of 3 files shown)


byroot:
Thank you for your contribution but this would be a breaking change so we simply cannot merge it.

Also the reason why they return boolean is so that you can use it like this:

```ruby
if redis.sadd(key, value)
  # value was added to the set
else
  # value was already present.
end
```

PR #822, opened 2019-01-24
Remove old (and inaccurate) documetation

jrmhaig:
This section of comments appears to be left in accidentally when the argument for the `xrange` and `xrevrange` methods were changed [here.](https://github.com/jrmhaig/redis-rb/commit/60079e44ae4d71a378ea248055fd37ca29118e59)


byroot:
Thanks!

PR #829, opened 2019-03-01
Fix to consider flags of slaves on connection via sentinel

### LINKED ISSUE #827: Redis client tries to connect to slaves marked as down by Redis Sentinel
While replacing our Redis Sentinel setup we found an issue with the redis Ruby gem trying (and failing) to connect to a slave marked as down by Redis Sentinel (all of them had the `+sdown slave` message in their logs). I can reproduce the issue in a VM, with two slaves and a master (using Redis 4.10, 4.13 and 5.03 and the Ruby gem 3.2.1 and 4.1.0). So far I have found that raising the `reconnect_attempts` option of the Redis client results in successful connections to the slave that is up (we previously set it to 3, and 10 appears to be the sweet spot from what I have observed).

Is that a bug in the client library, or am I doing something wrong? If that's working as intended (Sentinels returning a slave that is known as down) I will open an issue in the Redis server repository

supercaracal:
Hi,

I've fixed  #827.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -604,7 +604,10 @@ class Redis
         def resolve_slave
           sentinel_detect do |client|
             if reply = client.call(["sentinel", "slaves", @master])
-              slave = Hash[*reply.sample]
+              slave = reply.map { |r| Hash[*r] }
+                           .reject { |s| s['flags'].split(',').include?('s_down') }
+                           .sample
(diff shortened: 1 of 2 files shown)


byroot:
Thanks, merged as: https://github.com/redis/redis-rb/commit/c5c91fc2fabcb78cb4073a0bfc7371925cbdf0df

Issue #836, opened 2019-04-06
XREAD with BLOCK 0 should not time out


Kache:
**Observed**

```ruby
Redis.current.xread(:key, 0, block: 0)
# 10 secs later... => Redis::TimeoutError: Connection timed out
```

**Expected**

`BLOCK 0` should mean to not time out and wait forever.

**Notes**

The same [`xread` timeout (`block`)](https://github.com/redis/redis-rb/blob/v4.1.0/lib/redis.rb#L3076) is always ["mixed" with the TCP socket timeout](https://github.com/redis/redis-rb/blob/v4.1.0/lib/redis.rb#L3492).

**Workaround**

```ruby
XReadBlock = Struct.new(:msec) do
  def to_i
    msec.to_i
  end

  def to_f
    set_socket_timeout_to_zero = Redis.current._client.timeout * -1000
    to_i.zero? ? set_socket_timeout_to_zero : msec.to_f
  end
end

Redis.current.xread(:key, 0, block: XReadBlock.new(0)) # blocks forever, as expected
```

PR #837, opened 2019-04-11
Fix timeout error on xread with block zero option

supercaracal:
This PR is related to #836.

ref: #810

### DIFF
--- a/test/lint/streams.rb
+++ b/test/lint/streams.rb
@@ -346,6 +346,20 @@ module Lint
       assert_equal({}, actual)
     end
 
+    def test_xread_does_not_raise_timeout_error_when_the_block_option_is_zero_msec
+      prepared = false
+      actual = nil
+      wire = Wire.new do
+        prepared = true
+        actual = redis.xread('s1', 0, block: 0)
+      end
+      Wire.pass until prepared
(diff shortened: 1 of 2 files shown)


test/lint/streams.rb
```diff
@@ -346,6 +346,20 @@ def test_xread_with_block_option
...
 
+    def test_xread_does_not_raise_timeout_error_when_the_block_option_is_zero_msec
+      prepared = false
+      actual = nil
+      wire = Wire.new do
+        prepared = true
+        actual = redis.xread('s1', 0, block: 0)
+      end
+      Wire.pass until prepared
+      redis.dup.xadd('s1', { f: 'v1' }, id: '0-1')
```
byroot: Not sure if `.dup` is a good idea. It's only a shallow clone, so you might be sharing the mutex etc.

supercaracal: Hello,

I appreciate for your review. 

It seems that the [`.dup`](https://github.com/redis/redis-rb/blob/v4.1.0/lib/redis.rb#L3303-L3305) is overridden. It looks like that the [`.dup`](https://github.com/redis/redis-rb/blob/v4.1.0/lib/redis.rb#L3303-L3305) make another instance.

byroot: TIL.

LGTM then.

PR #838, opened 2019-04-12
Test against latest rubies, drop 2.3 testing

DmytroStepaniuk:


.travis.yml
```diff
@@ -9,9 +9,9 @@ before_install:
 script: make
 
 rvm:
-  - 2.3.8
-  - 2.4.5
-  - 2.5.3
+  - 2.4.6
+  - 2.5.5
+  - 2.6.2
```
springerigor: `2.6.3` is already there :)

mikong:
If Ruby 2.3 testing is dropped, does that mean the Redis gem is no longer officially supporting that version? If so, then we should also update the redis.gemspec file to specify that `s.required_ruby_version = '>= 2.4.0'`.

I can make a new PR if this one is stale, including updating the Ruby versions to the latest ones (2.4.9, 2.5.7, and 2.6.5) for Travis.

byroot:
> If Ruby 2.3 testing is dropped,

I don't see why it should be dropped though.

mikong:
@byroot, I wasn't particularly recommending to drop Ruby 2.3 testing. The PR started out recommending that and I was just building on it since it was approved somehow. I can create a new PR keeping Ruby 2.3 and updating the other Ruby versions to the latest ones.

I've seen other projects drop Ruby 2.3. I think the reasoning behind it is that Ruby 2.3 was marked EOL last March 2019.

byroot:
Superseeded by d2763f29a41265ee1678ddfd1b8458c1a8d18535

byroot:
> I've seen other projects drop Ruby 2.3. I think the reasoning behind it is that Ruby 2.3 was marked EOL last March 2019.

Indeed, I'm not particularly keen on maintaining EOL rubies, however 2.3 isn't that much of a burden at this point, so I'll leave it for now.

PR #840, opened 2019-04-17
Use nonblocking IO for SSL socket connect.

### LINKED ISSUE #835: Misconfiguration of SSL causes thread to hang
The redis library seems to hang when passing incorrect configuration. 
In particular, we came across a situation where we set `ssl:true` when the redis server was not configured for SSL. 

To reproduce:
Run plain old redis locally and run the following ruby file. 

```
require 'redis'
redis = Redis.new(
    url: 'redis://localhost:6379',
    ssl: true,
)
puts 'Requesting a value from redis'
redis.get('abc')
puts "done"
```

esiegel:
Attempting to use SSL when the server isn't configured this way will
lead to the connection hanging forever.  By using nonblocking IO we
can respect a timeout.

fixes #835

### DIFF
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -266,7 +266,22 @@ class Redis
 
           ssl_sock = new(tcp_sock, ctx)
           ssl_sock.hostname = host
-          ssl_sock.connect
+
+          begin
+            ssl_sock.connect_nonblock
+          rescue IO::WaitReadable, IO::WaitWritable
+            resp = IO.select([ssl_sock], [ssl_sock], nil, timeout)
+
+            if resp.nil?
+              raise TimeoutError
+            end
+
+            r, w = resp
+
+            if r.size > 0 || w.size > 0
+              retry
+            end
+          end
 
           unless ctx.verify_mode == OpenSSL::SSL::VERIFY_NONE
             ssl_sock.post_connection_check(host)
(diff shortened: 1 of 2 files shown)


esiegel:
The tests seem to pass now, but I do see warnings about being unable to run the mock server in some of the builds.  I do see this in other unrelated builds too, so maybe this isn't anything to worry about?

byroot:
I added a test for it, but it doesn't seem to work: https://github.com/redis/redis-rb/commit/6dac97215526339bfcba439f6a7af0cf6b17577c

esiegel:
Updated the diff to basically follow the code found in the kafka library.  Somewhat confused what my original code was doing wrong.  Maybe listening for reads and waits on IO, or maybe the size checks?

byroot:
> Somewhat confused what my original code was doing wrong. Maybe listening for reads and waits on IO, or maybe the size checks?

From what I saw in my breakpoint, the socket was ready for writes so it was looping forever.

esiegel:
@byroot do you notice any other issues, or do you think it is ready to be merged?

byroot:
I need to test it a bit myself, but right now I'm sick. I'll check this back in a couple days.

byroot:
Merged as https://github.com/redis/redis-rb/commit/ec3955209715d7e5a68ac84615ad2764de77a0fb

Thanks.

PR #841, opened 2019-04-30
Fix bug for equal params to georadius. Allow array param for geoadd

rahul342:
- Currently there's bug when searching with same params for Georadius
and georadiusbymember, due to `args.uniq` (not sure why uniq is being
called). Example calling for (lat = 15, lng = 15), it returns `ERR wrong number of arguments for 'georadius' command`
- Georadius and georadiusbymember accept array params while geoadd
doesnt. Making geoadd consistent and more user-friendly.

### DIFF
--- a/test/commands_on_geo_test.rb
+++ b/test/commands_on_geo_test.rb
@@ -12,6 +12,21 @@ class TestCommandsGeo < Test::Unit::TestCase
     end
   end
 
+  def test_geoadd_with_array_params
+    target_version "3.2.0" do
+      added_items_count = r.geoadd("SicilyArray", [13.361389, 38.115556, "Palermo", 15.087269, 37.502669, "Catania"])
+      assert_equal 2, added_items_count
+    end
+  end
+
+  def test_georadius_with_same_params
+    target_version "3.2.0" do
+      r.geoadd("Chad", 15, 15, "Kanem")
+      nearest_cities = r.georadius("Chad", 15, 15, 15, 'km', sort: 'asc')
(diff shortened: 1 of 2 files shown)


rahul342:
cc @supercaracal since you added Redis Cluster support (Thanks for dong that!)

rahul342:
@byroot Could you please help me review and fix this? Would love to not have to monkeypatch in our codebase for such an obvious bug.

lib/redis/version.rb
```diff
@@ -1,3 +1,3 @@
 class Redis
-  VERSION = '4.1.0'
+  VERSION = '4.1.1'
```
byroot: Please don't bump the version. 

Releases are done out of band of PRs.

test/commands_on_geo_test.rb
```diff
@@ -12,6 +12,21 @@ def setup
     end
   end
 
+  def test_geoadd_with_array_params
+    target_version "4.1.1" do
```
byroot: This is not what `target_version` is for.

It's there to specify the minimal redis server version, not the client one.

rahul342:
@byroot Thanks for checking. Comments addressed.

rahul342:
@byroot what's the release cadence like?

byroot:
There is no specific cadence, and there's plenty of PRs I need to review before I'd like to cut a new release. But I have very little time this days so might take a while.

Just use the gem using bundler git source:

```
gem 'redis', github: 'redis/redis-rb'
```

and you'll be fine.

Issue #487, opened 2014-12-15
remove reliance on Test::Unit


ktdreyer:
After the discussion in https://github.com/redis/redis-rb/pull/445 , what test library would you like to use for redis-rb? There was a mention of cutest. Would you accept a pull request for that?

ktdreyer:
Hi Damian, just wanted to check in again on this. Any thoughts regarding a future test library you'd accept for redis-rb? Would you accept a pull request for cutest?

ktdreyer:
Hi @djanowski , I just wanted to check in again regarding this Test::Unit dependency and how we could switch to something more modern.

Would you prefer cutest?

PR #844, opened 2019-05-06
Convert test suite to use Minitest

byroot:
Fix: https://github.com/redis/redis-rb/issues/487

Original authors had objections against minitest (https://github.com/redis/redis-rb/pull/445), but ultimately it's more or less today's standard. More idiomatic is better.


byroot:
@supercaracal since I see you 👍 the PR: The minitest migration seem to trigger a handful of inconsistent errors with various redis cluster test.

I highly suspect it's legit, and we're only seeing it now because Minitest now randomize test order, but haven't managed to repro locally yet. If you have a minute, maybe the errors will speak more to you than to me:

```

  1) Failure:
TestClusterClientReplicas#test_some_reference_commands_are_sent_to_slaves_if_needed [/home/travis/build/redis/redis-rb/test/cluster_client_replicas_test.rb:33]:
--- expected
+++ actual
@@ -1 +1 @@
-["key0", "key1", "key2", "key3", "key4"]
+[]
  2) Failure:
TestClusterClientReplicas#test_client_can_command_with_replica [/home/travis/build/redis/redis-rb/test/cluster_client_replicas_test.rb:17]:
--- expected
+++ actual
@@ -1,2 +1 @@
-# encoding: US-ASCII
-"0"
+nil
```

I'm thinking a previous test might have broken the replication, the problem is to figure out which...

supercaracal:
I will look into the above test cases ASAP.

byroot:
No rush, it's just in case it speaks to you. otherwise I'll keep digging myself.

supercaracal:
I'm sorry for missing replication delay. The [WAIT](https://redis.io/commands/wait) command might be needed after updatable commands as the followings.

https://github.com/redis/redis-rb/blob/v4.1.1/test/cluster_client_transactions_test.rb#L43
```
rc1.wait(1, TIMEOUT.to_i * 1000)
```

byroot:
❤️  Let's try that.

byroot:
Seems like it might not be enough:

```

  1) Failure:
TestClusterClientReplicas#test_some_reference_commands_are_sent_to_slaves_if_needed [/home/travis/build/redis/redis-rb/test/cluster_client_replicas_test.rb:37]:
--- expected
+++ actual
@@ -1 +1 @@
-["key0", "key1", "key2", "key3", "key4"]
+["key0", "key4"]
```

byroot:
Actually, there's 6 nodes in the test cluster, maybe I should simply wait on more of them.

byroot:
Ok, I'm trying a more aggressive version of that.

Another error that pop up from time to time is:

```
  1) Error:
TestClusterCommandsOnCluster#test_cluster_slots:
NoMethodError: undefined method `key?' for nil:NilClass
    /home/travis/build/redis/redis-rb/test/cluster_commands_on_cluster_test.rb:149:in `test_cluster_slots'
  2) Error:
TestClusterCommandsOnCluster#test_cluster_slaves:
NoMethodError: undefined method `fetch' for nil:NilClass
    /home/travis/build/redis/redis-rb/test/cluster_commands_on_cluster_test.rb:130:in `test_cluster_slaves'
```

So this I suppose is caused by some nodes being missing from the cluster?

supercaracal:
There might be a possibility that the cluster recovery after abnormal test cases is failure.
https://github.com/redis/redis-rb/blob/master/test/support/cluster/orchestrator.rb#L14-L26

But I'm not sure why that. I need time for looking into these failures. I'm sorry.

byroot:
> I need time for looking into these failures. I'm sorry.

No worries. After the `WAIT` addition I got 3 green builds in a row, which never happened before. So it might have been the same root cause.

Thanks a lot for the `WAIT` hint.

PR #853, opened 2019-05-18
Fix unstable test cases for CI

supercaracal:
Hi,

Several test cases are unstable. I tried to reproduce them. I found the following failure patterns.

```
SEED=61387    # TestClusterCommandsOnCluster#test_cluster_slaves
SEED=22998    # TestClusterClientReplicas#test_some_reference_commands_are_sent_to_slaves_if_needed
```

It tends to fail when the cluster test cases is executed earlier priority.

PR #854, opened 2019-05-20
Reduce redundant tarball downloads

supercaracal:
I found that `wget` command is executed four times in test. It might be better to use cache.

https://travis-ci.org/redis/redis-rb/jobs/534425900

### DIFF
--- a/makefile
+++ b/makefile
@@ -63,7 +63,7 @@ stop_slave:
 	@$(call kill-redis,${SLAVE_PID_PATH})
 
 start_slave: ${BINARY}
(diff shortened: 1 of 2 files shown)


makefile
```diff
@@ -63,7 +63,7 @@ stop_slave:
 	@$(call kill-redis,${SLAVE_PID_PATH})
 
 start_slave: ${BINARY}
-	${BINARY}\
+	@${BINARY}\
```
supercaracal: Sorry, I forgot about this when I requested #853.

Issue #855, opened 2019-05-21
ERR Client sent AUTH, but no password is set


alexesDev:
I tried to use https://cloud.yandex.com/docs/managed-redis/ and got `ERR Client sent AUTH, but no password is set` becouse `sentinel get-master-addr-by-name` must be call without password.

Is this a bug?

Monkey patch: https://gist.github.com/alexesDev/5b0fe8006775cb27a877b5a3766465c9

supercaracal:
Could you tell me about your environment.

1. Are you using the client version `4.1.1`?
2. What the managed Redis server's version is?
3. Are you specifying the `password` option in the client?
4. Are you configuring the `requirepass` and `masterauth` options in the servers?
5. Are you configuring the `sentinel auth-pass` option in the sentinel nodes?

alexesDev:
1. Are you using the client version 4.1.1? Yes
```
bundle list | grep redis 
* redis (4.1.1)
```
2. What the managed Redis server's version is?
redis_version:5.0.4

3. Are you specifying the password option in the client? Yes
```ruby
Sidekiq.configure_server do |config|
  config.redis = {
    url: 'redis://my-instance-name',
    password: 'my-password',
    sentinels: [{ host: ..., port: ... }],
  }
end
```
4. 5. I cannot see config files.

alexesDev:
Сonnection instructions
```
#!/bin/bash
host=$(redis-cli -h ****.mdb.yandexcloud.net -p 26379 sentinel get-master-addr-by-name my-instance name | head -n 1)
redis-cli -h $host -a %(password) ping
```
```
const Redis = require('ioredis');
const redis = new Redis({
    sentinels: [
        {host: '***.mdb.yandexcloud.net', port: 26379}
    ],
    name: 'my-instance',
    password: '<password>',
    role: 'master'
});

redis.set('foo', 'bar');
```

supercaracal:
| No | Sentinel Auth | Master or Slave Auth | redis-rb |
| --- | --- | --- | --- |
| 1 | enabled | enabled | The client works fine only if both passwords are same. |
| 2 | enabled | disabled | There might be no use case. |
| 3 | disabled | enabled | There is a bug. |
| 4 | disabled | disabled | The client works fine. |

I guess your env might be No.3.

I think it might be better to separate the password option into 2 types like as follows. (feature request)

```ruby
sentinels = [{ host: '127.0.0.1', port: 6400, password: 'foo' },
[...]

PR #856, opened 2019-05-21
Fix Sentinel authentication

supercaracal:
This is related to #855.

I modified the Sentinel authentication like separating the `password` option into for Sentinel and for Redis.

```ruby
sentinels = [{ host: '127.0.0.1', port: 6400, password: 'foo' },
             { host: '127.0.0.1', port: 6401, password: 'foo' }]

redis = Redis.new(host: 'mymaster', sentinels: sentinels, role: :master, password: 'bar')
```

* [Sentinel and Redis authentication](https://redis.io/topics/sentinel#sentinel-and-redis-authentication)
* [Configuring Sentinel instances with authentication](https://redis.io/topics/sentinel#configuring-sentinel-instances-with-authentication)

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -577,6 +577,7 @@ class Redis
             client = Client.new(@options.merge({
               :host => sentinel[:host],
               :port => sentinel[:port],
+              password: sentinel[:password],
               :reconnect_attempts => 0,
             }))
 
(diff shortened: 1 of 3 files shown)


lib/redis/client.rb
```diff
@@ -595,11 +596,6 @@ def sentinel_detect
           end
 
           raise CannotConnectError, "No sentinels available."
-        rescue Redis::CommandError => err
```
supercaracal: We can remove this error handling as workaround because the `password` option is separated.

PR #873, opened 2019-09-16
 Add mexists method that allow to check the existance of multiple keys

GustavoCaso:
When calling `exists` with multiple keys, the return value depending on the arguments could be either a Boolean or an Integer.

This PR aims to fix that by creating a new method called `mexists` that accepts an array of keys

Example of `exists`

```ruby
redis.set("foo", "world")
=> "OK"
[3] pry(main)> redis.exists(['foo', 'foo2'])
=> true
[4] pry(main)> redis.exists(['foo3', 'foo2'])
=> false
[5] pry(main)> redis.exists(['foo', 'foo2'])
=> true
[6] pry(main)> redis.set('foo2', 'world')
=> "OK"
[7] pry(main)> redis.exists(['foo', 'foo2'])
=> 2
```

Example of `exists`

```ruby
redis.set("foo", "world")
=> "OK"
[3] pry(main)> redis.mexists(['foo', 'foo2'])
=> 1
[4] pry(main)> redis.exists(['foo3', 'foo2'])
=> 0
[5] pry(main)> redis.set('foo2', 'world')
=> "OK"
[7] pry(main)> redis.exists(['foo', 'foo2'])
=> 2
```

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -559,6 +559,16 @@ class Redis
     end
   end
 
+  # Count existing keys.
+  #
+  # @param [Array<String>] key
+  # @return [Fixnum]
+  def mexists(*keys)
+    synchronize do |client|
+      client.call([:exists] + keys)
+    end
+  end
+
   # Find all keys matching the given pattern.
   #
   # @param [String] pattern
--- a/lib/redis/distributed.rb
+++ b/lib/redis/distributed.rb
@@ -174,6 +174,14 @@ class Redis
       node_for(key).exists(key)
     end
 
+    # Count existing keys.
+    def mexists(*args)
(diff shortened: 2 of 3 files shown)


rafaelfranca:
Tests are broken. Can you take a look?

byroot:
There's also a backward compatibility concern here, I think that's why previous similar PRs stalled:

  - https://github.com/redis/redis-rb/pull/728
  - https://github.com/redis/redis-rb/issues/698
  - https://github.com/redis/redis-rb/pull/741

byroot:
@GustavoCaso the CI failure is legit (`mexist` keys hashing to different servers)

GustavoCaso:
Looking into it

byroot:

PR #874, opened 2019-09-16
Handle permanent node failure for redis cluster

darshanime:
Currently, if a node fails permanently, the client will not recover even if an existing slave is promoted to replace the failed master. 
Only when the master rejoins the cluster (as a slave now) and responds with a `MOVED` response, will the client be able to serve requests again.

This PR refreshes the node-slot mapping on `CannotConnectError`, which will make the client route subsequent requests to the promoted slave of the failed master.

If `replica: true` is set, a new slave which replaces the old one will be able to serve requests.

### DIFF
--- a/lib/redis/cluster.rb
+++ b/lib/redis/cluster.rb
@@ -69,12 +69,7 @@ class Redis
     end
 
     def call(command, &block)
-      begin
-        send_command(command, &block)
-      rescue CannotConnectError
-        update_cluster_info!
-        raise
-      end
+      send_command(command, &block)
     end
 
     def call_loop(command, timeout = 0, &block)
@@ -231,6 +226,9 @@ class Redis
       else
         raise
       end
+    rescue CannotConnectError
+      update_cluster_info!
+      raise
     end
 
     def assign_redirection_node(err_msg)
--- a/lib/redis/cluster.rb
+++ b/lib/redis/cluster.rb
@@ -71,9 +71,9 @@ class Redis
     def call(command, &block)
       begin
         send_command(command, &block)
-      rescue CannotConnectError => error
+      rescue CannotConnectError
         update_cluster_info!
-        raise error
+        raise
       end
     end
 
--- a/lib/redis/cluster.rb
+++ b/lib/redis/cluster.rb
@@ -69,7 +69,12 @@ class Redis
     end
 
     def call(command, &block)
-      send_command(command, &block)
+      begin
+        send_command(command, &block)
+      rescue CannotConnectError => error
+        update_cluster_info!
+        raise error
+      end
     end
 
     def call_loop(command, timeout = 0, &block)
--- a/test/cluster_abnormal_state_test.rb
+++ b/test/cluster_abnormal_state_test.rb
@@ -18,11 +18,11 @@ class TestClusterAbnormalState < Minitest::Test
 
   def test_the_state_of_cluster_failover
     redis_cluster_failover do
-      100.times do |i|
+      10.times do |i|
         assert_equal 'OK', r.set("key#{i}", i)
       end
 
-      100.times do |i|
+      10.times do |i|
         assert_equal i.to_s, r.get("key#{i}")
       end
 
@@ -36,11 +36,11 @@ class TestClusterAbnormalState < Minitest::Test
         r.set('key0', 0)
       end
 
-      100.times do |i|
+      10.times do |i|
         assert_equal 'OK', r.set("key#{i}", i)
       end
 
-      100.times do |i|
+      10.times do |i|
         assert_equal i.to_s, r.get("key#{i}")
       end
 
--- a/test/cluster_abnormal_state_test.rb
+++ b/test/cluster_abnormal_state_test.rb
@@ -30,6 +30,24 @@ class TestClusterAbnormalState < Minitest::Test
     end
   end
 
+  def test_the_state_of_cluster_node_failure
+    redis_cluster_fail_master do
+      assert_raises(Redis::CannotConnectError, 'Error connecting to Redis on 127.0.0.1:7002') do
+        r.set('key0', 0)
+      end
+
+      100.times do |i|
+        assert_equal 'OK', r.set("key#{i}", i)
+      end
+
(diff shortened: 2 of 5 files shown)


lib/redis/cluster.rb
```diff
@@ -69,7 +69,12 @@ def with_reconnect(val = true, &block)
     end
 
     def call(command, &block)
-      send_command(command, &block)
+      begin
+        send_command(command, &block)
+      rescue CannotConnectError => error
+        update_cluster_info!
+        raise error
```
byroot: Please use a straight `raise` without argument, so that the original backtrace is preserved.

byroot: Which also means you won't have to assign `error`

test/cluster_abnormal_state_test.rb
```diff
@@ -30,6 +30,24 @@ def test_the_state_of_cluster_failover
     end
   end
 
+  def test_the_state_of_cluster_node_failure
+    redis_cluster_fail_master do
+      assert_raises(Redis::CannotConnectError, 'Error connecting to Redis on 127.0.0.1:7002') do
+        r.set('key0', 0)
+      end
+
+      100.times do |i|
```
byroot: do we need `100.times` ? CI is not particularly fast, so I'd rather not do too much unnecessary round trips.

test/support/cluster/orchestrator.rb
```diff
@@ -17,6 +17,10 @@ def initialize(node_addrs, timeout: 30.0)
     @timeout = timeout
   end
 
+  def restart_cluster_nodes
+    system("make --no-print-directory start_cluster > /dev/null 2>&1")
```
byroot: It's better to pass the command arguments separately to avoid creating a subshell:

```ruby
system('make', '--no-print-directory', 'start_cluster', out: File::NULL, err: File::NULL)
```

test/support/cluster/orchestrator.rb
```diff
@@ -36,6 +40,18 @@ def down
     reset_cluster(@clients)
   end
 
+  def fail_serving_master
+    master, slave = take_replication_pairs(@clients)
+    master.shutdown
+    max_attempts = 600
```
byroot: `600` seem excessive, again regarding CI time.

lib/redis/cluster.rb
```diff
@@ -69,7 +69,12 @@ def with_reconnect(val = true, &block)
     end
 
     def call(command, &block)
```
supercaracal: Should we also care the following method using in `#call_loop`, `#call_pipeline` and so on?

https://github.com/redis/redis-rb/blob/78bf94051d8176a1a5721b8647a1ceb730119ab2/lib/redis/cluster.rb#L215-L229

test/cluster_abnormal_state_test.rb
darshanime: updated this and other test 👍

lib/redis/cluster.rb
```diff
@@ -69,7 +69,12 @@ def with_reconnect(val = true, &block)
...
-      send_command(command, &block)
+      begin
+        send_command(command, &block)
+      rescue CannotConnectError
+        update_cluster_info!
```
supercaracal: Should we reject the `fail` node in the following code?

https://github.com/redis/redis-rb/blob/78bf94051d8176a1a5721b8647a1ceb730119ab2/lib/redis/cluster/node_loader.rb#L24-L32

```
$ make start_cluster
$ make create_cluster
$ telnet localhost 7000
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
shutdown
Connection closed by foreign host.

$ telnet localhost 7001
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
cluster nodes
$728
e261a9204067566d27f3b97b1418017ce65171f3 127.0.0.1:7001@17001 myself,master - 0 1568692342000 2 connected 5461-10922
f4093471beaad607406d1933d6e0eb7c7bdb4d90 127.0.0.1:7000@17000 master,fail - 1568692271534 1568692270531 1 disconnected
8613ac5796f3516b2eee2321cb82a78648fe0eb1 127.0.0.1:7003@17003 master - 0 1568692343143 10 connected 0-5460
0de04c346cf9c5b24a076e80073f6bfc4377a866 127.0.0.1:7004@17004 slave e261a9204067566d27f3b97b1418017ce65171f3 0 1568692344148 2 connected
0a1054fbb6504d392989a7492efbaf5731ee65cc 127.0.0.1:7002@17002 master - 0 1568692343545 3 connected 10923-16383
36352f5a62cb950e4d1b26c28adf5ddc47be7fe7 127.0.0.1:7005@17005 slave 0a1054fbb6504d392989a7492efbaf5731ee65cc 0 1568692343000 6 connected
```

In the above case, the `127.0.0.1:7000` node is failing.

lib/redis/cluster.rb
darshanime: good point, done 👍

test/support/cluster/orchestrator.rb
darshanime: tried with 10 and 100, the tests get flakey as cluster sometimes takes longer to recover (depending on `cluster-node-timeout` etc). Changing it to `500`, it shouldn't be a big problem since we return as soon as the failover is achieved.

lib/redis/cluster.rb
darshanime: hmm, the failed node may rejoin as a slave. Moreover, since the call is done for each node, it shouldn't fail even if we don't remove the failed node
```go
        nodes.each do |node|
          info = fetch_node_info(node)
          info.empty? ? next : break
        end
```

supercaracal: I see. :+1:

supercaracal:
@byroot I have reviewed. Thank you for mentioning.

byroot:
LGTM now. Thank you both!

darshanime:
Thanks. Can we cut a release for this please?

byroot:
4.1.3 is out.

PR #875, opened 2019-09-19
Raise FutureNotReady when trying to compare Redis::Future when the pipeline hasn't finsih executing

GustavoCaso:
When opening a pipeline and trying to compare the value of the command we get different behaviour depending on the comparison method we use.

This is happening because `Redis::Future` inherit from [`BasicObject`](https://ruby-doc.org/core-2.5.0/BasicObject.htm)

Here is a little script that proves it [gist](https://gist.github.com/GustavoCaso/8305eef685725a02bb547e4e7e1f291e)

I think the solution would be to raise `FutureNotReady`  when trying to compare a `Redis::Future` that hasn't resolved yet

### DIFF
--- a/lib/redis/pipeline.rb
+++ b/lib/redis/pipeline.rb
@@ -129,6 +129,7 @@ class Redis
 
   class Future < BasicObject
     FutureNotReady = ::Redis::FutureNotReady.new
+    COMPARISON_METHODS = %i(== != >= <=)
 
     attr_reader :timeout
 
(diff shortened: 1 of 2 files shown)


kirs:
The check makes sense to me 👍 

@byroot thoughts?

byroot:
Sorry but I don't get this.

You are not supposed to compare the future itself, but their value.

I don't want to make a slippery slope argument, but if we are delegating various comparison methods, why not all others?

I get how you got bit by this, but the same would have happened with any other `BasicObject` method.

byroot:
If anything I'd rather undefine `==` etc.

GustavoCaso:
Thanks, @byroot for the answers.

> but the same would have happened with any other BasicObject method.

Totally agree with you.

I will open a PR to remove `==` from `Redis::Future`

PR #876, opened 2019-09-19
Deprecate ==  and != from Redis::Future

GustavoCaso:
For more information please check https://github.com/redis/redis-rb/pull/875

Most developers when using `==` after calling a redis command they expect to compare the result returned from Redis.

Inside a pipelined, we are operating with `Redis::Future` instances that are `BasicObject`

`BasicObject#==` is doing identity comparison instead of value comparison which is less intuitive for developers.

Example:

```ruby
result = nil

redis.pipelined do 
  result = redis.set('foo', 'bar') == "OK"
end

puts result 
# => false
```

By removing the ability to use `==` with a redis future developers would get NoMethodError and this helps to have more consistent errors when operating with `Redis::Future` inside a pipelined block

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -5,6 +5,10 @@ require_relative "redis/errors"
 
 class Redis
 
+  def self.deprecate(message, trace = caller[0])
+    $stderr.puts "\n#{message} (in #{trace})"
+  end
+
   def self.current
     @current ||= Redis.new
   end
--- a/lib/redis/pipeline.rb
+++ b/lib/redis/pipeline.rb
@@ -132,14 +132,18 @@ class Redis
 
     attr_reader :timeout
 
-    # Avoid allowing people to use == inside a pipelined
-    undef :==
-
     def initialize(command, transformation, timeout)
       @command = command
       @transformation = transformation
       @timeout = timeout
       @object = FutureNotReady
+    end
+
+    def ==(*)
+      message = "The method == and != is deprecated for Redis::Future and will be removed in 4.2.0"
+      message << " - You probably meant to call .value == or .value !="
+      ::Redis.deprecate(message)
+      super
     end
 
     def inspect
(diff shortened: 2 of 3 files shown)


casperisfine:
A couple things:

  - I'm generally favorable to this, however `redis` being very widely used, I think it needs a deprecation period
  - There is also the `!=` method
  - People can get tricked if they use Yoda comparisons `"OK" == redis.set('foo', 'bar')`, but I guess that's fine.

GustavoCaso:
Example of trying to call the `==` or `!=` method 

```
bin/console
irb(main):001:0> redis = Redis.new
=> #<Redis client v4.1.3 for redis://127.0.0.1:6379/0>
irb(main):002:0> redis.pipelined do
irb(main):003:1* redis.set("foo", "bar") == 'OK"
irb(main):004:1' end
irb(main):005:1' ^C
irb(main):005:0> redis.pipelined do
irb(main):006:1* redis.set("foo", "bar") == 'OK'
irb(main):007:1> end

The method == and != is deprecated for Redis::Future and will be removed in 4.2.0 - You probably meant to call .value == or .value != (in /Users/gustavocaso/src/github.com/redis-rb/lib/redis/pipeline.rb:145:in `==')
=> ["OK"]
irb(main):008:0> redis.pipelined do
irb(main):009:1* redis.set("foo", "bar") != 'OK'
irb(main):010:1> end

The method == and != is deprecated for Redis::Future and will be removed in 4.2.0 - You probably meant to call .value == or .value != (in /Users/gustavocaso/src/github.com/redis-rb/lib/redis/pipeline.rb:145:in `==')
=> ["OK"]
irb(main):011:0>
```

GustavoCaso:
I didn't have to deprecate `!=` explicitly since `!=` uses `==` underneath and we will see two deprecated messages.

Issue #888, opened 2020-01-31
Cluster with auth connects and retrieves cluster nodes but without auth


zedtux:
I have a Rails application which uses this gem in order to connect to a single Redis instance with a password, where all is working fine.

I'm migrating Redis as a Redis Cluster, and therefore I have updated my config so that it no more passes `url: 'redis://:<password>@redis-host:6379/0'` but `cluster: ['redis://:<password>@a-redis-cluster-master-host:6379/0']` which works as the `Redis.new` returned this:

```
#<Redis client v4.1.3 for redis://10.244.5.253:6379/0 redis://10.244.7.251:6379/0 redis://10.244.7.252:6379/0>
```

But now, if I try to use it, it fails with a NOAUTH error :

```ruby
irb(main):005:0>redis = Redis.new(cluster: ['redis://:<password>@redis-cluster:6379/0'])
=> #<Redis client v4.1.3 for redis://10.244.5.253:6379/0 redis://10.244.7.251:6379/0 redis://10.244.7.252:6379/0>
irb(main):006:0> redis.ping
Traceback (most recent call last):
        1: from (irb):6
Redis::CommandError (NOAUTH Authentication required.)
irb(main):007:0> redis.auth '<password>'
=> "OK"
irb(main):008:0> redis.ping
=> "PONG"
```

So it seems that the password is used to open the connection and retrieve the cluster nodes, but then it is missing or not used to open the connection to the other nodes, something like that.

I tried to add the `redis.auth "<password>"` in my initialiser, but I still have the issue in my Rails app 😞.

Is there anything wrong am I doing? Can you please help me?

zedtux:
> I tried to add the redis.auth "<password>" in my initialiser, but I still have the issue in my Rails app 😞.

Actually I had an old version of the code on my server. So adding the `redis.auth "<password>"` in the Redis initialiser seem to work.

supercaracal:
This is a bug. I am sorry. I would like to fix it ASAP.

supercaracal:
Thank you for your reporting.

byroot:
@zedtux can you try master, just to confirm it properly fixes your issue?

supercaracal:
:pray:

MEMO:
I debugged in my local like the follows:

```diff
$ diff -u /tmp/makefile.orig makefile 
--- /tmp/makefile.orig  2020-02-01 21:47:19.499437323 +0900
+++ makefile    2020-02-01 21:44:54.487897316 +0900
[...]

PR #889, opened 2020-02-01
Fix cluster connecting option bug

supercaracal:
This is related #888.

I modified as keeping node information as Hash instead of URI String.

### DIFF
--- a/lib/redis/cluster.rb
+++ b/lib/redis/cluster.rb
@@ -116,7 +116,7 @@ class Redis
       [Node.new(option.per_node_key, node_flags, option.use_replica?),
        Slot.new(available_slots, node_flags, option.use_replica?)]
     ensure
-      node&.map(&:disconnect)
+      node&.each(&:disconnect)
     end
 
     def fetch_command_details(nodes)
--- a/lib/redis/cluster.rb
+++ b/lib/redis/cluster.rb
@@ -112,12 +112,11 @@ class Redis
       node = Node.new(option.per_node_key)
       available_slots = SlotLoader.load(node)
       node_flags = NodeLoader.load_flags(node)
-      available_node_urls = NodeKey.to_node_urls(available_slots.keys, secure: option.secure?)
-      option.update_node(available_node_urls)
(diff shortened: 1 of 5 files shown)


lib/redis/cluster.rb
```diff
@@ -112,12 +112,11 @@ def fetch_cluster_info!(option)
...
       available_slots = SlotLoader.load(node)
       node_flags = NodeLoader.load_flags(node)
-      available_node_urls = NodeKey.to_node_urls(available_slots.keys, secure: option.secure?)
-      option.update_node(available_node_urls)
+      option.update_node(available_slots.keys.map { |k| NodeKey.optionize(k) })
       [Node.new(option.per_node_key, node_flags, option.use_replica?),
        Slot.new(available_slots, node_flags, option.use_replica?)]
     ensure
-      node.map(&:disconnect)
+      node&.map(&:disconnect)
```
byroot: Nitpick, but since the return values doesn't matter, `.each` would make sense here.

lib/redis/cluster/option.rb
```diff
@@ -62,15 +60,27 @@ def parse_node_addr(addr)
...
+        addr
+      end
+
+      # Redis cluster node returns only host and port information.
+      # So we should complement additional information such as:
+      #   scheme, password and so on.
+      def add_common_node_option_if_needed(options, node_opts, key)
+        return options if options[key].nil? && node_opts.first[key].nil?
+
+        options[key] = options[key] || node_opts.first[key]
```
byroot: You can use `||=` here.

supercaracal:
> NoMethodError: undefined method `compact'

oh... Ruby 2.3 is EOL but I will accommodate it.

lib/redis/cluster.rb
supercaracal: fixed 584a8c6

lib/redis/cluster/option.rb
supercaracal: fixed 584a8c6

byroot:
Thanks!

supercaracal:
Thank you for your quick and polite reviewing.

PR #894, opened 2020-03-22
Extend SlotLoader to work with multiple slot ranges

rahul342:
With multiple resharding, different slot ranges can be assigned to a
node. 
e.g. 
Screenshot 2020-03-22 02 23 01

Current logic only preassigns the last slot range, causing errors & retries for unassigned slots. 
This commit makes it work with multiple slot ranges and assigning all slots.

I tested this locally, but couldn't figure how to add tests for this scenario.
**edit**: Added partial tests.

### DIFF
--- a/lib/redis/cluster/slot_loader.rb
+++ b/lib/redis/cluster/slot_loader.rb
@@ -13,7 +13,7 @@ class Redis
         info = {}
 
         nodes.each do |node|
-          info = Hash[*fetch_slot_info(node)]
+          info = fetch_slot_info(node)
           info.empty? ? next : break
         end
 
@@ -23,9 +23,11 @@ class Redis
       end
 
       def fetch_slot_info(node)
+        hash_with_default_arr = Hash.new { |h, k| h[k] = [] }
         node.call(%i[cluster slots])
-            .map { |arr| parse_slot_info(arr, default_ip: node.host) }
-            .flatten
+          .flat_map { |arr| parse_slot_info(arr, default_ip: node.host) }
+          .each_with_object(hash_with_default_arr) { |arr, h| h[arr[0]] << arr[1] }
+
       rescue CannotConnectError, ConnectionError, CommandError
         {} # can retry on another node
       end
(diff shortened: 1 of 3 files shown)


test/cluster_client_slots_test.rb
```diff
@@ -26,8 +26,36 @@ def test_slot_class
...
+    assert_equal '127.0.0.1:7000', slot.find_node_key_of_slave(30)
+    assert_equal '127.0.0.1:7000', slot.find_node_key_of_slave(40)
+    assert_nil slot.find_node_key_of_master(11)
+    assert_nil slot.find_node_key_of_slave(11)
+    assert_nil slot.find_node_key_of_master(41)
+    assert_nil slot.find_node_key_of_slave(41)
+
+    assert_nil slot.put(1, '127.0.0.1:7001')
+    assert_nil slot.put(30, '127.0.0.1:7001')
+  end
```
supercaracal: I'd say that a additional test case with `node_flags` and `with_replica` args is a nice-to-have. We want to make sure the scale read feature in multiple slot ranges case.

```ruby
Redis::Cluster::Slot.new({ '127.0.0.1:7000' => [1..10, 30..40], '127.0.0.1:7001' => [1..10, 30..40] },
                         { '127.0.0.1:7000' => 'master', '127.0.0.1:7001' => 'slave' }, true)
```

lib/redis/cluster/slot_loader.rb
```diff
@@ -23,9 +23,12 @@ def load(nodes)
       end
 
       def fetch_slot_info(node)
+        hash_with_default_arr = Hash.new { |h, k| h[k] = [] }
         node.call(%i[cluster slots])
-            .map { |arr| parse_slot_info(arr, default_ip: node.host) }
-            .flatten
+          .map { |arr| parse_slot_info(arr, default_ip: node.host) }
+          .flatten(1)
```
byroot: I think you can use `flat_map` rather than `map.flatten`

byroot:
Code looks sane to me. I'll merge once you add the test @supercaracal required.

rahul342:
@byroot done

rahul342:
cc @cliu55

byroot:
Thanks.

PR #897, opened 2020-04-22
Replace Fixnum with in Integer in methods docs

007lva:
Because Fixnum and Bignum were unified into Integer since Ruby 2.4 and now are deprecated, even [Ruby 2.4 is now EOL](https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/).


byroot:
Thanks.

PR #898, opened 2020-04-23
fix: do not allocate empty string

isqad:
Hello!

Here is my report:

Before:
```
Allocated String Report
-----------------------------------
    268824  ""
     38906  /bundle/2.2-0.9.0/gems/redis-4.1.3/lib/redis/connection/command_helper.rb:27
````

After my change:
```
Allocated String Report
-----------------------------------
    229421  ""
     77740  /localgems/redis-rb/lib/redis/connection/ruby.rb:43
     38906  /localgems/redis-rb/lib/redis/connection/ruby.rb:396
     38906  /localgems/redis-rb/lib/redis/connection/ruby.rb:59
```

As shown above difference is 39403 empty strings )

### DIFF
--- a/lib/redis/connection/command_helper.rb
+++ b/lib/redis/connection/command_helper.rb
@@ -1,8 +1,10 @@
+# frozen_string_literal: true
 class Redis
   module Connection
     module CommandHelper
 
-      COMMAND_DELIMITER = "\r\n"
+      COMMAND_DELIMITER = "\r\n".freeze
+      TRAILING_DELIMITER = "".freeze
 
       def build_command(args)
         command = [nil]
@@ -24,7 +26,7 @@ class Redis
         command[0] = "*#{(command.length - 1) / 2}"
 
         # Trailing delimiter
-        command << ""
+        command << TRAILING_DELIMITER
         command.join(COMMAND_DELIMITER)
       end
 

### CODE REVIEW

File: lib/redis/connection/command_helper.rb [hunk 1]
```diff
@@ -1,8 +1,10 @@
+# frozen_string_literal: true
 class Redis
   module Connection
     module CommandHelper
 
-      COMMAND_DELIMITER = "\r\n"
+      COMMAND_DELIMITER = "\r\n".freeze
```
byroot, line 6: You don't need these `.freeze` if you also put the `# frozen_string_literal: true` magic comment.

lib/redis/connection/command_helper.rb
```diff
@@ -24,7 +26,7 @@ def build_command(args)
         command[0] = "*#{(command.length - 1) / 2}"
 
         # Trailing delimiter
-        command << ""
+        command << TRAILING_DELIMITER
```
byroot, line 29: I actually wonder what's the point of appending an empty string...

isqad: Hmmm...) I don't know) May be remove the line?

byroot: So I dug the git history and it was added a very long time ago: https://github.com/redis/redis-rb/commit/dcb2498ce58daf276af2b0c7a76db295c842d8b7

So apparently it's so that after you join you have the delimiter at the end:

```ruby
['a', 'b'].join("\n") # => "a\nb"
['a', 'b', ''].join("\n") # => "a\nb\n"
```
So it kinda makes sense.

byroot:
What kind of code did you profile? I wonder if there was other sources of seemingly useless allocations.

Might be worth converting the entire gem to `# frozen_string_literal: true`

lib/redis/connection/command_helper.rb
```diff
@@ -1,8 +1,10 @@
+# frozen_string_literal: true
 class Redis
   module Connection
     module CommandHelper
 
-      COMMAND_DELIMITER = "\r\n"
+      COMMAND_DELIMITER = "\r\n".freeze
+      TRAILING_DELIMITER = "".freeze
```
byroot, line 7: I don't think it's worth naming it. Empty string literal is clear enough IMO.

isqad:
> What kind of code did you profile?

We have mailer that deliver emails in loop. For every envelope our mailer connect to redis and retrieve info about user.
I used memory_profiler gem for profile memory consuption.

isqad:
> Might be worth converting the entire gem to # frozen_string_literal: true

Nice idea, so I'll do it, ok?

byroot:
> We have mailer that deliver emails in loop.

Yeah, my question was more about the kind of paths you tested and the number of iteration. Because just based on the diff, your change saves 1 allocation per command. So I wonder how many commands you profiled, just to confirm my assumption is correct.

> I used memory_profiler gem for profile memory consuption.

I know, I'm quite familiar with that gem ;). 

> Nice idea, so I'll do it, ok?

If you feel like it, then yeah I'd merge such PR. I'd say close this one, and open a new one that only adds `# frozen_string_literal: true` to every files (including the test suite). And then fix whatever is broken.

protip: simply running `rubocop -a --only Style/FrozenStringLiteralComment` should do 99% of the work for you. The harder part is to fix the breakage if there's some.

PR #900, opened 2020-04-25
Fix TLS connections not verified by default

bpo:
If you specify an SSL URL for your client but do not specify some values in `ssl_params`, you'll end up with a connection that does not actually verify the certificate. This is surprising: usually network clients using TLS would use VERIFY_PEER with a system certificate store. This probably went undetected because domain verification still happens by default - it's just the certificate chain that is skipped.

`SSLContext#set_params` is the step in SSL connection establishment that
merges user-provided SSL parameters with "saner defaults".

Previously `SSLContext#set_params` was only being called when
`ssl_params` was provided to the SSLConnection, so the SSLContexet was
left unconfigured. This leads to the surprising default `verify_mode` of
`nil`, resulting in SSL connections verifying hostnames but never
verifying certificate trust.

### DIFF
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -263,7 +263,9 @@ class Redis
           tcp_sock = TCPSocket.connect(host, port, timeout)
 
           ctx = OpenSSL::SSL::SSLContext.new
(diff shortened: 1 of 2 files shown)


byroot:
Thanks

PR #901, opened 2020-04-30
Alias the disconnect method as close

mperham:
The connection_pool gem is implementing  automatic reaping of idle connections. In order to do this generically, clients need to provide a `close` method that the pool can call on reaped connections in order to close the network connection cleanly.

mysql2 and dalli already provide `close`. This PR adds this convention to redis-rb.

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -250,6 +250,7 @@ class Redis
     def disconnect
       connection.disconnect if connected?
     end
+    alias_method :close, :disconnect
 
     def reconnect
       disconnect


byroot:
> In order to do this generically, clients need to provide a `close` method that the pool can call on reaped connections in order to close the network connection cleanly.

That alias make sense, and I highly doubt it will ever conflict with a Redis command, so I'll merge.

But I wonder why `connection_pool` couldn't check for a few common names such as `disconnect`.

mperham:
Convention over configuration. `close` seems to be a widely accepted convention for the method we need to call. I don't want to maintain a mapping of clients to method names.

Redis already provides the [QUIT](https://redis.io/commands/quit) command which does what you say; I don't see why they would add CLOSE.

Issue #905, opened 2020-05-17
XCLAIM exception with nil result


eadz:
XCLAIM can return a nil array result.  ( `[nil]` )
It's passed to `HashifyStreamEntries` which then calls `each_slice` on nil causing an exception. 

In REDIS, the following command can return an array with a nil element. 
```sh
"xclaim" "mystream" "default" "3734fd61-ec3f-4bc1-b59d-3719d8ae40be" "1" "1589669825174-0"`
1) (nil)
```

eadz:
Closing until I can reproduce reliably.

eadz:
So I haven't managed to write a failing test, and my PR fix is no good, but I'm pretty sure there is a bug in `HashifyStreamEntries`

( for reference: )
```ruby
HashifyStreamEntries = lambda { |reply|
    reply.map do |entry_id, values|
      [entry_id, values.each_slice(2).to_h]
    end
  }
```

Arrays can have nil elements. The `reply` coming into the `HashifyStreamEntries` that caused the error is `[nil]`. 

<img width="1123" alt="Screen Shot 2020-05-22 at 7 38 06 PM" src="https://user-images.githubusercontent.com/11939/82643689-161ec780-9c64-11ea-90f8-7495222a8656.png">

```
# current HashifyStreamEntries
reply = [nil]
HashifyStreamEntries.call(reply) # => NoMethodError (undefined method `each_slice' for nil:NilClass)
```

Personally, I don't care about nil results, so I am using a fix of `compact` following fix:

```ruby
HashifyStreamEntries = lambda { |reply|
    reply.compact.map do |entry_id, values|
      [entry_id, values.each_slice(2).to_h]
    end
  }
```

eadz:
would a patch using compact be acceptable?

byroot:
Not without a reproduction.

eadz:
I thought I explained clearly how `HashifyStreamEntries` cannot handle an array with a nil value. 
It's self evident from the code. 

Is a reproduction necessary because you do not think Redis will return an array with a nil value? Even though that's the documented behavior?

byroot:
> Even though that's the documented behavior?

That's not the documented behavior. What the documentation says is that it returns an Array, then points to the RESPv3 (which this library doesn't use) generic description of an array, and say that an array can contain a `nil`.

The documentation doesn't say that XCLAIM can return nil.

[...]

PR #906, opened 2020-05-17
Fix XCLAIM exception on nil result

eadz:
See issue https://github.com/redis/redis-rb/issues/905

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Fix xclaim exception on [nil] response. See #905
+
 # 4.1.4
 
 * Alias `Redis#disconnect` as `#close`. See #901.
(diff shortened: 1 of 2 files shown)


byroot:
In which circumstances? Could you add a unit test to show it and ensure it won't be broken again in the future?

eadz:
@byroot Happy to add tests, but I was unable to see any tests that tests the lamdas, so a little unsure where to start. Are there any tests at all for those lamdas and/or respone processing from the streams stuff?

The issue #905 has the redis call that will trigger the error. You may need to create the stream and consumer group first, but then in that response you should get an array with [nil]. You can see how this will break the lamda as it tries to call `each_slice` on nil.

eadz:
Ok, I'm having trouble reproducing this. I'll close for now, and re-open when I can reproduce.

PR #908, opened 2020-05-21
Provide `on_disconnect` hook for connectors.

tgxworld:
When a [custom connector](https://github.com/redis/redis-rb/blob/f597f21a6b954b685cf939febbc638f6c803e3a7/lib/redis/client.rb#L94-L95) is provided, it has access to the client instance on connect by overriding the `Connector#check` instance method. In certain cases, knowing when a client has disconnected can also be useful. In particular, I'm building out a custom failover connector such that it `Connector#resolve` to the replica Redis config if it fails to connect to the master Redis server. As part of my connector, I'm keeping track of all Redis clients that are connected so that I can disconnect them when the master Redis server is back up. Disconnecting Redis clients will result in the [Redis client reconnecting](https://github.com/redis/redis-rb/blob/f597f21a6b954b685cf939febbc638f6c803e3a7/lib/redis/client.rb#L231) the next time it has to process a command. As I'm keeping track all Redis clients, I need a way to not track those clients once they have been disconnected so that the client object can be GC-ed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* `Redis::Client#disconnect` provides a `on_disconnect` connector hook.
+
 # 4.1.4
 
 * Alias `Redis#disconnect` as `#close`. See #901.
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -248,7 +248,11 @@ class Redis
     end
 
     def disconnect
-      connection.disconnect if connected?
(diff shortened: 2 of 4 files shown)


byroot:
If you'd like this to be merged, please write a proper PR description with your use case etc. 

Because right now I don't really see the point.

tgxworld:
@byroot Updated the PR description to describe my use case. Please let me know if anything is not clear to you. Thank you :bow:

PR #910, opened 2020-05-29
Implement variadic hset and mapped_hset

motoroller95:
Because `HMSET` is deprecated and `HSET` is variadic, I added support variadic parameters to `hset`, and added `mapped_hset` for convenience

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -2101,16 +2101,36 @@ class Redis
     end
   end
 
-  # Set the string value of a hash field.
+  # Set one or more hash values.
+  #
+  # @example
+  #   redis.hset("hash", "f1", "v1", "f2", "v2")
+  #     # => 2
   #
   # @param [String] key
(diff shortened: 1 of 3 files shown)


motoroller95:
@byroot Hi! What do you think about this MR?

byroot:
I'm planing to add it as part of the next release, I just need to take some time but I'll probably make some changes.

I don't like the distinct `mapped_hset`, I'd check the arguments received in `hset` to accept both a map and key pairs.

motoroller95:
I maked `mapped_hset` like `mapped_hmset`. why not?

byroot:
> why not?

Because it's not good for discoverability. Pretty much all methods maps directly to a redis command.

byroot:
Merged as https://github.com/redis/redis-rb/commit/ad7191f3a1ff8170bac6f61555ec8cf67fca4047

PR #911, opened 2020-06-01
Support multiple keys for `Redis#exists`.

tgxworld:
Since Redis 3.0.3, exists supports multiple keys and would return a
count of the keys that exists.


tgxworld:
Just saw https://github.com/redis/redis-rb/pull/873

PR #913, opened 2020-06-05
Support KEEPTTL option in SET command

ferrous26:
This option was added in Redis 6.0.0. See https://redis.io/commands/set for documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Increase buffer size in the ruby connector. See #880.
 * Fix thread safety of `Redis.queue`. See #878.
 * Deprecate `Redis::Future#==` as it's likely to be a mistake. See #876.
+* Support `KEEPTTL` option for SET command. See #913.
 
 # 4.1.3
 
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -786,6 +786,7 @@ class Redis
   #   - `:px => Integer`: Set the specified expire time, in milliseconds.
   #   - `:nx => true`: Only set the key if it does not already exist.
   #   - `:xx => true`: Only set the key if it already exist.
+  #   - `:keepttl => true`: Retain the time to live associated with the key.
   # @return [String, Boolean] `"OK"` or true, false if `:nx => true` or `:xx => true`
   def set(key, value, options = {})
     args = []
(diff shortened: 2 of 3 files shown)


ferrous26:
Hello, in the process of adding this small feature, I had to enable testing against redis 6. 

I'm not sure what the process for dealing with breaking behaviour changes, such as [this test](https://github.com/redis/redis-rb/pull/913/files#diff-daf87d478ef9af3b2624b7ad051015baR104). If this is scope creep, or not something that you accept in a PR, I will remove it.

byroot:
Hi @ferrous26, can you split this in two PRs ? First test against Redis 6.0, then after that another PR to add the new feature.

.travis.yml
```diff
@@ -43,6 +43,8 @@ matrix:
       env: DRIVER=ruby REDIS_BRANCH=3.2
     - rvm: "2.3"
       env: DRIVER=ruby REDIS_BRANCH=4.0
+    - rvm: "2.3"
+      env: DRIVER=ruby REDIS_BRANCH=6.0
```
byroot, line 47: It would be better to make 6.0 the default version in test, so that it's tested against all rubies.

test/cluster_commands_on_keys_test.rb
```diff
@@ -98,7 +98,13 @@ def test_touch
       set_some_keys
       assert_equal 1, redis.touch('key1')
       assert_equal 1, redis.touch('key2')
-      assert_equal 1, redis.touch('key1', 'key2')
+      if version < '6.0.0'
+        assert_equal 1, redis.touch('key1', 'key2')
```
byroot, line 102: WI doesn't if fail anymore ? Is it just the hashing that changed?

ferrous26: I am not sure. I am somewhat new to redis, and looking at the surrounding tests, I'm surprised that this was passing before.

A search through PRs in antirez/redis shows this change: 
https://github.com/antirez/redis/pull/5391/files

If I am understanding that correctly, the TOUCH command previously did not actually look past the first key argument.

byroot: That makes sense. Thanks!

byroot:
Thanks!

PR #918, opened 2020-06-10
Fix `Redis#exists?` to return boolean value for multiple key match.

tgxworld:
```
$ bin/console
irb(main):001:0> redis = Redis.new
irb(main):002:0> redis.exists?("test", "test5", "test2")
=> 2
```


tgxworld:
@byroot I was trying to upgrade to 4.2 and noticed `Redis#exists?` returning the integer when more than one key matches. 

The test failure here is unrelated.

byroot:
Thanks.

PR #920, opened 2020-06-11
Add ability to silence `exists` warning message.

simonrussell:
Related to #698.

The warning message can be disabled by setting `exists_returns_integer` to `false`.  You will receive a single warning message when this is set.

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -5,7 +5,20 @@ require_relative "redis/errors"
 
 class Redis
   class << self
-    attr_accessor :exists_returns_integer
+    attr_reader :exists_returns_integer
+
+    def exists_returns_integer=(value)
+      unless value
+        message = "`Redis#exists(key)` will return an Integer by default in redis-rb 4.3. The option to explicitly " \
+          "disable this behaviour via `Redis.exists_returns_integer` will be removed in 5.0.  You should use " \
+          "`exists?` instead. "
+
+        ::Kernel.warn(message)
+      end
+
+      @exists_returns_integer = value
+    end
+
     attr_writer :current
   end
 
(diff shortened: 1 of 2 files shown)


simonrussell:
I'm not too sure what that broken test is relating to.

lib/redis.rb
```diff
@@ -5,7 +5,20 @@
 
 class Redis
   class << self
-    attr_accessor :exists_returns_integer
+    attr_reader :exists_returns_integer
+
+    def exists_returns_integer=(value)
+      unless value
+        message = "`Redis#exists(key)` will return an Integer in redis-rb 4.3. The option to explicitly disable this " \
```
byroot, line 12: Let's keep that flag until 5.0, that should solve most of the semver concerns.

byroot: > exists(key)` will return an Integer **by default** in redis-rb 4.3.

> via `Redis.exists_returns_integer` will be removed **in 5.0**.

lib/redis.rb
```diff
@@ -5,7 +5,20 @@
...
+          "behaviour via `Redis.exists_returns_integer` will be removed.  You should use `exists?` instead. " \
+          "(#{::Kernel.caller(1, 1).first})\n"
```
byroot, line 14: I don't think we need to print the caller for this.

lib/redis.rb
```diff
@@ -561,11 +574,16 @@ def unlink(*keys)
   # @return [Integer]
   def exists(*keys)
     if !Redis.exists_returns_integer && keys.size == 1
-      message = "`Redis#exists(key)` will return an Integer in redis-rb 4.3, if you want to keep the old behavior, " \
-        "use `exists?` instead. To opt-in to the new behavior now you can set Redis.exists_returns_integer = true. " \
-        "(#{::Kernel.caller(1, 1).first})\n"
+      unless Redis.exists_returns_integer == false
```
byroot, line 577: `if Redis.exists_returns_integer.nil?`

byroot:
Looks good. Please squash your commits together and I'll merge.

byroot:
> I'm not too sure what that broken test is relating to.

Looks like rubocop is unhappy. Some lines are too long.

simonrussell:
> Looks like rubocop is unhappy. Some lines are too long.

This was actually in relation to actual test failures in the pubsub ruby stuff, looked like duplicate things were in there.

But yes, seems that commit made Rubocop unhappy for other reasons.  Normally I'd set up a local dev environment but I'm just relying on CI for this :)

byroot:
> This was actually in relation to actual test failures in the pubsub ruby stuff

Unfortunately we have a couple flaky tests in the suite. It's getting better but it's really hard to stabilize, lot's of moving parts with the distributed redis testing.

simonrussell:
Right I thought it might be something like that.  I think anything complex has some flaky tests somewhere...

PR #923, opened 2020-06-15
Add `Redis#hset` change to the 4.2.0 changelog

davidtaylorhq:
Since https://github.com/redis/redis-rb/commit/ad7191f3a1ff8170bac6f61555ec8cf67fca4047, `Redis#hset` now returns an integer indicating the number of fields added to the hash. Previously, it would return a boolean. 

This PR adds this change of behaviour to the 4.2.0 changelog.


byroot:
Thanks, totally forgot that one -_-.

PR #926, opened 2020-06-17
Use the exception: false API of read/write nonblock

casperisfine:
Should be a nice optimization, just need to make sure it doesn't break anything (`Errno::EAGAIN` ,`Errno::EWOULDBLOCK` in connect is fishy).

### DIFF
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -49,43 +49,47 @@ class Redis
       end
 
       def _read_from_socket(nbytes)
-        begin
-          read_nonblock(nbytes)
-        rescue IO::WaitReadable
-          if IO.select([self], nil, nil, @timeout)
-            retry
-          else
-            raise Redis::TimeoutError
-          end
-        rescue IO::WaitWritable
-          if IO.select(nil, [self], nil, @timeout)
-            retry
-          else
-            raise Redis::TimeoutError
+        loop do
+          case chunk = read_nonblock(nbytes, exception: false)
+          when :wait_readable
+            unless IO.select([self], nil, nil, @timeout)
+              raise Redis::TimeoutError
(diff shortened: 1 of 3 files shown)


lib/redis/connection/ruby.rb
```diff
@@ -49,39 +49,43 @@ def gets
...
+            end
+          when :wait_writable
+            unless IO.select(nil, [self], nil, @timeout)
+              raise Redis::TimeoutError
+            end
+          when String
+            return chunk
           end
         end
       rescue EOFError
```
supercaracal, line 66: https://docs.ruby-lang.org/en/2.7.0/IO.html#method-i-read_nonblock

> At EOF, it will return nil instead of raising EOFError.

Should we handle `nil` at `case` statement?

casperisfine:
The connect_noblock is a bit trickier to handle, so I reduce the PR to just read and write.

lib/redis/connection/ruby.rb
```diff
@@ -49,43 +49,47 @@ def gets
...
-        rescue IO::WaitReadable
-          if IO.select([self], nil, nil, @write_timeout)
-            retry
-          else
-            raise Redis::TimeoutError
+        total_bytes_written = 0
+        loop do
+          case bytes_written = write_nonblock(data, exception: false)
+          when :wait_readable
+            unless IO.select([self], nil, nil, @timeout)
```
eregon, line 75: Should this still use `@write_timeout` (and not `@timeout`)?

byroot: Indeed. Good catch.

byroot: Fixed in 087a11b585978cd3970d22d066b7d8ccd89a40f6

PR #930, opened 2020-06-21
Fix several unstable test cases

supercaracal:
Closes #917

PR #941, opened 2020-09-04
Add transaction support for redis distributed

EiNSTeiN-:
I am trying to use `watch` and `multi` with `Redis::Distributed`. My code looks like this:

```ruby
r.watch("{uuid}:data", "{uuid}:version") do
  current = r.get("{uuid}:version")
  r.multi do
    r.set("{uuid}:data", new_data)
    r.set("{uuid}:version", current + 1)
  end
end
```

Given all keys can be tagged, they will all end up on the same node, so this should work with `Redis::Distributed` without issue. An additional assumption is that the caller will use the block syntax for `watch` which is thread safe, or will handle thread safety in another way outside of `Redis::Distributed`.

I implemented that by keeping a copy of the key tag or the key if the key itself isn't tagged, and ensure that all commands are issued with the same key or key tag until one of these things happen:
1) `exec`, `unwatch` or `discard` is explicitly called
2) `exec` is implicitly called at the end of the `multi` block
3) `unwatch` is implicitly called when a `StandardError` is raised within a `watch` block

### DIFF
--- a/lib/redis/distributed.rb
+++ b/lib/redis/distributed.rb
@@ -30,7 +30,8 @@ class Redis
     def node_for(key)
       key = key_tag(key.to_s) || key.to_s
       raise CannotDistribute, :watch if @watch_key && @watch_key != key
-      node = @ring.get_node(key)
+
+      @ring.get_node(key)
     end
 
     def nodes
@@ -818,6 +819,7 @@ class Redis
     # Forget about all watched keys.
     def unwatch
       raise CannotDistribute, :unwatch unless @watch_key
+
       result = node_for(@watch_key).unwatch
       @watch_key = nil
       result
@@ -830,6 +832,7 @@ class Redis
     # Mark the start of a transaction block.
     def multi(&block)
       raise CannotDistribute, :multi unless @watch_key
+
       result = node_for(@watch_key).multi(&block)
       @watch_key = nil if block_given?
       result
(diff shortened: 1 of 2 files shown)


supercaracal:
CI tests of the repository tend to be flaky now. These irrelevant failures are ignorable. They will be resolved by #930.

lib/redis/distributed.rb
```diff
@@ -799,32 +803,57 @@ def punsubscribe(*channels)
...
-    def watch(*_keys)
-      raise CannotDistribute, :watch
+    def watch(*keys, &block)
+      ensure_same_node(:watch, keys) do |node|
+        @watch_key = key_tag(keys.first) || keys.first.to_s
+
+        begin
+          node.watch(*keys, &block)
+        rescue StandardError
+          @watch_key = nil
```
byroot, line 813: That's not good enough, you want to reset `@watch_key` from an `ensure` block. Here you only reset it on error.

byroot (changes requested):

lib/redis/distributed.rb
byroot: Hum, actually scratch that, it seems `watch do` is not expected to unwatch... that's a very weird behavior :/

byroot (approved):
Ok, so ignore my previous comment. I wasn't so familiar with the watch interface. Turns out the unwatch has to be manual, sounds like a terrible API to me but hey...

PR #945, opened 2020-09-10
fix: corruption of client during interrupt

trcarden:
Due to the fact that `original` is hoisted to nil and its assignment
is not the first operation interrupts can corrupt the redis client
pointer.

This commit addresses issue by using a reference guaranteed to
be set correctly to reassign the client in the ensure block

### Steps to reproduce

1. Connect to redis via client 
2. Raise interrupt during either `pipelined` or `multi` (Timeout.timeout, sigint, etc, pick your poison)
3. Ensure the interrupt happens before the assignment of `original = @client` (see test cases for example)

### Expected behavior
The library is resilient to interrupts and the redis client can continue to accept commands afterwards

### Actual behavior
The underlying `client` is net to nil and is corrupted forever after that raising undefined method calls 

### System configuration
**Ruby version**: 2.7.1
**Gem version**: 4.2.1

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -2438,14 +2438,13 @@ class Redis
   end
 
   def pipelined
-    synchronize do |_client|
+    synchronize do |prior_client|
       begin
-        pipeline = Pipeline.new(@client)
-        original, @client = @client, pipeline
+        @client = Pipeline.new(prior_client)
         yield(self)
-        original.call_pipeline(@client)
+        prior_client.call_pipeline(@client)
       ensure
-        @client = original
+        @client = prior_client
       end
     end
   end
@@ -2481,17 +2480,16 @@ class Redis
   # @see #watch
   # @see #unwatch
   def multi
-    synchronize do |client|
+    synchronize do |prior_client|
       if !block_given?
-        client.call([:multi])
+        prior_client.call([:multi])
       else
         begin
-          pipeline = Pipeline::Multi.new(@client)
-          original, @client = @client, pipeline
+          @client = Pipeline::Multi.new(prior_client)
           yield(self)
-          original.call_pipeline(pipeline)
+          prior_client.call_pipeline(@client)
         ensure
-          @client = original
+          @client = prior_client
         end
       end
     end
--- a/test/transactions_test.rb
+++ b/test/transactions_test.rb
@@ -196,6 +196,13 @@ class TestTransactions < Minitest::Test
     assert_equal "s1", r.get("foo")
   end
 
+  def test_multi_with_interrupt_preserves_client
+    original = r._client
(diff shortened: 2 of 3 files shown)


test/transactions_test.rb
```diff
@@ -196,6 +196,16 @@ def test_multi_with_a_block_yielding_the_client
     assert_equal "s1", r.get("foo")
   end
 
+  def test_multi_with_interrupt_preserves_client
+    original = r._client
+    begin
+      Redis::Pipeline::Multi.stubs(:new).raises(Interrupt)
+      r.multi {}
+    rescue Interrupt
```
byroot, line 204: It would be better to use `assert_raises Interrupt` as to ensure that your stub is indeed called.

lib/redis.rb
```diff
@@ -2445,7 +2445,7 @@ def pipelined
         yield(self)
         original.call_pipeline(@client)
       ensure
-        @client = original
+        @client = _client
```
byroot, line 2448: I'm not totally sure this is correct. I'll need to have a look later, but this always restore the original client, so if you nest two pipeline/multi, that might cause issues. That being said I'm not so sure such nesting is properly handled either...

test/transactions_test.rb
trcarden: Roger, changed.

lib/redis.rb
trcarden: @byroot I think if you take a look at `synchronize` it references the same instance variable `@client` and passes it in as either `_client` or `client` in `pipelined` and `multi` (respectively). Given that is the same value and all the nested calls would reference the same instance value via the `synchronize` call I think it checks out. Test suite passes too which is a good sign. 

```
 def synchronize
    mon_synchronize { yield(@client) }
  end
```

Obviously could do some variable name cleanup and remove duplication but thought to keep the changeset minimal until a maintainer could take a look.

byroot: > I think if you take a look at `synchronize`

Thanks, that's exactly the kind of thing I didn't have fresh in mind and meant to check tomorrow. So yeah,, LGTM.

> could do some variable name cleanup and remove duplication

Please do.

>  until a maintainer could take a look

One is talking with you right now 😉

byroot (approved):
Looks good. Please squash your commits together and I'll merge.

trcarden:
@byroot ok i tried to give you a few options to pick from ....

haha. I guess you liked the last one then? I just squashed them to the original commit and flattened it to follow your other commits for the project. Feel free to modify as you need. I don't mind.

byroot:
> ok i tried to give you a few options to pick from ....

Sorry, I didn't notice. I don't have commits push notifications on, it's way too noisy. Anything last version was fine.

byroot:
Thanks for the patch.

PR #951, opened 2020-10-06
Fix #950

### LINKED ISSUE #950: Wrong client default options
### Steps to repro
```
REDIS_URL=rediss://:secret@127.0.0.1:1234 irb
```
```ruby
require 'redis'
Redis::Client.new(host: '127.0.0.2')
```

### Expected
A client with default port 6379 with no password

### Actual
A client with the special port and password from the default REDIS_URL
```
 => #<Redis::Client:0x00007f9819106090 @options={:host=>"127.0.0.2", :url=>"rediss://:secret@127.0.0.1:1234", :scheme=>"rediss", :port=>1234, :path=>nil, :timeout=>5.0, :password=>"secret", :db=>0, :driver=>Redis::Connection::Ruby, :id=>nil, :tcp_keepalive=>0, :reconnect_attempts=>1, :reconnect_delay=>0.0, :reconnect_delay_max=>0.5, :inherit_socket=>false, :sentinels=>nil, :role=>:master, :ssl=>true, :connect_timeout=>5.0, :read_timeout=>5.0, :write_timeout=>5.0, :_parsed=>true}, @reconnect=true, @logger=nil, @connection=nil, @command_map={}, @pending_reads=0, @connector=#<Redis::Client::Connector:0x00007f9819105938 @options={:host=>"127.0.0.2", :url=>"rediss://:secret@127.0.0.1:1234", :scheme=>"rediss", :port=>1234, :path=>nil, :timeout=>5.0, :password=>"secret", :db=>0, :driver=>Redis::Connection::Ruby, :id=>nil, :tcp_keepalive=>0, :reconnect_attempts=>1, :reconnect_delay=>0.0, :reconnect_delay_max=>0.5, :inherit_socket=>false, :sentinels=>nil, :role=>:master, :ssl=>true, :connect_timeout=>5.0, :read_timeout=>5.0, :write_timeout=>5.0, :_parsed=>true}>>
```

This behavior is especially confusing when using Redis cluster, where REDIS_URLs are explicitly given but it's using the wrong default values to fill the client options.

donaldong:
This is an attempt to address #950.

It's probably not a good idea to grep pieces of the default REDIS_URL when it's present. The REDIS_URL could be used when no options are given: `Redis::Client.new({})`

Thoughts?

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -415,7 +415,7 @@ class Redis
       end
 
       url = options[:url]
-      url = defaults[:url] if url.nil?
+      url = defaults[:url] if url.nil? && options.empty?
 
       # Override defaults from URL if given
       if url


lib/redis/client.rb
```diff
@@ -415,7 +415,7 @@ def _parse_options(options)
       end
 
       url = options[:url]
-      url = defaults[:url] if url.nil?
+      url = defaults[:url] if url.nil? && options.empty?
```
byroot, line 418: I'm afraid this might break it for many users. Especially ones relying on the default URL, but adding options like `timeout` etc.

donaldong: Good point.

To this is how it could work:
- if the options overlap with any information that can be derived from REDIS_URL, use options as the source of truth and ignore REDIS_URL
- when all options and information derived from REDIS_URL can coexist, put them together

Partially overwriting information from the REDIS_URL is a rather strange behavior IMHO

byroot: > Partially overwriting information on top of the REDIS_URL is a rather strange behavior IMHO

Yeah, there is already a bunch of open issues about that. Nobody is really happy with it, but it's almost impossible to change it without breaking meany users.

So it will have to yet for the next major.

byroot:
This is fixed on master (upcoming 5.0). If either `host`, `port` or `path` is passed `$REDIS_URL` is ignored.

PR #963, opened 2020-11-19
DEV: add test case covering partial writes

SamSaffron:
This moves the responsibility of copying data into write_to_socket. The
caller never mutates the string, this is cleaner cause we do no longer
mutate the string we pass in.

### DIFF
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -68,6 +68,7 @@ class Redis
       end
 
       def _write_to_socket(data)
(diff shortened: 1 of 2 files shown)


byroot:
The test looks good.

> The caller never mutates the string

I think it probably should for performance reason. Using `slice!` we could avoid creating lots of strings. But I'd need to setup a benchmark.

byroot:
Closing in favor of https://github.com/redis/redis-rb/pull/964

Issue #962, opened 2020-11-18
_write_to_socket is changing input, which breaks write


SamSaffron:
This code is amending the input of the data

https://github.com/redis/redis-rb/blob/d896ae2c7391ba9af667ee3ec8f70192e8daf1da/lib/redis/connection/ruby.rb#L87-L87

And so is this code: 

https://github.com/redis/redis-rb/blob/d896ae2c7391ba9af667ee3ec8f70192e8daf1da/lib/redis/connection/ruby.rb#L106-L106

Simplest fix here is simply to stop the `slice!` call in `_write_to_socket` 

This is very urgent @byroot latest release is broken. Ideally we should have some sort of test to catch this.

End result is that if you are lucky, partial things are shipped to redis, if you are unlucky stuff explodes on:

```
NoMethodError: undefined method `bytesize' for nil:NilClass
  /var/www/discourse/vendor/bundle/ruby/2.6.0/gems/redis-4.2.3/lib/redis/connection/ruby.rb:103:in `block in write'
  /var/www/discourse/vendor/bundle/ruby/2.6.0/gems/redis-4.2.3/lib/redis/connection/ruby.rb:102:in `loop'
  /var/www/discourse/vendor/bundle/ruby/2.6.0/gems/redis-4.2.3/lib/redis/connection/ruby.rb:102:in `write'
  /var/www/discourse/vendor/bundle/ruby/2.6.0/gems/redis-4.2.3/lib/redis/connection/ruby.rb:394:in `write'
  /var/www/discourse/vendor/bundle/ruby/2.6.0/gems/redis-4.2.3/lib/redis/client.rb:284:in `block in write'
  /var/www/discourse/vendor/bundle/ruby/2.6.0/gems/redis-4.2.3/lib/redis/client.rb:263:in `io'
  /var/www/discourse/vendor/bundle/ruby/2.6.0/gems/redis-4.2.3/lib/redis/client.rb:282:in `write'
  /var/www/discourse/vendor/bundle/ruby/2.6.0/gems/redis-4.2.3/lib/redis/client.rb:240:in `block (3 levels) in process'
  /var/www/discourse/vendor/bundle/ruby/2.6.0/gems/redis-4.2.3/lib/redis/client.rb:234:in `each'
  /var/www/discourse/vendor/bundle/ruby/2.6.0/gems/redis-4.2.3/lib/redis/client.rb:234:in `block (2 levels) in process'
[...]

PR #964, opened 2020-11-19
Optimize SocketMixin#write to avoid string copying

casperisfine:
Ref: https://github.com/redis/redis-rb/issues/962#issuecomment-730287111

The key thing here is that `byteslice` is smart enough to create shared strings:

```ruby
>> puts ObjectSpace.dump(PAYLOAD.byteslice(4096..-1).byteslice(4096..-1).byteslice(4096..-1))
{"address":"0x7fd1c93e0178", "type":"STRING", "class":"0x7fd1c90b7578", "shared":true, "encoding":"UTF-8", "references":["0x7fd1c93a3778"], "memsize":40, "flags":{"wb_protected":true}}
```

In such case only a `RString` is instantiated, the content isn't copied, so it's way faster. This was benched as 75x faster, however it's a synthetic benchmark without IOs.

cc @SamSaffron maybe you'd like to take a look?

### DIFF
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -67,10 +67,13 @@ class Redis
         end
       end
 
-      def _write_to_socket(data)
+      def write(buffer)
+        return super(buffer) unless @write_timeout
+
+        bytes_to_write = buffer.bytesize
         total_bytes_written = 0
         loop do
-          case bytes_written = write_nonblock(data, exception: false)
+          case bytes_written = write_nonblock(buffer, exception: false)
           when :wait_readable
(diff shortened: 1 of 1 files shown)


SamSaffron:
This is a fantastic change, I very much support it! 

It reduces a bunch of code duplication and makes stuff much faster for large payloads which is awesome.

When I looked at this yesterday it "seemed" like stuff was structured in a way that we allowed people to override the _write_to_socket method, I think this is probably a bit too much flexibility especially it kind of follows from the naming that this is a private thing, but it may break some people so it is something to watch for.

That said, I 100% support this change and think we should go ahead with it!!!

Issue #965, opened 2020-11-26
Cannot use zincrby inside multi


souljuse:
I am trying to use a zincrby inside a multi statement but it gives me an error.

To reproduce:

```ruby
redis.multi
redis.zincrby("foo", 1,  "bar")
redis.exec

```

It gives me this error 

```
ArgumentError: invalid value for Float(): "QUEUED"
from .../redis.rb:3394:in `Float'
```

byroot:
Uh, I'm surprised this would even work, but indeed seems like the interface authorize it.

In the mean time I'd say use the block API:

```ruby
redis.multi do
  redis.zincrby ...
end
```

byroot:
Fixed in https://github.com/redis/redis-rb/commit/086221ec109835b3051b3f122820c370ad758014

byroot:
Actually, I figured since it do work for commands that don't parse the response, removing it might break some people's code. 

So instead I pushed https://github.com/redis/redis-rb/commit/c4be8f49c6d18f1e5ee2d2c8dbf7e4c94fa109ad

PR #966, opened 2020-11-26
Allow zincrby to be used in multi transaction block

souljuse:
Linked to issue #965


byroot:
Sorry but this is an ad hoc fix for just the specific command you use. If I were to merge a fix it would need to be more general than that.

byroot:
Fixed in https://github.com/redis/redis-rb/commit/086221ec109835b3051b3f122820c370ad758014

PR #975, opened 2021-02-04
fix driver failover

masato-hi:
The failover code seems to be incorrect if the driver is not found in require_relative.

ref: https://github.com/redis/redis-rb/commit/f3bc68f4305792a14c86556832627e13c9bbd6b0#diff-5bc007010e6c2e0aa70b64d6f87985c20986ee1b2882b63a89b52659ee9c91f8L520


byroot:
Agreed, looks like a bug.

Issue #978, opened 2021-02-24
Add support for Redis 6.2


rhymes:
[Redis 6.2](https://redis.io/download) has officially been released, its [release notes](https://raw.githubusercontent.com/redis/redis/6.2/00-RELEASENOTES) contain quite a few new commands, improvements and some changes in behavior.

I'm new to this library but I'm willing to help. I'd imagine the first step is to add `Redis 6.2` to the GitHub test actions and have the existing tests pass before starting to tackle new things.

I opened a "general" issue, but I guess this should be then split into separate issues/PRs for new commands and features.

This for example is list of new commands:

```text
Here is a comprehensive list of changes in this release compared to 6.0.9,
each one includes the PR number that added it, so you can get more details
at https://github.com/redis/redis/pull/<number>

New commands / args:
* Add SMISMEMBER command that checks multiple members (#7615)
* Add ZMSCORE command that returns an array of scores (#7593)
* Add LMOVE and BLMOVE commands that pop and push arbitrarily (#6929)
* Add RESET command that resets client connection state (#7982)
* Add COPY command that copies keys (#7953)
* Add ZDIFF and ZDIFFSTORE commands (#7961)
* Add ZINTER and ZUNION commands (#7794)
* Add GEOSEARCH/GEOSEARCHSTORE commands for bounding box spatial queries (#8094)
* Add GET parameter to SET command, for more powerful GETSET (#7852)
* Add exclusive range query to XPENDING (#8130)
* Add exclusive range query to X[REV]RANGE (#8072)
* Add GT and LT options to ZADD for conditional score updates (#7818)
* Add CLIENT INFO and CLIENT LIST for specific ids (#8113)
* Add IDLE argument to XPENDING command (#7972)
* Add local address to CLIENT LIST, and a CLIENT KILL filter. (#7913)
* Add NOMKSTREAM option to XADD command (#7910)
* Add command introspection to Sentinel (#7940)
* Add SENTINEL MYID subcommand (#7858)
* Add the REV, BYLEX and BYSCORE arguments to ZRANGE, and the ZRANGESTORE command (#7844)
* Add the XAUTOCLAIM command (#7973)
* Add the MINID trimming strategy and the LIMIT argument to XADD and XTRIM (#8169)
* Add the ANY argument to GEOSEARCH and GEORADIUS (#8259)
* Add the CH, NX, XX arguments to GEOADD (#8227)
* Add the COUNT argument to LPOP and RPOP (#8179)
[...]

PR #979, opened 2021-02-24
Add Redis 6.2 to the GitHub Actions test suite

rhymes:
Refers to https://github.com/redis/redis-rb/issues/978

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -87,7 +87,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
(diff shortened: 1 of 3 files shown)


byroot:
Sorry but that's not the way to do it. Testing all redis versions against all ruby version creates a giant matric that takes agests to complete.

Just set Redis version to 6.2 in the original main matrix, and add 6.0 in other matrix.

rhymes:
@byroot, will do! Thanks for the feedback!

byroot:
Might want to rebase, I pushed a CI fix.

rhymes:
@byroot sorry for taking so long to circle back on this. I fixed the errors :-)

PR #985, opened 2021-05-08
[Redis 6.0+] Add the TYPE argument to `scan`and `scan_each`

### LINKED ISSUE #984: Redis SCAN TYPE Option not supported
Currently the `redis-rb` gem does not support the `TYPE` option in `SCAN`: https://redis.io/commands/scan#the-type-option. The current definition of `scan` in the gem:

```Ruby
  # Scan the keyspace
  #
  # ...
  #
  # @param [String, Integer] cursor the cursor of the iteration
  # @param [Hash] options
  #   - `:match => String`: only return keys matching the pattern
  #   - `:count => Integer`: return count keys at most per iteration
  #
  # @return [String, Array<String>] the next cursor and all found keys
  def scan(cursor, **options)
    _scan(:scan, cursor, [], **options)
  end
```

If I have some time soon, I'll take a stab at this.

tednology:
This PR adds support for the Redis [`TYPE` option in `SCAN`](https://redis.io/commands/scan#the-type-option), available since 6.0.

Resolves #984.

### DIFF
--- a/test/scanning_test.rb
+++ b/test/scanning_test.rb
@@ -53,6 +53,25 @@ class TestScanning < Minitest::Test
     end
   end
 
+  def test_scan_type
+    target_version "6.0.0" do
+      r.debug :populate, 1000
+      r.zadd("foo", [1, "s1", 2, "s2", 3, "s3"])
+      r.zadd("bar", [6, "s1", 5, "s2", 4, "s3"])
+      r.hset("baz", "k1", "v1")
+
+      cursor = 0
+      all_keys = []
+      loop do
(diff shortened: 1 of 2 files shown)


test/scanning_test.rb
```diff
@@ -53,6 +53,25 @@ def test_scan_match
     end
   end
 
+  def test_scan_type
+    target_version "2.7.105" do
```
byroot, line 57: You meant 6.0 right?

tednology: Woops, yes! Will amend.

tednology: Done.

Issue #989, opened 2021-06-11
[v4.3.0] Redis::CommandError: ERR wrong number of arguments for 'auth' command


lucascaton:
Hi there,

I just updated to [version 4.3.0](https://github.com/redis/redis-rb/releases/tag/v4.3.0) and Sidekiq fails to start because of something related to `Redis::CommandError`.
Here's the full stack trace:

```
pid=12 tid=gv0 INFO: Booting Sidekiq 6.2.1 with redis options {:url=>"redis://myapp-production:REDACTED@redis-myapp-production:6379"}
pid=12 tid=gv0 WARN: Redis::CommandError: ERR wrong number of arguments for 'auth' command
pid=12 tid=gv0 WARN: /app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis/client.rb:137:in `call'
/app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis/client.rb:118:in `block in connect'
/app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis/client.rb:318:in `with_reconnect'
/app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis/client.rb:116:in `connect'
/app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis/client.rb:391:in `ensure_connected'
/app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis/client.rb:243:in `block in process'
/app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis/client.rb:330:in `logging'
/app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis/client.rb:242:in `process'
/app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis/client.rb:136:in `call'
/app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis.rb:2194:in `block in hget'
/app/vendor/bundle/ruby/2.7.0/gems/redis-4.3.0/lib/redis.rb:70:in `block in synchronize'
```

byroot:
What is your redis server version?

lucascaton:
@ byroot `Redis 6.2.4 (00000000/0) 64 bit`

lucascaton:
@byroot `Redis 6.2.4 (00000000/0) 64 bit`

[...]

PR #967, opened 2020-12-13
Support AUTH command for ACL

supercaracal:
Hello,

I think we can support AUTH command for ACL without any breaking changes until RESP2's EOL.
Do you have any other plans?

Related issues:
* resolve #867 
* resolve #956

Documents:
* https://redis.io/topics/acl
* https://redis.io/commands/auth

Affected features:
* Standalone mode
* Sentinel mode
* Cluster mode

New options:
* `username`

Usage:
```ruby
redis = Redis.new(username: 'myname', password: 'mysecret')
```

```ruby
redis = Redis.new(url: 'redis://myname:mysecret@127.0.0.1:6379/0')
```

```ruby
sentinels = [{ host: '127.0.0.1', port: 26380, username: 'myname', password: 'mysecret' },
             { host: '127.0.0.1', port: 26381, username: 'myname', password: 'mysecret' }]

redis = Redis.new(host: 'mymaster', sentinels: sentinels, role: :master, username: 'myname', password: 'mysecret')
```

```ruby
nodes = (7000..7005).map { |port| "redis://myname:mysecret@127.0.0.1:#{port}" }
redis = Redis.new(cluster: nodes)
```

```ruby
redis.auth('myname', 'mysecret')
```

PR #990, opened 2021-06-11
Fix password authentication on Redis 5 and older

byroot:
Fix: https://github.com/redis/redis-rb/issues/989
Ref: https://github.com/redis/redis-rb/pull/967

We should have a test for this, but it requires a different `redis.conf`. I'll have a quick look at adding this, but if it's complicated I might ship a release ASAP.

cc @supercaracal @lucascaton

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -115,7 +115,17 @@ class Redis
       # Don't try to reconnect when the connection is fresh
       with_reconnect(false) do
         establish_connection
-        call [:auth, username, password].compact if username || password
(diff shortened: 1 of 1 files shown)


byroot:
Ok, adding a test for password auth is gonna be tricky. @lucascaton can I ask you to test this branch as a git gem to make sure I haven't overlooked anything?

byroot:
Ok, I did some manual testing against both Redis 5 and 6.2, should be good enough for now, but we should really add coverage for this.

PR #995, opened 2021-06-20
add ZINTER command, added to Redis in 6.2

SkipKayhil:
This adds support for the [ZINTER](https://redis.io/commands/zinter) command, which was added in Redis 6.2.0.

I tried to imitate the API/wording of similar functions, let me know if there is anything I need to change

Ref: #978


byroot:
Thank you!

PR #998, opened 2021-06-30
Handle delete of empty list of keys

haroldsultan:
Handle deletion of an empty list of keys

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -555,6 +555,7 @@ class Redis
   # @param [String, Array<String>] keys
   # @return [Integer] number of keys that were deleted
   def del(*keys)
+    keys.flatten!(1)
     return 0 if keys.empty?
 
     synchronize do |client|
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -555,9 +555,7 @@ class Redis
   # @param [String, Array<String>] keys
   # @return [Integer] number of keys that were deleted
   def del(*keys)
-    if keys.is_a?(Array) && keys.empty?
-      return 0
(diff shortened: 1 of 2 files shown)


haroldsultan:
Please advise as to how to test.  Thanks

lib/redis.rb
```diff
@@ -555,6 +555,10 @@ def migrate(key, options)
   # @param [String, Array<String>] keys
   # @return [Integer] number of keys that were deleted
   def del(*keys)
+    if keys.is_a?(Array) && keys.empty?
+      return 0
+    end
```
byroot, line 560:
```suggestion
    return 0 if keys.empty?
```

`keys` can't be anything else than an array.

byroot:
> Please advise as to how to test. Thanks

https://github.com/redis/redis-rb/blob/6542934f01b9c390ee450bd372209a04bc3a239b/test/commands_on_value_types_test.rb#L10-L24

test/commands_on_value_types_test.rb
```diff
@@ -30,6 +32,8 @@ def test_del_with_array_argument
 
     assert_equal ["bar", "baz", "foo"], r.keys("*").sort
 
+    assert_equal 0, r.del([])
```
byroot, line 35: This isn't handled by your code, you need to add a `keys.flatten!` in `del`.

byroot: Or even `flatten!(1)`

haroldsultan:
Thanks @byroot for the review and feedback!

PR #1018, opened 2021-07-12
add XAUTOCLAIM command, added to Redis in 6.2

alexmreis:
This adds support for the [XAUTOCLAIM](https://redis.io/commands/xautoclaim) command, which was added in Redis 6.2.0.

I tried to imitate the API/wording and hashifying of similar functions, let me know if there is anything I need to change

Ref: #978

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -3280,6 +3280,38 @@ class Redis
     synchronize { |client| client.call(args, &blk) }
   end
 
+  # Transfers ownership of pending stream entries that match the specified criteria.
+  #
+  # @example Claim next pending message stuck > 5 minutes  and mark as retry
+  #   redis.xautoclaim('mystream', 'mygroup', 'consumer1', 3600000, '0-0')
+  # @example Claim 50 next pending messages stuck > 5 minutes  and mark as retry
+  #   redis.xclaim('mystream', 'mygroup', 'consumer1', 3600000, '0-0', count: 50)
+  # @example Claim next pending message stuck > 5 minutes and don't mark as retry
+  #   redis.xclaim('mystream', 'mygroup', 'consumer1', 3600000, '0-0', justid: true)
+  # @example Claim next pending message after this id stuck > 5 minutes  and mark as retry
+  #   redis.xautoclaim('mystream', 'mygroup', 'consumer1', 3600000, '1641321233-0')
+  #
+  # @param key           [String]        the stream key
+  # @param group         [String]        the consumer group name
+  # @param consumer      [String]        the consumer name
+  # @param min_idle_time [Integer]       the number of milliseconds
+  # @param start         [String]        entry id to start scanning from or 0-0 for everything
+  # @param count         [Integer]       number of messages to claim (default 1)
+  # @param justid        [Boolean]       whether to fetch just an array of entry ids or not.
+  #                                      Does not increment retry count when true
+  #
+  # @return [Hash{String => Hash}] the entries successfully claimed
+  # @return [Array<String>]        the entry ids successfully claimed if justid option is `true`
+  def xautoclaim(key, group, consumer, min_idle_time, start, count: nil, justid: false)
+    args = [:xautoclaim, key, group, consumer, min_idle_time, start]
+    if count
+      args << 'COUNT' << count.to_s
+    end
+    args << 'JUSTID' if justid
+    blk = justid ? HashifyStreamAutoclaimJustId : HashifyStreamAutoclaim
+    synchronize { |client| client.call(args, &blk) }
+  end
+
   # Fetches not acknowledging pending entries
   #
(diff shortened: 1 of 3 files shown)


alexmreis:
@rhymes @byroot XAUTOCLAIM implementation is done for #978 . Could you maybe have a look and comment/approve?

byroot:
I got the first notification. I'll get to it when I have a bit of time. Thanks for your patience.

alexmreis:
> I got the first notification. I'll get to it when I have a bit of time. Thanks for your patience.

Sorry for the double notification! Wasn't sure if you were automatically notified. Thanks for considering this PR 🙏

lib/redis.rb
```diff
@@ -3280,6 +3280,38 @@ def xclaim(key, group, consumer, min_idle_time, *ids, **opts)
...
+  # @param start         [String]        entry id to start scanning from or 0-0 for everything
+  # @param opts          [Hash]          several options for `XCLAIM` command
+  #
+  # @option opts [Integer] :count      number of messages to claim (default 1)
+  # @option opts [Boolean] :justid     whether to fetch just an array of entry ids or not.
+  #                                    Does not increment retry count when true
+  #
+  # @return [Hash{String => Hash}] the entries successfully claimed
+  # @return [Array<String>]        the entry ids successfully claimed if justid option is `true`
+  def xautoclaim(key, group, consumer, min_idle_time, start, **opts)
```
byroot, line 3307: We shouldn't accept `**opts`, there's only two possible options, we should define both as regular keyword args. Any extra unexpected kwarg should fail.

lib/redis.rb
```diff
@@ -3280,6 +3280,38 @@ def xclaim(key, group, consumer, min_idle_time, *ids, **opts)
...
+    args = [:xautoclaim, key, group, consumer, min_idle_time, start]
+    args.concat(['COUNT', opts[:count]])           if opts[:count]
+    args << 'JUSTID'                               if opts[:justid]
```
byroot, line 3310: nitpick: please don't align the `if` like this, it doesn't help readability.

byroot (changes requested):
Two small things, otherwise LGTM.

lib/redis.rb
alexmreis: Fixed. Agree completely. I did it to try to remain consistent the same code style used in `xclaim` and the other x commands. 

Maybe we should add an issue to replace all `**opts` hashes in the `X...` stream commands with kwargs? Would break backwards compatibility though.

lib/redis.rb
alexmreis: Same here, again was just keeping it consistent with the rest of the streams-related code. Fixed.

lib/redis.rb
```diff
@@ -3280,6 +3280,36 @@ def xclaim(key, group, consumer, min_idle_time, *ids, **opts)
...
+  # @param start         [String]        entry id to start scanning from or 0-0 for everything
+  # @param count         [Integer]       number of messages to claim (default 1)
+  # @param justid        [Boolean]       whether to fetch just an array of entry ids or not.
+  #                                      Does not increment retry count when true
+  #
+  # @return [Hash{String => Hash}] the entries successfully claimed
+  # @return [Array<String>]        the entry ids successfully claimed if justid option is `true`
+  def xautoclaim(key, group, consumer, min_idle_time, start, count: nil, justid: false)
+    args = [:xautoclaim, key, group, consumer, min_idle_time, start]
+    args.concat(['COUNT', count]) if count
```
byroot, line 3307: nitpick: `args << 'COUNT' << count.to_s`. Saves allocation an array for nothing (not that I expect this too be in a hotspot but still..)

alexmreis: Fixed in latest push

test/lint/streams.rb
```diff
@@ -633,6 +634,72 @@ def test_xclaim_with_invalid_arguments
...
 
+    def test_xautoclaim
+      omit_version(MIN_REDIS_VERSION_XAUTOCLAIM)
+
+      redis.xadd('s1', { f: 'v1' }, id: '0-1')
+      redis.xgroup(:create, 's1', 'g1', '$')
+      redis.xadd('s1', { f: 'v2' }, id: '0-2')
+      redis.xadd('s1', { f: 'v3' }, id: '0-3')
+      redis.xreadgroup('g1', 'c1', 's1', '>')
+      sleep 0.01
```
byroot, line 645: What's the sleep for?

alexmreis: `XPENDING` and `XAUTOCLAIM` both return entries in the consumer group that have been idle for `n` ms. We need them to be idle for a few ms in order to get anything in the test.

byroot: 👍

byroot (approved):

PR #1020, opened 2021-07-26
Fix a bug for cross-slot validation in pipelining

### LINKED ISSUE #1019: `CrossSlotPipeliningError` validation seems incorrect?
From my reading of https://redis.io/topics/cluster-spec#keys-hash-tags it seems that the magic squiggly brackets to force a specific value for hashing should be acceptable anywhere in the key.  However the raise of `CrossSlotPipeliningError` here:

https://github.com/redis/redis-rb/blob/6542934f01b9c390ee450bd372209a04bc3a239b/lib/redis/cluster.rb#L82

Seems to only accept the magic squigglies as a prefix:

```
irb(main):001:0> REDIS.pipelined { ['{zed}one', '{zed}two'].each { |k| REDIS.exists?(k) } }
=> [false, false]
irb(main):002:0> REDIS.pipelined { ['one{zed}', 'two{zed}'].each { |k| REDIS.exists?(k) } }
        1: from (irb):2
Redis::Cluster::CrossSlotPipeliningError (Cluster client couldn't send pipelining to single node. The commands include cross slot keys. ["zed", "zed"])
irb(main):003:0> REDIS.pipelined { ['one{zed}one', 'two{zed}one'].each { |k| REDIS.exists?(k) } }
        2: from (irb):2
        1: from (irb):3:in `rescue in irb_binding'
Redis::Cluster::CrossSlotPipeliningError (Cluster client couldn't send pipelining to single node. The commands include cross slot keys. ["zed", "zed"])
```

This is v4.3.1

supercaracal:
This PR will close #1019.

The bug can be reproduced when we use multiple replicas in a cluster.

### DIFF
--- a/test/cluster_client_pipelining_test.rb
+++ b/test/cluster_client_pipelining_test.rb
@@ -56,4 +56,26 @@ class TestClusterClientPipelining < Minitest::Test
       end
     end
   end
+
+  def test_pipelining_with_multiple_replicas
+    rc = build_another_client(replica: true)
+    rc.instance_variable_get(:@client).instance_variable_get(:@slot).instance_variable_get(:@map).each do |_, v|
+      v[:slaves] << v[:master] if v[:slaves].size < 2 # reproducing multiple replicas
+    end
+
+    rc.pipelined do |r|
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+    end
+
+    rc.close
+  end
 end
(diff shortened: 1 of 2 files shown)


test/cluster_client_pipelining_test.rb
```diff
@@ -56,4 +56,26 @@ def test_pipelining_without_hash_tags
...
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
+      r.get('key1')
```
byroot, line 76: maybe  for concision? 
```suggestion
      10.times { r.get('key1') }
```

PR #1021, opened 2021-07-29
Remove note about writing keys in Redis Cluster replica mode

oleg-vinted:
I don't think `replica: true` prevents writes from working in Cluster mode. This works fine for us in production and the code seems to explicitly direct writes to the master node.


oleg-vinted:
Closing because of inactivity.

Issue #936, opened 2020-08-20
Claiming trimmed messages from a stream leads to an error


flash-gordon:
What happens: 
1. Consumer reads a message then fails.
I think messages that were trimmed should result in `nil`.

flash-gordon:
There's `justid` option in `xclaim` that will avoid hashifying fetched records. Still, the current error is confusing, it'd be better to have `MessageDeletedError` or something like that.

PR #1022, opened 2021-07-29
Improve empty stream responses deserialization

godric:
This is to fix an error that still happens despite #905:

When attempting to read trimmed, pending messages for a consumer group from a stream, we still get
> `NoMethodError: undefined method `each_slice' for nil:NilClass

This is because Redis keeps the entry_id despite the entry itself having been deleted, and will respond with, eg:
```bash
% redis-cli XREADGROUP GROUP test_group test_consumer STREAMS test_key 0
1) 1) "test_key"
   2) 1) 1) "1627563001384-0"
         2) (nil)
```
The branch provides a spec to reproduces the issue and validate the fix. Let me know if you need me to do changes.
And thanks for this great gem 👍 

PS: This is related to the bug mentioned in issues #936, #929. Although I wasn't able to reproduce the issue with XCLAIM, potentially thanks to the previous fix in #905


test/streams_test.rb
```diff
@@ -0,0 +1,18 @@
...
+require_relative "helper"
+
+class TestStreams < Minitest::Test
+  include Helper::Client
+
+  def test_read_a_trimmed_entry
+    r.xgroup(:create, 'k1', 'g1', '0', mkstream: true)
+    entry_id = r.xadd('k1', { value: 'v1' })
+
+    assert_equal r.xreadgroup('g1', 'c1', 'k1', '>'), { 'k1' => [[entry_id, { 'value' => 'v1' }]] }
```
byroot, line 12: The order of arguments is `assert_equal expectation, actual_value`.

byroot:
The CI failures look legit.

test/streams_test.rb
godric: I fixed it.
FYI, I based my suite on `ssl_test` that also has this problem 😅
https://github.com/redis/redis-rb/blob/6542934f01b9c390ee450bd372209a04bc3a239b/test/ssl_test.rb#L19

godric:
Sorry, I utterly failed my attempt to rebase on `redis-rb/master`
I also realised one of my assertion was wrong, and fixed it.

byroot:
```
Redis::CommandError: ERR unknown command `xgroup`, with args beginning with: `create`, `k1`, `g1`, `0`, `MKSTREAM`,
```

You need to handle older redises in the test suite.

test/streams_test.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require_relative "helper"
+
+class TestStreams < Minitest::Test
+  include Helper::Client
+
+  def test_read_a_trimmed_entry
```
byroot, line 8: This should go in https://github.com/redis/redis-rb/blob/master/test/lint/streams.rb (that will give you the automatic skip if redis is too old)

godric: I renamed the test and moved it with other `xreadgroup` tests ✅

byroot:
Thank you.

Issue #935, opened 2020-08-18
Value retrieval slowdown for payloads over 5 MB with SSL transport


Joas1988:
To replicate connect to redis using SSL and test set and get with payload of more than 5 MB.

If you attach an strace to a process that is trying to do a large redis read, you can see that it is in a loop of

- allocate a buffer of the length of the [bulk string](https://github.com/redis/redis-rb/blob/d75e074c9997f8d099b6a0f65f0cf517274943f2/lib/redis/connection/ruby.rb#L38). The allocation in this case is an mmap - as malloc normally uses mmap when trying to alloc giant memory sizes.

- syscall to read from the socket - usually around 1400 bytes or so

- repeat until we are done

It boils down to that Ruby OpenSSL is reading 1400 bytes and without using buffer it [allocates buffer internally](https://github.com/ruby/openssl/blob/41587f69e17b9f0983c1f2a37b8661599119fc0e/ext/openssl/ossl_ssl.c#L1834) for length provided as argument to [def read_nonblock](https://github.com/ruby/openssl/blob/master/lib/openssl/buffering.rb#L195.) In my company we solved that issue by this patch.

```ruby 
Redis::Connection::SocketMixin.module_eval do
    def read(nbytes)
      result = @buffer.slice!(0, nbytes)
      # always use this pre-allocated buffer when calling into openssl
      buffer = String.new(capacity: nbytes, encoding: Encoding::ASCII_8BIT)
      result << _read_from_socket(nbytes - result.bytesize, buffer) while result.bytesize < nbytes
      result
    end

    def _read_from_socket(nbytes, buffer = nil)
      # addition here to use the pre allocated buffer
      read_nonblock(nbytes, buffer)
    rescue IO::WaitReadable
      if IO.select([self], nil, nil, @timeout)
        retry
      else
        raise Redis::TimeoutError
      end
    rescue IO::WaitWritable
      if IO.select(nil, [self], nil, @timeout)
        retry
      else
        raise Redis::TimeoutError
      end
    end
  end
```

rbroemeling:
@Joas1988 Thanks for discovering this and posting this patch -- we encountered exactly this problem in our production environment with larger result sizes (i.e., results of a few MB or higher). It was very noticeable, and your patch mitigated the impact greatly.

Have you heard anything back from the maintainers of `redis-rb` on this topic? This seems like a serious problem in the library, so I'm a little curious at the lack of activity or interest on this change? Any ideas?

[...]

PR #1023, opened 2021-08-10
Preallocate a buffer for reading from a socket

kevinhughes27:
closes https://github.com/redis/redis-rb/issues/935 cc  @Joas1988

We encountered the same issue described in https://github.com/redis/redis-rb/issues/935 - fetching large (2 - 5 megabyte) values from Redis with SSL performed much slower than expected. We applied the patch from the issue to our production systems and saw a dramatic increase in performance.

As described in the issue the root cause of the performance degradation is:

> It boils down to that Ruby OpenSSL is reading 1400 bytes and without using buffer it [allocates buffer internally](https://github.com/ruby/openssl/blob/41587f69e17b9f0983c1f2a37b8661599119fc0e/ext/openssl/ossl_ssl.c#L1834) for length provided as argument to def [read_nonblock](https://github.com/ruby/openssl/blob/master/lib/openssl/buffering.rb#L214)

I further confirmed this by logging `nbytes` inside of `Redis::Connection::SocketMixin#_read_from_socket`. The interesting part of the logs looked like this:

```
13635520
13619136
13602752
13586368
```

Where `nbytes` starts around the full size of the payload and is decreasing by 16384 until it is finished reading. As described in the issue the call to `read_nonblock` allocates a new string based on the length provided resulting in many large allocations.

We experienced one other oddity during this investigation: the issue was not reproducible from a plain Ruby script, only from Rails. To dig into this I was able to reproduce the performance difference using only string allocations:

```ruby
def alloc_loop
  initial_size = 13650000
  chunk_size = 16384

  size = initial_size
  while size > 0 do
    size -= chunk_size
    buffer = String.new(capacity: size, encoding: Encoding::ASCII_8BIT)
  end
end

def timeit(name, &block)
  start = Time.now
  block.call
  puts "#{name} took: #{Time.now - start}"
end

timeit "alloc_loop" do
  alloc_loop
end
```

In plain Ruby I got `alloc_loop took: 0.054234` while in Rails I was getting `alloc_loop took: 1.622729` which very closely matched the difference we were seeing when comparing the Redis reads.

After ruling out patches and gems I compared to a new Rails app and I got somewhere in between the two. This led me to a new theory that the issue was more prominent in larger applications. I was able to confirm this by running this function before the string allocation test.

```ruby
def bloat
  100000.times do |i|
    eval "class BloatClass#{i}; end"
  end
end
```

What appears to be happening is that larger programs have a higher memory baseline causing the garbage collector to be more active. The output of `GC.stat` shows that it is a lot more active during the string allocations when the program memory space is higher.

I figured this was worth sharing in case someone wants to confirm the change and does so with a plain Ruby script.

This PR applies the patch from the issue. We’ve been running it on production for around a week and haven’t had any issues. There didn’t seem to be an obvious place to test this but the SocketMixin would be implicitly covered by the existing tests. Let me know if there is anything else you’d like to see in order to get this merged!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -106,6 +106,9 @@ Lint/EndAlignment:
 Layout/ElseAlignment:
   Enabled: false
 
+Layout/RescueEnsureAlignment:
+  Enabled: false
+
 Naming/HeredocDelimiterNaming:
   Enabled: false
 
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -32,12 +32,30 @@ class Redis
         @write_timeout = (timeout if timeout && timeout > 0)
       end
 
-      def read(nbytes)
-        result = @buffer.slice!(0, nbytes)
+      string_capacity_support = begin
+        String.new(capacity: 0)
+        true # Ruby 2.4+
+      rescue ArgumentError
+        false # Ruby 2.3
+      end
 
-        result << _read_from_socket(nbytes - result.bytesize) while result.bytesize < nbytes
+      if string_capacity_support
+        def read(nbytes)
+          result = @buffer.slice!(0, nbytes)
 
-        result
+          buffer = String.new(capacity: nbytes, encoding: Encoding::ASCII_8BIT)
+          result << _read_from_socket(nbytes - result.bytesize, buffer) while result.bytesize < nbytes
+
+          result
+        end
+      else
+        def read(nbytes)
+          result = @buffer.slice!(0, nbytes)
+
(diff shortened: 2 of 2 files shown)


Joas1988:
Thx for jumping on this. Additional research that you did is in line with what we saw 1 year ago. That said, we should look at other gems that are using OpenSSL, just to see if or how they are dealing with this problem.

byroot:
Thanks for the PR. Looks like `capacity` was introduced in 2.4. It should be easy to work around, but it might also be time to drop 2.3 support 🤔

byroot:
I merged a slightly modified version for Ruby 2.3 support: fe18b814628a95ac0addc2064caf15b54ca7c004.

Thanks!

PR #1024, opened 2021-08-18
[Redis 2.6] Add support to getdel and getex

heyvito:
This adds support to `getdel` and `getex` commands.

Reviewers: I'm not sure whether documentation I placed is enough. Please let me know in case you folks need changes. This also applies to tests.

 References #978


byroot:
Thank you.

PR #1028, opened 2021-09-13
[Redis 6.2] Add EXAT/PXAT options to SET

magni-:
First PR here 👋🏼 For tests, I took inspiration from https://github.com/redis/redis-rb/pull/1024 as well as the existing tests for `SET` using `EX` / `PX`.
Documentation is copy-pasted from https://redis.io/commands/set#options


byroot (approved):
Thanks. I'll merge on green CI.

PR #1031, opened 2021-10-07
[Redis 6.2] Add SMISMEMBER command

fatkodima:
This adds support for the [SMISMEMBER](https://redis.io/commands/smismember) command, which was added in Redis 6.2.

Reference: #978


byroot:
Thank you!

PR #1040, opened 2021-10-15
[Redis 6.2] Add HRANDFIELD command

fatkodima:
Support for [HRANDFIELD](https://redis.io/commands/hrandfield) from redis 6.2.

Reference: #978

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -1907,7 +1907,7 @@ class Redis
   # @example Get multiple random members
   #   redis.zrandmember("zset", 2)
   #     # => ["a", "b"]
-  # @example Gem multiple random members with scores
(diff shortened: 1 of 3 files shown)


lib/redis.rb
```diff
@@ -2457,6 +2457,48 @@ def mapped_hmget(key, *fields)
...
+    args << "WITHVALUES" if with_values
+
+    synchronize do |client|
+      client.call(args) do |reply|
+        if with_values
+          Pairify.call(reply)
+        else
+          reply
+        end
+      end
```
byroot, line 2498:
```suggestion
      parser = Pairify if with_values
      client.call(args, &parser)
```

byroot (changes requested):

PR #1042, opened 2021-10-19
[Redis 6.2] Add ZUNION command

fatkodima:
Adds [ZUNION](https://redis.io/commands/zunion) command from redis 6.2.
I decided to extract a private helper method for both `zinter` and `zunion` to remove duplication.

Reference: #978


lib/redis.rb
```diff
@@ -3826,6 +3833,32 @@ def _normalize_move_wheres(where_source, where_destination)
 
     [where_source, where_destination]
   end
+
+  def _zsets_operation(cmd, args)
+    options = args.pop if args.last.is_a?(Hash)
```
byroot, line 3838: I think I'd prefer if you continued to use keyword arguments. To properly forward arguments from the other methods you can use `ruby2_keywords`.

PR #1048, opened 2021-11-04
[Redis 6.2] Add COPY command

fatkodima:
Support for [COPY](https://redis.io/commands/copy) command from redis 6.2.

Reference: #978

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -656,6 +656,41 @@ class Redis
     end
   end
 
+  # Copy a value from one key to another.
+  #
+  # @example Copy a value to another key
+  #   redis.set "foo", "value"
+  #     # => "OK"
+  #   redis.copy "foo", "bar"
+  #     # => true
+  #   redis.get "bar"
+  #     # => "value"
+  #
+  # @example Copy a value to a key in another database
+  #   redis.set "foo", "value"
+  #     # => "OK"
+  #   redis.copy "foo", "bar", db: 2
+  #     # => true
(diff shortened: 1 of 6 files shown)


lib/redis.rb
```diff
@@ -656,6 +656,41 @@ def move(key, db)
...
+  #
+  # @param [String] source
+  # @param [String] destination
+  # @param [Integer] db
+  # @param [Boolean] replace removes the `destination` key before copying value to it
+  # @return [Boolean] whether the key was copied or not
+  def copy(source, destination, db: nil, replace: false)
+    comamnd = [:copy, source, destination]
+    comamnd << "DB" << db if db
+    comamnd << "REPLACE" if replace
```
byroot, line 687:
```suggestion
    command = [:copy, source, destination]
    command << "DB" << db if db
    command << "REPLACE" if replace
```

lib/redis.rb
```diff
@@ -656,6 +656,41 @@ def move(key, db)
...
+
+    synchronize do |client|
+      client.call(comamnd, &Boolify)
```
byroot, line 690:
```suggestion
      client.call(command, &Boolify)
```

byroot (changes requested):
There's a small typo, other than that LGTM.

fatkodima:
Ah, sorry, didn't notice. Thank you 👍

byroot:
No worries. As always thanks for the patch!

Issue #1045, opened 2021-10-22
Cluster mode #scan is not working properly


estepnv:
redis (4.5.1)

I assume that KEYS * resulting set must be equal to the sum of SCAN X MATCH * calls

Currently when connected to in Redis cluster mode `Redis#scan` and `Redis#scan_each` does not conform to the rule above.

According to https://github.com/redis/redis-rb/blob/master/lib/redis/cluster.rb#L130
`Redis::Cluster#send_command` will execute `#scan_each` on a random node, however, IMO it has to iterate over all nodes and collect data from them.

If anybody helps me with local setup (docker-compose file) I'll try to implement this with tests

supercaracal:
Example of Docker Compose:

```yaml
version: '3.9'
services:
  node1: &node
    image: "redis:6.2.6"
    command: >
      redis-server --maxmemory            64mb
                   --maxmemory-policy     allkeys-lru
                   --appendonly           yes
                   --cluster-enabled      yes
                   --cluster-config-file  nodes.conf
                   --cluster-node-timeout 5000
    healthcheck:
      test: ["CMD", "redis-cli", "ping"]
      interval: '3s'
      timeout: '1s'
      retries: 10
    ports:
      - "7000:6379"
  node2:
    <<: *node
    ports:
      - "7001:6379"
  node3:
    <<: *node
    ports:
      - "7002:6379"
  node4:
    <<: *node
    ports:
      - "7003:6379"
  node5:
    <<: *node
    ports:
      - "7004:6379"
  node6:
    <<: *node
    ports:
      - "7005:6379"
  node-redis-cluster:
    image: "redis:6.2.6"
    command: >
      bash -c "apt-get update
      && apt-get install --no-install-recommends --no-install-suggests -y dnsutils
      && yes yes | redis-cli --cluster create
      $$(dig node1 +short):6379
      $$(dig node2 +short):6379
      $$(dig node3 +short):6379
      $$(dig node4 +short):6379
      $$(dig node5 +short):6379
      $$(dig node6 +short):6379
      --cluster-replicas 1"
    depends_on:
      node1:
        condition: service_healthy
      node2:
        condition: service_healthy
      node3:
        condition: service_healthy
      node4:
        condition: service_healthy
      node5:
        condition: service_healthy
      node6:
        condition: service_healthy
```

```
$ docker-compose ps
           Name                         Command                  State                        Ports
---------------------------------------------------------------------------------------------------------------------
[...]

PR #1049, opened 2021-11-06
Add cluster scan implementation

estepnv:
Fixes redis/redis-rb#1045

### DIFF
--- a/lib/redis/cluster.rb
+++ b/lib/redis/cluster.rb
@@ -137,6 +137,7 @@ class Redis
       when 'wait'     then @node.call_master(command, &block).reduce(:+)
       when 'keys'     then @node.call_slave(command, &block).flatten.sort
       when 'dbsize'   then @node.call_slave(command, &block).reduce(:+)
+      when 'scan'     then _scan(command, &block)
       when 'lastsave' then @node.call_all(command, &block).sort
       when 'role'     then @node.call_all(command, &block)
       when 'config'   then send_config_command(command, &block)
@@ -236,6 +237,29 @@ class Redis
     rescue CannotConnectError
       update_cluster_info!
       raise
+    end
+
+    def _scan(command, &block)
+      input_cursor = Integer(command[1])
+
+      client_index = input_cursor % 256
+      raw_cursor = input_cursor >> 8
+
+      clients = @node.scale_reading_clients
+
+      client = clients[client_index]
+      return ['0', []] unless client
+
+      command[1] = raw_cursor.to_s
+
+      result_cursor, result_keys = client.call(command, &block)
+      result_cursor = Integer(result_cursor)
+
+      if result_cursor == 0
+        client_index += 1
+      end
+
+      [((result_cursor << 8) + client_index).to_s, result_keys]
     end
 
     def assign_redirection_node(err_msg)
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ Gemfile.lock
 /test/test.conf
 appendonly.aof
 temp-rewriteaof-*.aof
+.history
+
--- a/lib/redis/cluster/node.rb
+++ b/lib/redis/cluster/node.rb
@@ -58,6 +58,18 @@ class Redis
         try_map { |_, client| client.process(commands, &block) }.values
       end
 
+      def scale_reading_clients
+        reading_clients = []
+
+        @clients.each do |node_key, client|
+          next unless replica_disabled? ? master?(node_key) : slave?(node_key)
+
+          reading_clients << client
+        end
+
+        reading_clients
+      end
+
       private
 
       def replica_disabled?
--- a/test/cluster_commands_on_keys_test.rb
+++ b/test/cluster_commands_on_keys_test.rb
@@ -136,4 +136,22 @@ class TestClusterCommandsOnKeys < Minitest::Test
 
     assert_equal 2, all_keys.uniq.size
   end
+
+  def test_scan_each
+    require 'securerandom'
+
+    1000.times do |n|
+      redis.set("test-#{::SecureRandom.uuid}", n)
+    end
+
+    1000.times do |n|
+      redis.set("random-#{::SecureRandom.uuid}", n)
+    end
+
+    keys_result = redis.keys('test-*')
+    scan_result = redis.scan_each(match: 'test-*').to_a
+    assert_equal(keys_result.size, 1000)
+    assert_equal(scan_result.size, 1000)
+    assert_equal(scan_result.sort, keys_result.sort)
+  end
 end


estepnv:
please suggest what additional tests should I add

lib/redis/cluster.rb
```diff
@@ -25,6 +26,7 @@ def initialize(options = {})
       @option = Option.new(options)
       @node, @slot = fetch_cluster_info!(@option)
       @command = fetch_command_details(@node)
+      @scan_state = Hash.new { |h,k| h[k] = Hash.new { |h,k| h[k] = {} } }
```
supercaracal, line 29: [nitpick] Rubocop says:

```
Offenses:

lib/redis/cluster.rb:29:34: C: Layout/SpaceAfterComma: Space missing after comma.
      @scan_state = Hash.new { |h,k| h[k] = Hash.new { |h,k| h[k] = {} } }
                                 ^
lib/redis/cluster.rb:29:57: W: Lint/ShadowingOuterLocalVariable: Shadowing outer local variable - h.
      @scan_state = Hash.new { |h,k| h[k] = Hash.new { |h,k| h[k] = {} } }
                                                        ^
lib/redis/cluster.rb:29:58: C: Layout/SpaceAfterComma: Space missing after comma.
      @scan_state = Hash.new { |h,k| h[k] = Hash.new { |h,k| h[k] = {} } }
                                                         ^
lib/redis/cluster.rb:29:59: W: Lint/ShadowingOuterLocalVariable: Shadowing outer local variable - k.
      @scan_state = Hash.new { |h,k| h[k] = Hash.new { |h,k| h[k] = {} } }
                                                          ^
```

estepnv: fixed

lib/redis/cluster.rb
```diff
@@ -238,6 +241,46 @@ def try_send(node, method_name, *args, retry_count: 3, &block)
       raise
     end
 
+    def _scan(command, &block)
```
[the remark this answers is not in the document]
supercaracal: Does it leak when client quits scanning half-way?

```
irb(main):011:0> r = Redis.new cluster: %w[redis://127.0.0.1:7000]
=> #<Redis client v4.5.1 for redis://127.0.0.1:7000/0 redis://127.0.0.1:7001/0 redis://127.0.0.1:7002/0>

irb(main):012:0> r.keys '*'
=> ["key1", "key2", "{key}1", "{key}2"]

irb(main):013:0> r.scan 0, count: 1
=> ["x-2f620d-61490425", ["key2", "key1", "{key}2"]]

irb(main):014:0> r.instance_variable_get(:@client).instance_variable_get(:@scan_state)
=> {"2f620d"=>{"127.0.0.1:7000"=>{"x-2f620d-61490425"=>"0"}, "127.0.0.1:7001"=>{"x-2f620d-61490425"=>"3"}, "127.0.0.1:7002"=>{"x-2f620d-61490425"=>"1"}}}

irb(main):015:0> r.scan 'x-2f620d-61490425'
=> ["x-2f620d-55435333", ["{key}1"]]

irb(main):016:0> r.instance_variable_get(:@client).instance_variable_get(:@scan_state)
=> {"2f620d"=>{"127.0.0.1:7000"=>{}, "127.0.0.1:7001"=>{"x-2f620d-55435333"=>"0"}, "127.0.0.1:7002"=>{"x-2f620d-55435333"=>"0"}}}

irb(main):017:0> r.scan 'x-2f620d-55435333'
=> ["0", []]

irb(main):018:0> r.instance_variable_get(:@client).instance_variable_get(:@scan_state)
=> {}
```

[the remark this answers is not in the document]
supercaracal: ref: https://redis.io/commands/scan#terminating-iterations-in-the-middle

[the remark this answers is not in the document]
estepnv: or use a dumb solution like this
```ruby
# typed: true
# frozen_string_literal: true

class LruCache
  def initialize(max = 10)
    @cache = {}
    @cache_keys = []
    @size = 0
    @max_size = max
  end

  def fetch(key)
    val = get(key)

    if val.nil?
      if block_given?
        val = yield
        set(key, val)
      end
    end

    sweep

    val
  end

  def sweep
    if @size > @max_size
      @cache.delete(@cache_keys.pop)
      @size -= 1
    end
  end

  def set(key, val)
    @cache[key] = val
    @cache_keys.prepend(@cache_keys.delete(key) || key)
    @size += 1
  end

  def up(key)
    if @cache_keys[0] != key
      @cache_keys.prepend(@cache_keys.delete(key))
    end
  end

  def get(key)
    val = @cache[key]

    up(key) if !val.nil?

    val
  end

  def del(key)
    @cache.delete(key)
    @cache_keys.delete(key)
    @size -= 1
  end
end

```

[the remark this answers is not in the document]
estepnv: or we could prohibit `#scan` usage and raise error. and force users to always use the `#Xscan_each` , to make sure and cleans the state object

supercaracal:
https://github.com/redis/redis/issues/1962#issuecomment-568483242

Redis server maintainer said:

> I don't think that `--scan` and `--cluster` modes play nicely together.

I don't know that how much the feature is needed.
But current redis gem's behavior is a bug indeed. Randomly assign is nonsense.

https://github.com/RedisLabs/redis-cluster-proxy#commands-that-act-differently-from-standard-redis-commands-or-that-have-special-behavior

> SCAN: performs the scan on all the master nodes of the cluster.
> The cursor contained in the reply will have a special four-digits suffix indicating the index of the node that has to be scanned. 
> Note: sometimes the cursor could be something like "00001", so you mustn't convert it to an integer when your client has to use it to perform the next scan.

The above proxy feature supports `SCAN` command by similar interface.
But it seems that the implementation is more simply.

https://github.com/RedisLabs/redis-cluster-proxy/blob/ac83840d1108901a475e24232b7a0986e97e65dd/src/proxy.c#L812-L819

> Scan all master nodes in the cluster.
> Node order is alphabeticallty and it's taken from `clusterGetMasterNames`.
> The index of the node is taken from the cursor last 4 digits.
> If the index is beyond the list of all the master the choosen node will be the last one.
> If the cursor is '0', the choosen node will be the first one.
> This special cursor format will be handled by `handleScanReply`, that will append the index of the node to scan to the cursor replied from the cluster.
> The real cursor sent to the cluster won't actually contain the index suffix.

https://redis.io/commands/scan
Also, It seems that redis server has just a state as a cursor.

Is it better if we implement it by stateless logic?

lib/redis/cluster.rb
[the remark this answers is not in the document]
supercaracal: I'd say that the latter solution might be better.
https://github.com/redis/redis-rb/blob/0c43a516bd90127850031ec970e11f343f63fbaf/lib/redis.rb#L3023-L3032

We can simply iterate per nodes without any state.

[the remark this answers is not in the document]
estepnv: by latter you meant stateless, right? as far as i understand, the idea is to try to preserve the state within a returned cursor and we iterate nodes one by one. I'll see what I can do.

supercaracal:
I am just a contributor. It is eventually needed to review by maintainer.

lib/redis/cluster.rb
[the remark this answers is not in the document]
supercaracal: > by latter you meant stateless, right?

Yes, it is.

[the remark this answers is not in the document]
supercaracal: I am sorry for my bad English. My "stateless" means cursor only. It seems that maintaining any additional state might be a bit complex.

supercaracal:
### approach 1
* Client raises an error when calling public `#scan` method in cluster mode.
* Only public `#scan_each` method is legitimate when using cluster mode.
* It needs conditional branching in public `#scan_each` method.

### approach 2
* Client knows sorted node keys such that `host:port`.
  * ex) `127.0.0.1:7000, 127.0.0.1:7001, ...`
* At first, we can call public `#scan` method with `0`.
* Client returns next cursor and scanned keys. The cursor includes current node key index.
  * ex) `17-0001, 0-0002, 42-0002, 0-0003, ...`
* Client can iterate every nodes using only cursor value.

lib/redis/cluster.rb
```diff
@@ -238,6 +240,36 @@ def try_send(node, method_name, *args, retry_count: 3, &block)
...
```
estepnv, line 243: @supercaracal pls take a look. 
so it's pretty simple
I use last 8 bits to encode the client index the rest is the raw key
I iterate keys on a node until I receive `"0"` cursor, after that I increment the index
if current client is nil it's the return condition

estepnv: it's the second approach you proposed but without strings

lib/redis/cluster.rb
```diff
@@ -238,6 +239,36 @@ def try_send(node, method_name, *args, retry_count: 3, &block)
...
+      input_cursor = command[1].to_i
+
+      client_index = input_cursor % 2**8
+      raw_cursor = input_cursor >> 8
+
+      clients = @node
```
supercaracal, line 248: It might be readable if this step is added as a public method to `Redis::Cluster::Node` class.

```ruby
class Redis
  class Cluster
    class Node
      def scale_reading_clients
        @clients.select { |k, v| replica_disabled? ? master?(k) : slave?(k) }.values
      end
    end
  end
end
```

lib/redis/cluster.rb
supercaracal: Thank you for your improvement.

supercaracal (approved):
:+1:

lib/redis/cluster.rb
```diff
@@ -238,6 +239,29 @@ def try_send(node, method_name, *args, retry_count: 3, &block)
       raise
     end
 
+    def _scan(command, &block)
+      input_cursor = command[1].to_i
```
byroot, line 243: Maybe use `Integer` here? So that it raises if we get something that's not an integer.

lib/redis/cluster.rb
```diff
@@ -238,6 +239,29 @@ def try_send(node, method_name, *args, retry_count: 3, &block)
...
+
+      clients = @node.scale_reading_clients
+
+      client = clients[client_index]
+      return ['0', []] unless client
+
+      command[1] = raw_cursor.to_s
+
+      result_cursor, result_keys = client.call(command, &block)
+      result_cursor = result_cursor.to_i
```
byroot, line 256: Same here. `Integer` would make sense.

lib/redis/cluster.rb
```diff
@@ -238,6 +239,29 @@ def try_send(node, method_name, *args, retry_count: 3, &block)
       raise
     end
 
+    def _scan(command, &block)
+      input_cursor = command[1].to_i
+
+      client_index = input_cursor % 2**8
```
byroot, line 245: Probably not much, but we could directly use `256` to not repeatedly recompute that exponent.

byroot (commented):
Please squash your commits together and I'll merge.

lib/redis/cluster.rb
estepnv: @byroot updated

lib/redis/cluster.rb
```diff
@@ -137,6 +137,7 @@ def send_command(command, &block)
       when 'wait'     then @node.call_master(command, &block).reduce(:+)
       when 'keys'     then @node.call_slave(command, &block).flatten.sort
       when 'dbsize'   then @node.call_slave(command, &block).reduce(:+)
+      when 'scan'     then _scan(command, &block)
```
estepnv, line 140: @byroot what about zscan/hscan/sscan, will it work? may I should add more tests? I'm not very familiar with internals.

byroot: More tests wouldn't hurt.

PR #1055, opened 2022-01-06
Fix zpopmax zpopmin in pipelines

illusori:
ZPOPMAX/ZPOPMIN cause an exception when used within a pipeline:

```
NoMethodError: undefined method `first' for <Redis::Future [:zpopmax, "sortedset"]>:Redis::Future
```

This is caused by attempting to modify the return value directly rather than within a transform block.

PR includes two commits:
 * first is an addition to the test suite to demonstrate the bug.
 * second is a fix for the bug, that passes the test suite.

### DIFF
--- a/test/pipelining_commands_test.rb
+++ b/test/pipelining_commands_test.rb
@@ -199,6 +199,17 @@ class TestPipeliningCommands < Minitest::Test
     assert_equal result.first, { "field" => "value" }
   end
 
+  def test_zpopmax_in_a_pipeline_produces_future
+    r.zadd("sortedset", 1.0, "value")
+    future = nil
+    result = r.pipelined do
+      future = r.zpopmax("sortedset")
+    end
+
+    assert_equal [["value", 1.0]], result
+    assert_equal ["value", 1.0], future.value
+  end
+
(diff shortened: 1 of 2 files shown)


test/pipelining_commands_test.rb
```diff
@@ -199,6 +199,17 @@ def test_hgetall_in_a_pipeline_returns_hash
...
 
+  def test_zpopmax_in_a_pipeline_produces_future
+    r.zadd("sortedset", 1.0, "value")
+    future = nil
+    result = r.pipelined do
+      future = r.zpopmax("sortedset")
+    end
+
+    assert_equal [["value", 1.0]], result
+    assert_equal ["value", 1.0], future.value
```
byroot, line 210: This test needs to be wrapped in `target_version('5.0.0') do`, so it's skipped on older redis versions.

byroot (commented):
Thanks, just one minor issue and it should be good to go. It's best to squash your commit together so that when we git blame we see both the change and the associated test.

byroot:
Merged as bc7b0a2b105df1e5700b6a628a66f39179e2d577

PR #1056, opened 2022-01-08
Add Ruby 3.1 to CI

petergoldstein:
In addition to adding Ruby 3.1 to the CI matrix, this PR does the following:

- Upgrades the Rubocop constraint in the Gemfile to '~> 1', '< 1.12', allowing compatibility with both Ruby 3.1 and Ruby 2.4
- Disables those Cops which generate warnings as a result of this upgrade in the `.rubocop.yml` file.


byroot:
> * Disables those Cops which generate warnings as a result of this upgrade in the `.rubocop.yml` file.

I totally understand that you'd rather not fix them yourself, but I'd prefer if they were in a separate `.rubocop_todo.yml` file. Because I'll have to fix them at some point.

petergoldstein:
Actually most of them are easy to fix, I just didn't want to expand the scope of the PR too much.  How about I fix the easy ones and leave anything I don't fix in a rubocop_todo?

byroot:
Whatever is most convenient for you! I don't mind the "scope" being extended a bit in this kind of newer ruby PRs. As long as it doesn't add any new logic, feel free to make the changes you need.

petergoldstein:
@byroot Let me know if that looks better to you.  I fixed many (most?) of the lints, wound up reverting and disabling a false positive, and put the rest in a .rubocop_todo.yml file.

If you want any additional changes just let me know.  Thanks.

lib/redis.rb
```diff
@@ -3857,7 +3857,9 @@ def method_missing(command, *args) # rubocop:disable Style/MissingRespondToMissi
     when nil
       {}
     else
+      # rubocop:disable Style/HashTransformValues
```
byroot, line 3860: Hum, that's a weird one, I'd have to dig, but I suppose `reply` is an Array here. Might be better to just disable that cop entirely if it has false positives like that.

lib/redis/client.rb
```diff
@@ -120,9 +118,10 @@ def connect
             begin
               call [:auth, username, password]
             rescue CommandError => err # Likely on Redis < 6
-              if err.message.match?(/ERR wrong number of arguments for \'auth\' command/)
+              case err.message
+              when /ERR wrong number of arguments for 'auth' command/
```
byroot, line 122: It doesn't really matter much in this code path, but the `match?` was deliberate to avoid allocating a useless `MatchData`.

byroot (approved):
Just one tiny thing (the cop to disable globally rather than inline).

Other than that looks great o me Thank you!

petergoldstein:
Switched it to a global disable for that Cop.  Let me know if anything else needs to change.

byroot:
All good. Thanks!

PR #1057, opened 2022-01-16
Expose Redis#queue_key

machty:
Prior to this commit, the Redis class maintains a queue
mapping Thread.current.object_id to the current queue of
not-yet-committed commands for the particular thread
accessing the Redis client. This works for the common
case where a Redis client is shared amongst multiple
threads, but with Ruby 3's Fiber Scheduler and recent
strides in the Async gem, we'll want a way to choose a
different concurrency primitive (such as a Fiber-backed
Async task) to map queues to. Previously we exposed
`def synchronize` so that the concurrency scheme for
checking out the Redis client could be overridden in
subclasses, but internally the implicit queue key was
hardwired to Thread.current.object_id. This commit takes
a similar approach be exposing `def cache_key` so that it
too can be overriden by Redis subclasses.

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -68,8 +68,22 @@ class Redis
     super() # Monitor#initialize
   end
 
+  # Acquire the mutex wrapping this client, and yield the locked client to the provided block.
+  # This method is used internally to wrap critical sections of thread-unsafe (and fiber-unsafe)
+  # code, and can be overriden in subclasses to provide a custom concurrency model, such as
+  # a connection pool.
   def synchronize
     mon_synchronize { yield(@client) }
+  end
+
+  # Returns the key used to determine the queue of commands that have been
+  # queued up by the various threads using this Redis client. By default, this
+  # returns the object_id of Thread.current, which ensures that two (or more)
+  # threads enqueuing commands on the same Redis client each maintain their own
+  # queues of commands rather than clobbering each other's.
+  # This method is guaranteed to be called within a `synchronize` block.
+  def queue_key
+    Thread.current.object_id
   end
 
   # Run code with the client reconnecting
@@ -117,7 +131,7 @@ class Redis
   #
   def queue(*command)
     synchronize do
-      @queue[Thread.current.object_id] << command
+      @queue[queue_key] << command
     end
   end
 
@@ -129,13 +143,13 @@ class Redis
     synchronize do |client|
       begin
         pipeline = Pipeline.new(client)
-        @queue[Thread.current.object_id].each do |command|
+        @queue[queue_key].each do |command|
           pipeline.call(command)
         end
 
         client.call_pipelined(pipeline)
       ensure
-        @queue.delete(Thread.current.object_id)
+        @queue.delete(queue_key)
       end
     end
   end


byroot:
My plan was to just entirely get rid of that `@queue` thing, I think it's quite a misfeature. `pipeline / multi` should give you an object on which to queue your commands, instead of mutating the client.

It's just gonna take a while to deprecate.

machty:
I think that makes a lot of sense. I also felt the queue mutation was weird but wanted to pick the most conservative route iterating towards Fiber-safety.

machty:
@byroot what's the plan for these block-form methods as far as holding lock-term locks / `synchronize` blocks goes?

For instance:

```rb
```

machty:
@byroot `def pipelined` already [yields self](https://github.com/machty/redis-rb/blob/1912288e896a7364ec235ec6abcf3081dc73518e/lib/redis.rb#L2893) and there appear to be [some tests](https://github.com/machty/redis-rb/blob/1912288e896a7364ec235ec6abcf3081dc73518e/test/pipelining_commands_test.rb#L223) that test this behavior. How can a block that yields an object (perhaps a Pipeline obj?) rather than a Redis instance be introduced in a backwards compatible manner? I wouldn't mind submitting the PR with some guidance here.

byroot:
> How can a block that yields an object (perhaps a Pipeline obj?) rather than a Redis instance be introduced in a backwards compatible manner?

Like you said, it shouldn't break backward compat.

Short term we should yield a `Pipeline` object that respond to all the redis command methods and queue the methods, and at the end of the block we flush it.

The most important is to throw a deprecation warning if the original client is called from inside `pipeline/multi` or if the `Pipeline` object didn't queue anything.

machty:
@byroot the Redis class uses a dynamic [method_missing](https://github.com/machty/redis-rb/blob/1912288e896a7364ec235ec6abcf3081dc73518e/lib/redis.rb#L3792) to forward commands to @client; is there some definitive list of known redis commands we could consult to issue proper deprecation warnings?

byroot:
> is there some definitive list of known redis commands we could consult to issue proper deprecation warnings?

No need. You can use proxy objects.

`pipelined / multi` decorate the client: https://github.com/machty/redis-rb/blob/1912288e896a7364ec235ec6abcf3081dc73518e/lib/redis.rb#L2844-L2854

You can simply decorate with a `DeprecatedPipeline` which raise deprecation warnings.

Alternatively, you can raise a warning if the `Pipeline` object you yielded is still empty when the block ends (which likely means the caller didn't use the yielded object).

Yet another method would be to check the block arity.

It's also not a strict or, some combinaison of the above might make sense.

machty:
I have unfortunately run through my budget on being able to work on this; got a little sucked into corner cases involving nesting that I couldn't quite wrap my head around.

Some thoughts for posterity:

1. DeprecatedPipeline indeed seems like a promising approach
2. There are current tests that assert that empty pipeline blocks are fine; that might thwart your idea of checking empty pipelines
3. Even if we assert block.arity == 1, we'd probably still want checks that the global client wasn't interacted with

byroot:
> 2\. There are current tests that assert that empty pipeline blocks are fine; that might thwart your idea of checking empty pipelines

We can add a `allow_empty: true` parameter, not a huge deal. As long as it's only deprecations for now it's fine.

> Even if we assert block.arity == 1, we'd probably still want checks that the global client wasn't interacted with

Definitely, it would just be a good way to catch obvious mistakes.

> I have unfortunately run through my budget on being able to work on this

No worries. I'll find some time to do it.

PR #1059, opened 2022-01-20
Deprecate `Redis#queue`, `Redis#commit` and `Redis#pipelined` without block.

casperisfine:
All theses make a lot of assumptions on the threading model and are barely thread safe.

The new favored API is:

```ruby
redis.pipelined do |pipeline|
  pipeline.get("foo")
  pipeline.del("bar")
end
```

This API allow multiple threads to build pipelines concurrently on the same connection, and is more friendly to Fiber based concurrency.

Fix: https://github.com/redis/redis-rb/pull/1057

The same should be applied to `Redis#multi`, and we might want to add an easy way to turn off deprecation warnings.

FYI: @machty

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -132,3 +132,7 @@ Metrics/BlockNesting:
 
 Style/HashTransformValues:
   Enabled: false
+
+Style/SymbolProc:
+  Exclude:
+    - 'test/**/*'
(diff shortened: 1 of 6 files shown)


Envek:
In which version this “favored API” with block param was introduced?

> The new favored API is:
> ```ruby
> redis.pipelined do |pipeline|
> ```

This is required to properly constrain redis gem versions in dependent gems and applications. Thanks.

casperisfine:
In `3.0.0`.

yorozen:
Hi,
Not fully understood, the API usage was changed in 4.6.0, but we can still use the old method.
In 5.0.0 the old usage will be blocked?

PR #1061, opened 2022-01-21
Deprecate calling commands on the original Redis instance in multi (block)

casperisfine:
Followup https://github.com/redis/redis-rb/pull/1059 for `multi`.

The new favored API is:

```ruby
redis.multi do |transaction|
  transaction.get("foo")
  transaction.del("bar")
end
```

This API allow multiple threads to build pipelines concurrently on the same
connection, and is more friendly to Fiber based concurrency.

Fix: https://github.com/redis/redis-rb/pull/1057

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,20 @@
     pipeline.get("key")
   end
   ```
+* Deprecate calling commands on `Redis` inside `Redis#multi`. See #1059.
+  ```ruby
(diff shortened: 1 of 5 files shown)


capripot:
Would this deprecation also apply for calls to `Redis#multi` without a block?

byroot:
> Would this deprecation also apply for calls to `Redis#multi` without a block?

No.

PR #1063, opened 2022-02-02
Fix cluster node selection for SCAN, SSCAN, HSCAN, ZSCAN

cmcinnes-mdsol:
We found that while using `SSCAN` on a stable cluster (where no hash slots are being moved from one node to another) we were getting a lot of `MOVED` responses. 

Upon investigation we saw that for `SSCAN` the client was selecting a random writer node. I may be missing the original intent of the removed branch of the `case` statement but it seems like we shouldn't care about the position of the `match` option as the argument for it is a pattern not a key. Before this change `Cluster#find_node_key` was returning `nil` for all flavors of scan which was causing `Cluster#find_node` to select a [random node](https://github.com/redis/redis-rb/blob/9446688ab50386c4b05e9c2da9b1bdbe5a718aae/lib/redis/cluster.rb#L297). Tested out this branch and the `MOVED` responses went away while `SSCAN` continued to function normally.

### DIFF
--- a/lib/redis/cluster/command.rb
+++ b/lib/redis/cluster/command.rb
@@ -53,8 +53,6 @@ class Redis
         when 'object' then 2
         when 'memory'
           command[1].to_s.casecmp('usage').zero? ? 2 : 0
-        when 'scan', 'sscan', 'hscan', 'zscan'
-          determine_optional_key_position(command, 'match')
         when 'xread', 'xreadgroup'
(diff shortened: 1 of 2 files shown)


supercaracal (approved):

cmcinnes-mdsol:
thanks for the quick merge! do we think this is worth a changelog entry? if so can open another pr

PR #1066, opened 2022-02-04
Fix reutrning in subscription callbacks cause race condition

### LINKED ISSUE #1065: Returning in subscription callbacks cause race condition
Returning in subscription callbacks cause race condition which is Redis still send us messages that are published to the channel, but redis-rb do not receive it as Pub/Sub messages.
And if we use another command in that state, we get Pub/Sub messages as reply of the command.

E.g.
```ruby
redis.subscribe('channel_0') do |on|
  on.message do |_channel, _message| 
    return
  end
end

# Redis still send us messages which is published to a 'channel_0'

res = redis.get('hoge')

puts res
# => ['message', 'channel_0', 'message published to a channel_0']
```

this could cause bugs like getting messages from chat which is not intended or getting Pub/Sub messages as a reply of other methods depends on user implementation.

kazzix14:
This PR fixes #1065

### DIFF
--- a/lib/redis/subscribe.rb
+++ b/lib/redis/subscribe.rb
@@ -47,8 +47,8 @@ class Redis
         unsubscribed = type == stop && rest.last == 0
         break if unsubscribed
       end
-      # No need to unsubscribe here. The real client closes the connection
-      # whenever an exception is raised (see #ensure_connected).
+    ensure
+      @client.disconnect unless unsubscribed
     end
   end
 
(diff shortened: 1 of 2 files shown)


byroot:
Could you produce a test case for this? It's really a fairly sensitive thing, so I'm not really willing to change this without proper test coverage on it.

kazzix14:
I'll add tests for this PR when I have time. please wait for it.

kazzix14:
I wrote a test, but I'm not sure it is enough. Please teach me if there are other test cases that we should cover.

byroot:
Closing as stale. The bug should be solved in 5.x already.

Issue #1074, opened 2022-02-17
Connecting to Azure Redis Cluster


ftlc:
Hey folks, 

We have a rails app we're trying to connect to azure redis. I have a test cluster provisioned, and can connect as a single instance, but trying to connect in clustered mode gives an error: `Redis Client could not connect to any cluster nodes`

Here's a minimal config: 

```
host = 'hostname'
key = 'password'
port = 6380
database = 0

connection_url = "rediss://:#{CGI.escape(key)}@#{host}:#{port}/#{database}"

client = Redis.new(cluster: [connection_url])
```

The documentation stated that I can pass in one instance like this and it'll discover the rest of the nodes through the CLUSTER NODES command.

If I switch from `cluster` to `url` I can connect to a redis node, but of course half the writes fail with a MOVED message. 

I was able to find one SO post of somebody having a similar issue in 2019 that was never resolved, but otherwise no documentation that's redis-rb and Azure Redis specific

supercaracal:
The above way of connecting Redis cluster looks like legitimate. I will try to enhance the error messages. It might be caused by ACL.

* #1054
* [Redis commands not supported in Azure Cache for Redis](https://docs.microsoft.com/en-us/azure/azure-cache-for-redis/cache-configure#redis-commands-not-supported-in-azure-cache-for-redis)
* [What features aren't supported with Redis 6?](https://docs.microsoft.com/en-us/azure/azure-cache-for-redis/cache-how-to-version#what-features-arent-supported-with-redis-6)

ftlc:
I'm trying to connect to a Redis6 cluster, but it doesn't support ACL. Is there another way around this?

ftlc:
Thanks for the extra logging! Very helpful. The actual error I'm getting is `WRONGPASS invalid username-password pair (Redis::Cluster::InitialSetupError)` 

I'm not actually passing a username, only a `password@hostname`. Interesting this is, that configuration worked for non-clustered redis, or if I'm trying to call it as a single instance with `url: connection_string`. Is the client doing something different when connecting as a cluster? Or is it likely to be some config in Azure redis itself?

ftlc:
I was looking at the cluster_client_options_test.rb file on my own.


```
option = Redis::Cluster::Option.new(cluster: %w[redis://bazzap:@127.0.0.1:7000], username: 'foobar') 
[...]

Issue #1054, opened 2022-01-05
Configuration for cluster using ACL


Dralun:
Basically I'm trying to move over to using ACL for clusters, but for the life of me I can't get the configuration to work. For some reason adding a username to the cluster seems to break the connection.
What I'm trying to do is the ruby equivalent to this redis-cli line `redis-cli -c -h host_ip_here --user username --pass password`, but I'm unable to reproduce that logic with `redis-rb`.

I've tried `Redis.new(cluster: ["redis://username:password@redis_host:6379"])`, `Redis.new(cluster ["redis://redis_host:6379"], username: username, password: password)` and all other possible combinations of passing user and password in URI or as separate parameters, but nothing manages to connect.

`Redis.new(uri: "redis://username:password@redis_host:6379")` works fine, but only connects to one node, which is not what I want to do.

Is it currently possible to connect to a cluster using ACL with `redis-rb`? I can't find any documentation or any references elsewhere.

supercaracal:
> Is it currently possible to connect to a cluster using ACL with `redis-rb`?

Yes, it is.

* #967

Would you inform us of the followings?

1. redis' version
2. rerdis-rb's version
3. stack trace

There might be a bug.

Dralun:
Curious, that was the syntax that I attempted to use.

1) Redis version: 6.2.6
2) Redis-rb version: 4.5.1 (also tried with 4.4.0)
3) Stack trace:
```
irb(main):004:0> redis = Redis.new(cluster: ["redis://username:password@redis-cluster.clustername.svc.cluster.local:6379"])
/usr/local/bundle/gems/redis-4.5.1/lib/redis/cluster/slot_loader.rb:22:in `load': Redis client could not connect to any cluster nodes (Redis::CannotConnectError)
	from /usr/local/bundle/gems/redis-4.5.1/lib/redis/cluster.rb:116:in `fetch_cluster_info!'
	from /usr/local/bundle/gems/redis-4.5.1/lib/redis/cluster.rb:26:in `initialize'
	from /usr/local/bundle/gems/redis-4.5.1/lib/redis.rb:65:in `new'
	from /usr/local/bundle/gems/redis-4.5.1/lib/redis.rb:65:in `initialize'
	from (irb):4:in `new'
	from (irb):4:in `<main>'
	from /usr/local/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
	from /usr/local/bin/irb:23:in `load'
[...]

PR #1075, opened 2022-02-18
Improve error message for initial setup of cluster mode

supercaracal:
We are currently hard to find out causes when error occured while fetching cluster state information. So this PR adds a specific error class and messages.

current message:
```
Redis Client could not connect to any cluster nodes (Redis::CannotConnectError)
```

improved messages:
```
Redis client could not fetch cluster information: ERR SELECT is not allowed in cluster mode (Redis::Cluster::InitialSetupError)
Redis client could not fetch cluster information: ERR This instance has cluster support disabled (Redis::Cluster::InitialSetupError)
Redis client could not fetch cluster information: Error connecting to Redis on 127.0.0.1:7006 (Errno::ECONNREFUSED) (Redis::Cluster::InitialSetupError)
Redis client could not fetch cluster information: NOPERM this user has no permissions to run the 'cluster' command or its subcommand (Redis::Cluster::InitialSetupError)
```

related issues:
* #1054
* #1074

### DIFF
--- a/lib/redis/cluster/command_loader.rb
+++ b/lib/redis/cluster/command_loader.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require_relative '../errors'
+require 'redis/errors'
 
 class Redis
   class Cluster
@@ -10,15 +10,15 @@ class Redis
       module_function
 
       def load(nodes)
-        nodes.each do |node|
+        errors = nodes.map do |node|
           begin
             return fetch_command_details(node)
-          rescue CannotConnectError, ConnectionError, CommandError
-            next # can retry on another node
+          rescue CannotConnectError, ConnectionError, CommandError => error
+            error
           end
         end
 
-        raise CannotConnectError, 'Redis client could not connect to any cluster nodes'
+        raise InitialSetupError, errors
       end
 
       def fetch_command_details(node)
--- a/lib/redis/cluster/node_loader.rb
+++ b/lib/redis/cluster/node_loader.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require_relative '../errors'
+require 'redis/errors'
 
 class Redis
   class Cluster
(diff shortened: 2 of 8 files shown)


lib/redis/cluster/command_loader.rb
```diff
@@ -10,15 +10,15 @@ module CommandLoader
...
 
       def load(nodes)
-        nodes.each do |node|
+        errors = nodes.map do |node|
           begin
             return fetch_command_details(node)
-          rescue CannotConnectError, ConnectionError, CommandError
-            next # can retry on another node
+          rescue CannotConnectError, ConnectionError, CommandError => e
+            next e
```
byroot, line 17: nitpick, I think that `next` is making the code confusing.

Since it's on the last line you can  just let the expression return by itself.

```suggestion
          rescue CannotConnectError, ConnectionError, CommandError => error
            error
```

lib/redis/cluster/node_loader.rb
```diff
@@ -9,16 +9,15 @@ module NodeLoader
...
-        info = {}
-
-        nodes.each do |node|
-          info = fetch_node_info(node)
-          info.empty? ? next : break
+        errors = nodes.map do |node|
+          begin
+            return fetch_node_info(node)
+          rescue CannotConnectError, ConnectionError, CommandError => e
+            next e
```
byroot, line 16: Same here.

lib/redis/cluster/slot_loader.rb
```diff
@@ -10,25 +10,22 @@ module SlotLoader
...
-        info = {}
-
-        nodes.each do |node|
-          info = fetch_slot_info(node)
-          info.empty? ? next : break
+        errors = nodes.map do |node|
+          begin
+            return fetch_slot_info(node)
+          rescue CannotConnectError, ConnectionError, CommandError => e
+            next e
```
byroot, line 17: Same here.

byroot (approved):
One very small code style nitpick, other than that it looks very good.

As always, Thank you @supercaracal !

Issue #828, opened 2019-02-27
hiredis now supports SSL


tjwallace:
SSL support in hiredis has been merged: https://github.com/redis/hiredis/pull/645

Relevant code: https://github.com/redis/redis-rb/blob/master/lib/redis/connection/hiredis.rb#L17
hiredis-rb issue: https://github.com/redis/hiredis-rb/issues/58

espen:
Would just removing that line fix this or is there anything else that needs to be done?

byroot:
redis-rb 5.0 now uses `redis-client` under the hood, which does have `hiredis` bindings with SSL support.

PR #1077, opened 2022-02-21
Remove error caused by hiredis ssl not supported

ke-miz:
SSL support in hiredis was merged three years ago. So no need for ssl error anymore.
https://github.com/redis/redis-rb/issues/828

### DIFF
--- a/lib/redis/connection/hiredis.rb
+++ b/lib/redis/connection/hiredis.rb
@@ -15,8 +15,6 @@ class Redis
 
         if config[:scheme] == "unix"
           connection.connect_unix(config[:path], connect_timeout)
-        elsif config[:scheme] == "rediss" || config[:ssl]
-          raise NotImplementedError, "SSL not supported by hiredis driver"
         else
           connection.connect(config[:host], config[:port], connect_timeout)
         end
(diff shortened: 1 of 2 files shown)


stanhu:
@byroot @ke-miz This should be reverted. `hiredis` supports SSL, but `hiredis-rb` does not yet: https://github.com/redis/hiredis-rb/issues/58#issuecomment-1210983319

byroot:
You're right, but I disagree it should be reverted. If an adapter doesn't support something, it should raise itself, not rely on `redis-rb` to prevent that call from happening. 

I'd rather leave it as is. And once I migrate `redis-rb` to use `redis-client` it won't matter anymore since it supports SSL with hiredis.

stanhu:
Oh, great, `redis-client` looks much better than the current state of `hiredis-rb`. 

> If an adapter doesn't support something, it should raise itself, not rely on redis-rb to prevent that call from happening.

In this case, `lib/redis/connection/hiredis.rb` doesn't ever let `hiredis-rb` know that SSL is being used, so how would that work at the moment? I think `hiredis-rb` would need to be modified to take in the scheme or SSL options, and then raise the exception.

byroot:
> In this case, `lib/redis/connection/hiredis.rb` doesn't ever let `hiredis-rb` know that SSL is being used

Arf, right, I remembered the `ssl: true` being passed, but you're right it's not the case. I guess I'll revert ...

PR #1078, opened 2022-02-23
Fix SSL/TLS support issue for cluster mode

supercaracal:
At this moment, since there is a bug, we cannot use SSL/TLS support for cluster mode. [Internal client](https://github.com/redis/redis-rb/blob/master/lib/redis/client.rb) accepts `ssl` option but cluster client is passing `scheme` option . `scheme` option has no effect in internal client if `url` option isn't passed. This PR will fix it.

ref:
* #1074
* [TLS Support - Redis](https://redis.io/topics/encryption)


supercaracal:
https://github.com/redis/redis-rb/blob/a7c2a5bfacf9176494b705eebfc4479b1cc79cdd/lib/redis/connection/ruby.rb#L300-L317

PR #1080, opened 2022-02-24
[Redis 6.2] Add GEOSEARCH command

uxgnod:
Add support for [GEOSEARCH](https://redis.io/commands/geosearch) command, which is available since 6.2.0.

Reference: https://github.com/redis/redis-rb/issues/978

### DIFF
--- a/lib/redis/commands/geo.rb
+++ b/lib/redis/commands/geo.rb
@@ -51,6 +51,43 @@ class Redis
         send_command([:georadiusbymember, *geoarguments])
       end
 
+      # Query a sorted set representing a geospatial index to fetch members
+      # within the borders of the area specified by a given shape
+      # from a point or an already existing member
+      #
+      # @param [String] key
+      # @params [String, Array<String>] from use the position of a existing <member> in the sorted set,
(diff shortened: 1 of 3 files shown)


byroot:
The CI failure is legit, you have a syntax error.

As for the API, I'm not a fan of these keyword being used for two different things based on the passed type, it seems brittle to me.

I think I'd prefer:

```ruby
def geosearch(key, from_member: nil, from_lonlat, by_radius: nil, by_box: nil, **options)
end
```

uxgnod:
> The CI failure is legit, you have a syntax error.
> 
> As for the API, I'm not a fan of these keyword being used for two different things based on the passed type, it seems brittle to me.
> 
> I think I'd prefer:
> 
> ```ruby
> def geosearch(key, from_member: nil, from_lonlat, by_radius: nil, by_box: nil, **options)
> end
> ```

Thanks for the advice, I really don't have much experience and I will try to fix these errors.

Also, can you please tell me how I should test this code locally?

byroot:
> Also, can you please tell me how I should test this code locally?

You can run `make` then `make test`.

PR #1084, opened 2022-02-26
Disable SNI if SAN of certificate is a IP address

supercaracal:
* #1074
  * According to the report, [redis-cli](https://redis.io/topics/rediscli) works fine but our redis-rb doesn't work yet.
  * This is a rare use case but a issue indeed.
* Server Name Indication
  * https://datatracker.ietf.org/doc/html/rfc3546#section-3.1
* Subject Alternative Name
  * https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6

```mermaid
sequenceDiagram
    participant Client
    participant Server Shard 1
    participant Server Shard 2
    participant Server Shard 3

    Client->>+Server Shard 1: send CLUSTER NODES to rediss://endpoint.example.com:6379
    Server Shard 1-->>-Client: reply 192.168.13.11:7000, 192.168.13.11:7001, 192.168.13.11:7002
    Note over Client,Server Shard 1: Maybe using virtual IP address internally or be running on single node

    Client->>+Server Shard 3: send GET key3 to rediss://192.168.13.11:7002
    Server Shard 3-->>-Client: reply value
    Note over Client,Server Shard 3: Client should be able to connect to IP address with SSL/TLS
```

### DIFF
--- a/lib/redis/connection/ruby.rb
+++ b/lib/redis/connection/ruby.rb
@@ -6,6 +6,7 @@ require "redis/errors"
 
 require "socket"
 require "timeout"
+require 'resolv'
 
 begin
   require "openssl"
@@ -252,7 +253,7 @@ class Redis
           ctx.set_params(ssl_params || {})
 
           ssl_sock = new(tcp_sock, ctx)
-          ssl_sock.hostname = host
+          ssl_sock.hostname = host unless Resolv::AddressRegex.match?(host)
 
           begin
             # Initiate the socket connection in the background. If it doesn't fail


lib/redis/connection/ruby.rb
```diff
@@ -252,7 +253,7 @@ def self.connect(host, port, timeout, ssl_params)
           ctx.set_params(ssl_params || {})
 
           ssl_sock = new(tcp_sock, ctx)
-          ssl_sock.hostname = host
+          ssl_sock.hostname = host unless Resolv::AddressRegex.match?(host)
```
supercaracal, line 256: https://docs.ruby-lang.org/en/master/Resolv.html

> Address `Regexp` to use for matching IP addresses.

supercaracal:
I've tested with the following example code but I've received a error if hostname is nil.

> No SNI provided; please fix your client., CN=invalid2.invalid

```ruby
# frozen_string_literal: true

require 'socket'
require 'openssl'

IP_ADDR = '172.217.31.142'
HOSTNAME = 'encrypted.google.com'

tcp_sock = TCPSocket.new(IP_ADDR, 443)
ssl_sock = OpenSSL::SSL::SSLSocket.new(tcp_sock, OpenSSL::SSL::SSLContext.new)

unless ARGV.size.zero?
  case ARGV[0]
  when 'host' then ssl_sock.hostname = HOSTNAME
  when 'ip' then ssl_sock.hostname = IP_ADDR
  end
end

ssl_sock.connect
ssl_sock.write('foobar')
ssl_sock.peer_cert.to_text.split("\n").each { |line| puts line }
ssl_sock.close
tcp_sock.close
```

PR #1090, opened 2022-03-20
Add support for EXPIRE command group options (NX/XX/GT/LT)

araishikeiwai:
I was needing this and was already working on this halfway when I realised this is version 7.0 options only.

I'm putting this PR anyway for future use, is there any label for this?


byroot:
The code looks good, but if we start adding Redis 7 features, we need to add it to CI as well.

Issue #1106, opened 2022-06-30
OpenSSL::SSL::SSLError: SSL_read: unexpected eof while reading


jeremiemv:
Hello 👋

On Heroku, after upgrading to the `heroku-22` stack from `heroku-20` we started having intermittent `SSL_read: unexpected eof while reading` exceptions when enqueuing jobs via `sidekiq`. 

The root cause seems to be related to the strict unexpected EOFs implementation of OpenSSL 3. `heroku-22` uses solely OpenSSL 3 ([Ubuntu 22.04 ships with OpenSSL 3.0. OpenSSL 1.1 is not available as a runtime library](https://devcenter.heroku.com/articles/heroku-22-stack#openssl-3)).

Other projects faced the same issue. Ruby and other langauges.
For example https://github.com/puma/puma/pull/2800

Ruby version: 3.1.2
Rails version: 7.0.3
Sidekiq: 6.5.1
redis gem: 4.6.0
Heroku Redis: 6.2.3, premium-0

```
**OpenSSL::SSL::SSLError: SSL_read: unexpected eof while reading**
/app/vendor/ruby-3.1.2/lib/ruby/3.1.0/openssl/buffering.rb line 214 in sysread_nonblock
/app/vendor/ruby-3.1.2/lib/ruby/3.1.0/openssl/buffering.rb line 214 in read_nonblock
/app/vendor/bundle/ruby/3.1.0/gems/redis-4.6.0/lib/redis/connection/ruby.rb line 55 in block in _read_from_socket
/app/vendor/bundle/ruby/3.1.0/gems/redis-4.6.0/lib/redis/connection/ruby.rb line 54 in loop
/app/vendor/bundle/ruby/3.1.0/gems/redis-4.6.0/lib/redis/connection/ruby.rb line 54 in _read_from_socket
/app/vendor/bundle/ruby/3.1.0/gems/redis-4.6.0/lib/redis/connection/ruby.rb line 47 in gets
/app/vendor/bundle/ruby/3.1.0/gems/redis-4.6.0/lib/redis/connection/ruby.rb line 382 in read
/app/vendor/bundle/ruby/3.1.0/gems/redis-4.6.0/lib/redis/client.rb line 311 in block in read
/app/vendor/bundle/ruby/3.1.0/gems/redis-4.6.0/lib/redis/client.rb line 299 in io
/app/vendor/bundle/ruby/3.1.0/gems/redis-4.6.0/lib/redis/client.rb line 310 in read
[...]

PR #1107, opened 2022-06-30
Handle OpenSSL v3 EOF errors

casperisfine:
Fix: https://github.com/redis/redis-rb/issues/1106

When Ruby is compiled against OpenSSL V3, if an SSL connection is closed without sending a `close_notify`, a `OpenSSL::SSL::SSLError` will be raised.

I tested this manually, it works s expected. Unfortunately we don't have SSL setup on CI, and it's a big pain to do (I did it for `redis-client`), so I'm tempted to ship with just a manual test, or maybe just a mock? 😞

### DIFF
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -302,7 +302,7 @@ class Redis
       e2 = TimeoutError.new("Connection timed out")
       e2.set_backtrace(e1.backtrace)
       raise e2
-    rescue Errno::ECONNRESET, Errno::EPIPE, Errno::ECONNABORTED, Errno::EBADF, Errno::EINVAL => e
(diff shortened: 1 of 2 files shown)


davidrunger:
I deployed this branch to production using heroku-22 1.5 hours ago. No errors this time! 👍 🎉 (Previously, with the released `redis` gem version 4.7.0, I'd see multiple "unexpected eof while reading" errors per hour when using heroku-22.)

brent-yearone:
I ran into this problem on Heroku last week, and worked around it by setting the redis-side connection timeouts to 25 hours. Thanks a bunch for the fix! 😁

ezekg:
Been running 4.7.1 in production and everything seems to be good now. 👍

yashka713:
Today we are faced with 

```ruby
OpenSSL::SSL::SSLError SSL_read: unexpected eof while reading
```

Our current Redis Gem version is "4.8.1". But Heroku version is heroku-22. What could be the reason of this error?

Issue #1113, opened 2022-08-08
Redis 7 support


voxik:
Since Redis 7 landed in Fedora, we observe several failing test cases:

~~~
+ make BINARY=/usr/bin/redis-server REDIS_CLIENT=/usr/bin/redis-cli 'BUILD_DIR=${TMP}'
Run options: --seed 20345

# Running:

.....................................................................................................................................................................E.........E........................F...F..F.........................................................................................................................E......F..........................................Error running mock server: Errno::ECONNRESET: Connection reset by peer @ io_fillbuf - fd:8 
/builddir/build/BUILD/redis-4.7.1/usr/share/gems/gems/redis-4.7.1/test/support/redis_mock.rb:87:in `gets'
/builddir/build/BUILD/redis-4.7.1/usr/share/gems/gems/redis-4.7.1/test/support/redis_mock.rb:87:in `block in start'
/builddir/build/BUILD/redis-4.7.1/usr/share/gems/gems/redis-4.7.1/test/support/redis_mock.rb:40:in `block in run'
/builddir/build/BUILD/redis-4.7.1/usr/share/gems/gems/redis-4.7.1/test/support/redis_mock.rb:36:in `loop'
/builddir/build/BUILD/redis-4.7.1/usr/share/gems/gems/redis-4.7.1/test/support/redis_mock.rb:36:in `run'
/builddir/build/BUILD/redis-4.7.1/usr/share/gems/gems/redis-4.7.1/test/support/redis_mock.rb:28:in `block in start'
.....................F............................................................................................................................................................F......................F.....F.............................................................................................................................................................................................................................F....................................................................................................................................................F....F...............................................................................................................................................................................E....FEE....E.E.EE.E.F.E...................................................................................................................................

Finished in 132.480076s, 9.7600 runs/s, 40.2929 assertions/s.

  1) Error:
TestRemoteServerControlCommands#test_info_commandstats:
NoMethodError: undefined method `[]' for nil:NilClass

      assert_equal '2', result['config']['calls']
                                        ^^^^^^^^^
    /builddir/build/BUILD/redis-4.7.1/usr/share/gems/gems/redis-4.7.1/test/remote_server_control_commands_test.rb:33:in `block in test_info_commandstats'
    /builddir/build/BUILD/redis-4.7.1/usr/share/gems/gems/redis-4.7.1/test/helper.rb:158:in `target_version'
[...]

PR #1114, opened 2022-08-09
Add Redis 7.0 to CI matrix

casperisfine:
Fix: https://github.com/redis/redis-rb/issues/1113

Most of the failure were from tests that were asserting implementation details from Redis.


casperisfine:
Cluster tests are flay as often... Other than that it's green.

Issue #1070, opened 2022-02-14
Plan for redis-rb 5.0


byroot:
A large part of redis-rb maintenance cost comes from having to implement each command Redis add.
[A year after Redis 6.2 release, we're still missing a large part of the newly added commands](https://github.com/redis/redis-rb/issues/978),
and tengentially commands added by Redis modules are hard to use.

This comes down to the RESP2 protocol that requires to know how to parse each command output.

However Redis 6.0 introduced [the RESP3 protocol](https://github.com/antirez/RESP3/blob/74adea588783e463c7e84793b325b088fe6edd1c/spec.md), which is properly typed
allowing clients to not know anything about the commands they are sending, which means a drastically simpler and smaller implementation that don't need keeping up with commands added upstream.

So I'd like to split `redis-rb` in multiple gems (probably kept in the same git repository)

### `redis-client` gem

What I'd like to do is to start a new simpler `redis-client` gem, that only support Redis 6.0+ and pretty much only offer
the following API:

```ruby
redis = Redis::Client.new(...)

# sending command
redis.call('SMEMBERS', 'mykey', ...)

# pipelining
redis.pipeline do |pipe|
  pipe.call(...)
end
```

We may also need some API for dealing with subscriptions commands, but that's about it unless I'm missing something.

This simple client would:

  - Not support redis-cluster, it should be implemented as another gem, the vast majority of people don't need it.
  - No longer wrap calls with a mutex, it's a mis-feature of `redis-rb`, to share a client between theads use a thread pool.

Users should be encouraged to use `redis-client` directly.

[I explored this avenue back in 2019](https://github.com/byroot/resp3).

### `redis-cluster-client` gem

Clustering support adds lots of complexity, so I think it should be shipped separately for people that need it.

It should however be API compatible.

One difficulty with clustering is that the client must be able to extract the parts of the command that are keys.

Right now this is handled with lots of ad hoc code for each command, but we should instead load the list of command signatures from the redis server with [`COMMAND`](https://redis.io/commands/command) and extract the keys that way.

With the above `redis-cluster-client` should be able to be a drop-in replacement for `redis-client`

### `redis-rb` 5.0

`redis-rb` should then be refactored to be a thin wrapper around `redis-client` allowing to keep backward compatibility for the most part. Users would still be encouraged to migrate to `redis-client` even though there is not plan to sunset the `redis` gem any time soon.

[...]

PR #1120, opened 2022-08-12
redis-rb 5.0

byroot:
Ref: https://github.com/redis/redis-rb/issues/1070

Ok, let's start the work to see what kind of roadblocks we'll run into.

One idea I'm toying with right now, is that rather than delegate everything to `redis-client` which would require Redis 6+, we could use `redis-client` with the`RESP2` protocol and keep all our type casting. The downside is that it doesn't allow to delete as much code, the upside is that we don't break people who are still on Redis 5, and we get the much better drivers of `redis-client`.


mperham:
I'd suggest being more aggressive. Why not drop Redis 5? redis-rb 4.x supports Redis 5 just fine, people can continue using it if they don't want to upgrade for some reason. Redis 6.0 is from April 2020 and every major Redis provider supports 6.x.

byroot:
> Why not drop Redis 5?

The problem isn't so much Redis 5, but RESP2. Some people migrated to Redis 6 or 7, but used proxies such as https://github.com/envoyproxy/envoy/issues/17571 which don't yet support the newer protocol. That pains me but...

So if I can support RESP2 in redis-rb 5.0, I can remove all the deprecated stuff and allow these people to upgrade, and in a shorter time I can drop RESP2 support in 6.0.

That's the idea so far, we'll see how it goes. In https://github.com/casperisfine/redis-rb/pull/1 a surprising amount of tests pass with extremely little effort.

casperisfine:
Quick status update.

- The main test suite pass.
- The hiredis suite passes as well (including SSL tests).
- The sentinel suite pass but is still super flaky, maybe a bit more than before.
- I still need to look at the cluster test suite, but it seems to work for the most part.

casperisfine:
Ok, so another status update:

  - sentinel test suite is still super flaky, there might be some legit error in there, worth investigating.
  - cluster test suite: `412 runs, 1795 assertions, 6 failures, 7 errors,`, https://github.com/casperisfine/redis-rb/pull/3


Outstanding issues with `cluster`:

  - `redis-cluster-client` requires `ruby 2.7`, not sure if necessary or not, but that makes testing harder.
  - There's a bunch of properties normally access on the config object that are broken (e.g. `client.id`). I need to iterate on all the node, that may require some small change in `cluster-client`.
  - A handful of broken tests around transactions and pipelining, I'll have to dig into them.

Overall I'm quite tempted to extra `Redis::Cluster` to its own gem (or inside `redis-cluster-client`?). But not sure how much it would complicate testing. @supercaracl what do you think?

casperisfine:
After sleeping on it, I think short term the best is to indeed split it in another gem, but keep it in the same repo so that it's easy to share code.

One issue though is that the gem name is already in use -_- https://rubygems.org/gems/redis-cluster

casperisfine:
Ok so:

  - The `cluster` suite now pass:
    - We have a regression on transactions (`multi`). Unless I missed something `redis-cluster-client` outright don't support them, we used to support some of them before. Not sure how big a deal it is.
  - `sentinel` is still flaky I still need to debug, but I confirmed at least a part of the failures are happening consistently with specific seeds, so I think some tests put the sentinel group in a bad state and make later tests fail.
  - I tested this branch against `sidekiq` and `activesupport` test suites, aside from some test only code they have to select the driver, they all pass.

I think I'll bump the version number and merge this this afternoon, further changes can happen on master. I've pushed a `4.x` branch in case we want to backport some stuff to 4.x.

casperisfine:
Oh, and `jruby` is flaky too now...

byroot:
Aaaand it's green!

Further work will happen on master.

supercaracal:
I’m sorry for the late reply. I agree with the decision.

ZackMattor:
@byroot I might create a new issue... and let me know if I should, but wanted to see if I was missing anything. We use clustering for our Rails Cache (using the official `redis_cache_store` adapter in rails), and it seems like this change makes it so there is no easy path forward for us to upgrade to redis > 5... Before you could just pass the `cluster` parameter to the `redis_cache_store` adapter and it would work as intended and now it doesn't seem like it can support using Redis::Cluster.

https://github.com/rails/rails/blob/6-0-stable/activesupport/lib/active_support/cache/redis_cache_store.rb#L116-L128

Maybe I should work with the rails project to get clustering support working over there.

byroot:
Rails `RedisCacheStore` accept a `Redis` instance as argument:

```ruby
ActiveSupport::Cache::RedisCacheStore.new(redis: Redis::Cluster.new(....))
```

or 

```ruby
config.cache_store = :redis_store, { redis: Redis::Cluster.new(....) }
```

NB: I didn't test, I may have got some of the syntax partially wrong, but the general idea works.

PR #1123, opened 2022-08-12
Fix jruby CI

casperisfine:
I was hopping that removing the old openssl gem would fix it, but somehow it's still get stuck on CI.

Which is weird because it works relatively well locally.

If someone feels like digging into this.

### DIFF
--- a/test/support/redis_mock.rb
+++ b/test/support/redis_mock.rb
@@ -30,7 +30,6 @@ module RedisMock
 
     def shutdown
       @thread.kill
-      @thread.join
     end
 
     def run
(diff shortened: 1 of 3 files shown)


test/support/redis_mock.rb
```diff
@@ -30,7 +30,6 @@ def start(&block)
 
     def shutdown
       @thread.kill
-      @thread.join
```
casperisfine, line 33: Ok, so in the end this was the cause. Somehow the above kill doesn't reliably kill the thread, and we'd then be stuck joining forever...

Looks like a bug in JRuby to me, and I don't like that removing this join might cause use to leak threads but...

PR #1128, opened 2022-08-17
Lower client timeout on CI

byroot:
Because of the flakiness of the cluster and sentinel suite we've bumped the timeout to some absurd level. 

Now that the suites have been split and somewhat stabilized I think we can reduced it.

NB: I try the 1.0 default for all suite, but worst case I think we can keep it relatively high on `cluster` & `sentinel`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   lint:
     name: Rubocop
-    timeout-minutes: 30
+    timeout-minutes: 15
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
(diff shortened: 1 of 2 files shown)


byroot:
Interesting, so this reduce CI time way more than I expected 4-ish minutes down to 40-ish seconds. 

This is likely in good part because of a couple badly written tests that timeout after a very long time like this one: https://github.com/redis/redis-rb/blob/85753ef8894e6cdbaeceac0fc108e11dda409e03/redis_cluster/test/abnormal_state_test.rb#L33-L35.

What's even more interesting is that truffle timed out on both runs, on the same test: 

```ruby
TestInternals#test_large_payload:
Redis::TimeoutError: RedisClient::ReadTimeoutError
    /home/runner/work/redis-rb/redis-rb/vendor/bundle/truffleruby/3.0.3.22.2.0/bundler/gems/redis-client-3beae9a11f26/lib/redis_client/ruby_connection/buffered_io.rb:140:in `block in fill_buffer'
    <internal:core> core/kernel.rb:407:in `loop'
```

Maybe Truffle IOs are a bit slower? (Just FYI @eregon). I'll bump the timeout in that test only.

PR #1136, opened 2022-08-23
Add tests for `lpop`/`rpop` with the missing key

fatkodima:
I noticed, that the docs for `lpop`/`rpop` are incomplete regarding missing keys. Also added test cases for this.

### DIFF
--- a/lib/redis/commands/lists.rb
+++ b/lib/redis/commands/lists.rb
@@ -99,7 +99,7 @@ class Redis
       #
       # @param [String] key
       # @param [Integer] count number of elements to remove
-      # @return [String, Array<String>] the values of the first elements
+      # @return [nil, String, Array<String>] the values of the first elements
       def lpop(key, count = nil)
         command = [:lpop, key]
         command << Integer(count) if count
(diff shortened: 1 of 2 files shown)


lib/redis/commands/lists.rb
```diff
@@ -99,18 +99,21 @@ def rpushx(key, value)
       #
       # @param [String] key
       # @param [Integer] count number of elements to remove
-      # @return [String, Array<String>] the values of the first elements
+      # @return [nil, String, Array<String>] the values of the first elements,
+      #     or nil when the key does not exist
```
byroot, line 103: If the key doesn't exist, it's treated as an empty list.

I don't think we need to point that out particularly. Otherwise we'd have to say that on pretty much all commands documentation.

byroot: NB: I'm only talking about the text change, the added `nil` type is fine by me.

lib/redis/commands/lists.rb
```diff
@@ -99,18 +99,21 @@ def rpushx(key, value)
...
       def lpop(key, count = nil)
         command = [:lpop, key]
         command << Integer(count) if count
         send_command(command)
       end
 
+      # nil
```
byroot, line 110: ✂️ ?

fatkodima:
@byroot Updated.

I wish all the OSS maintainers were so fast 😄❤️

PR #1146, opened 2022-09-02
fix changelog instructions for enabling 5.0 behavior

bpo:
Hello, this minor instruction in the changelog is incorrect.

Default behavior for 4.x: `Redis.sadd_returns_boolean = true`
5.x behavior: `Redis.sadd_returns_boolean = false`

This PR is against the 4.x branch, but you'll probably want to apply to main as well.


byroot:
Thanks!

PR #1153, opened 2022-09-10
Shrink cluster test cases guaranteed by redis-cluster-client and focus on behavior test for method interfaces

supercaracal:
Since redis-clustering gem depends on redis-cluster-client gem, I think redis-clustering gem should test behavior focused on method interfaces.

* abnormal_state_test.rb
  * It takes a lot of wasting time for CI, although it is guaranteed by redis-cluster-client. So we can get rid of it.
* client_replicas_test.rb
  * It is guaranteed by redis-cluster-client. So we can syrink the test cases.
  * Also, redis-cluster-client does more test cases with multiple replicas.
* client_transaction_test.rb
  * It is simply needed a test case that throws an error.


byroot:
From my point of view, you're the official maintainer of the cluster support, so up to you 👍

PR #1156, opened 2022-09-20
Improve performance of mapped method's hash flattening

roharon:
In `mapped_hmset`, `mapped_mset` and `xadd` method,
Hash variable does not need to be cast as Array type for flattening.

Also In benchmark,
`hash.flatten` is faster than `hash.to_a.flatten`.

So I think it would be better.

```
hash = {'key1': '1', 'key2': '2', 'key3': '3', 'key4': '4'}

Benchmark.bm do |x|
  x.report { hash.to_a.flatten }
  x.report { hash.flatten }
end

=== OUTPUT ===

       user     system      total        real
   0.000011   0.000002   0.000013 (  0.000009)
   0.000004   0.000001   0.000005 (  0.000004)

```

### DIFF
--- a/lib/redis/commands/hashes.rb
+++ b/lib/redis/commands/hashes.rb
@@ -63,7 +63,7 @@ class Redis
       #
       # @see #hmset
       def mapped_hmset(key, hash)
-        hmset(key, hash.to_a.flatten)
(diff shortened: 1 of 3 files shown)


sangmokang:
Wow That's so great !

PR #1163, opened 2022-11-07
Fix error message on trying to use cluster options

jdelStrother:
redis_cluster is an unrelated gem


byroot:
Thanks! (For context that was the originally planed named until I realized it was taken)

Issue #1165, opened 2022-11-17
XAUTOCLAIM throws NoMethodError if entry is nil


wcmonty:
If an entry in the PEL is deleted, `xautoclaim` returns nil for the event in the response. When processing that event, it is passed to the `HashifyStreamAutoclaim` lambda where it throws an error because the lambda assumes there's an actual event. See the following sample script. I tested with v4.5.1, but it looks like the lambda is [unchanged in the current version](https://github.com/redis/redis-rb/blob/787462e8a120bf08759109c8ff68309d5efebe7f/lib/redis/commands.rb#L119-L124).

```ruby
redis.xgroup(:create, 'key', 'group1', '$', mkstream: true)
# => "OK"
redis.xadd('key', { foo: :bar }, id: '1')
# => "1-0"
redis.xreadgroup('group1', 'consumer1', 'key', '>', count: 1)
# => {"key"=>[["1-0", {"foo"=>"bar"}]]}
redis.xdel('key', '1-0')
# => 1
redis.xautoclaim('key', 'group1', 'consumer2', 0, '0-0', count: 1)
# NoMethodError: undefined method `[]' for nil:NilClass
# 
#      'entries' => reply[1].map { |entry| [entry[0], entry[1].each_slice(2).to_h] }
#                                                ^^^
# from /usr/local/bundle/gems/redis-4.5.1/lib/redis.rb:3704:in `block (2 levels) in <class:Redis>'
```

This is similar to https://github.com/redis/redis-rb/issues/905. It looks like [Redis server is changing its behavior as of v7.0](https://github.com/redis/redis/pull/10227), so this will no longer be an issue.

I think the solution is to compact the array before we map over it.

byroot:
Fixed in https://github.com/redis/redis-rb/pull/1166

PR #1166, opened 2022-11-18
Fix XAUTOCLAIM when entry is deleted

wcmonty:
This addresses https://github.com/redis/redis-rb/issues/1165. 

When a consumer has read an entry, it is added to the PEL. If the event is deleted, the entry in the PEL is not removed. If that is claimed with `XAUTOCLAIM` Redis responds with a null event (at least until version 7.0 where nulls are removed from the response). The HashifyStreamAutoclaim lambda doesn't handle the null value and throws a NoMethodError.

This change just compacts the entries so that nulls are removed.

### DIFF
--- a/lib/redis/commands.rb
+++ b/lib/redis/commands.rb
@@ -119,7 +119,9 @@ class Redis
     HashifyStreamAutoclaim = lambda { |reply|
       {
         'next' => reply[0],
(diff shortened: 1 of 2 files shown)


byroot:
Thanks!

PR #1167, opened 2022-11-18
[Redis 6.2] Add NOMKSTREAM option to XADD

wmontgomery-splunk:
This adds the ability to add the NOMKSTREAM option to XADD. By default it does not add the option to the command, meaning that a stream will be created if it doesn't exist.

From the Redis documentation at: https://redis.io/commands/xadd/
> Appends the specified stream entry to the stream at the specified key. If the key does not exist, as a side effect of running this command the key is created with a stream value. The creation of stream's key can be disabled with the NOMKSTREAM option.

References https://github.com/redis/redis-rb/issues/978

### DIFF
--- a/test/lint/streams.rb
+++ b/test/lint/streams.rb
@@ -3,7 +3,6 @@
 module Lint
   module Streams
     MIN_REDIS_VERSION = '4.9.0'
-    MIN_REDIS_VERSION_XAUTOCLAIM = '6.2.0'
     ENTRY_ID_FORMAT = /\d+-\d+/.freeze
 
     def setup
@@ -87,6 +86,16 @@ module Lint
 
     def test_xadd_with_maxlen_and_approximate_option
       actual = redis.xadd('s1', { f1: 'v1', f2: 'v2' }, maxlen: 2, approximate: true)
+      assert_match ENTRY_ID_FORMAT, actual
+    end
+
+    def test_xadd_with_nomkstream_option
+      omit_version('6.2.0')
+
+      actual = redis.xadd('s1', { f1: 'v1', f2: 'v2' }, nomkstream: true)
+      assert_nil actual
+
+      actual = redis.xadd('s1', { f1: 'v1', f2: 'v2' }, nomkstream: false)
       assert_match ENTRY_ID_FORMAT, actual
     end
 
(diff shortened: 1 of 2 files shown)


test/lint/streams.rb
```diff
@@ -90,6 +90,14 @@ def test_xadd_with_maxlen_and_approximate_option
       assert_match ENTRY_ID_FORMAT, actual
     end
 
+    def test_xadd_with_nomkstream_option
+      actual = redis.xadd('s1', { f1: 'v1', f2: 'v2' }, nomkstream: true)
+      assert_equal nil, actual
```
byroot, line 95:
```suggestion
      assert_nil actual
```

Otherwise minitest will emit a warning.

test/lint/streams.rb
```diff
@@ -90,6 +90,14 @@ def test_xadd_with_maxlen_and_approximate_option
       assert_match ENTRY_ID_FORMAT, actual
     end
 
+    def test_xadd_with_nomkstream_option
```
byroot, line 93: You need a call to `omit_version` here so that this test doesn't run against redis 6 and older.

test/lint/streams.rb
wmontgomery-splunk: Updated in aae269592bc5b30eb2a9927c5e58ac225e375efc.

test/lint/streams.rb
byroot: Please squash your commits though.

test/lint/streams.rb
```diff
@@ -3,7 +3,7 @@
 module Lint
   module Streams
     MIN_REDIS_VERSION = '4.9.0'
-    MIN_REDIS_VERSION_XAUTOCLAIM = '6.2.0'
+    MIN_REDIS_VERSION_6_2 = '6.2.0'
```
byroot, line 6: Might as well use the literal string directly.

wmontgomery-splunk: Does this comment mean that you'd like me to inline or you didn't like the name I chose?

byroot: Please inline it.

test/lint/streams.rb
wmontgomery-splunk: Done.

test/lint/streams.rb
wmontgomery-splunk: Inlined and squashed in 2c0f0dbdb6f465e12655dd7af0b812790d207103.

byroot:
Thank you!

PR #1168, opened 2022-11-19
treat ReadOnlyError as ConnectionError

tylerbrewer2:
`ReadOnlyError` is now considered a connection error in redis-client as of `0.11.2` (https://github.com/redis-rb/redis-client/pull/65). This error is usually raised during failover scenarios and reconnection will likely work.

### DIFF
--- a/lib/redis/errors.rb
+++ b/lib/redis/errors.rb
@@ -26,9 +26,6 @@ class Redis
   class WrongTypeError < CommandError
   end
 
-  class ReadOnlyError < CommandError
-  end
-
   class OutOfMemoryError < CommandError
   end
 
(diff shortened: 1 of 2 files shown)


byroot:
Thanks. A test would be nice, but worst case I can add it myself on Monday.

tylerbrewer2:
Do you want to test that reconnect attempts are made or just that the error is a `CommandError`?

byroot:
I'd like something that fully test yes. You can probably re-use the test in `redis-client`.

tylerbrewer2:
I'm getting an `ECONNRESET` thrown [here](https://github.com/redis-rb/redis-client/blob/master/lib/redis_client/ruby_connection/buffered_io.rb#L145) in the test that I'm not making much progress on. So feel free to take this if you'd like

casperisfine:
Thanks, I made your test pass and merged directly to master.

tylerbrewer2:
Sounds great, thank you!

PR #1170, opened 2022-11-28
[Redis 6.2] Add MINID and LIMIT options to xtrim

wcmonty:
Adds the following options to the xtrim command:
- Set the strategy to MINID
- Set the limit option

This probably isn't the ideal interface for the `#xtrim` method, but I tried not to add breaking changes.

```ruby
# Interface in this PR
xtrim(key, len_or_id, strategy: 'MAXLEN', approximate: false, limit: nil)

# Possible alternative interface (but requires a breaking change)
xtrim(key, maxlen: nil, minid: nil, approximate: false, limit: nil)
```

See the documentation at https://redis.io/commands/xtrim/

References https://github.com/redis/redis-rb/issues/978

### DIFF
--- a/lib/redis/commands/streams.rb
+++ b/lib/redis/commands/streams.rb
@@ -65,14 +65,30 @@ class Redis
       #   redis.xtrim('mystream', 1000)
       # @example With options
       #   redis.xtrim('mystream', 1000, approximate: true)
+      # @example With strategy
+      #   redis.xtrim('mystream', '1-0', strategy: 'MINID')
       #
-      # @param key         [String]  the stream key
-      # @param mexlen      [Integer] max length of entries
-      # @param approximate [Boolean] whether to add `~` modifier of maxlen or not
+      # @overload xtrim(key, maxlen, strategy: 'MAXLEN', approximate: true)
+      #   @param key         [String]  the stream key
+      #   @param maxlen      [Integer] max length of entries
+      #   @param strategy    [String]  the limit strategy, must be MAXLEN
+      #   @param approximate [Boolean] whether to add `~` modifier of maxlen or not
+      #   @param limit       [Integer] maximum count of entries to be evicted
+      # @overload xtrim(key, minid, strategy: 'MINID', approximate: true)
+      #   @param key         [String]  the stream key
+      #   @param minid       [String]  minimum id of entries
+      #   @param strategy    [String]  the limit strategy, must be MINID
+      #   @param approximate [Boolean] whether to add `~` modifier of minid or not
+      #   @param limit       [Integer] maximum count of entries to be evicted
       #
       # @return [Integer] the number of entries actually deleted
-      def xtrim(key, maxlen, approximate: false)
-        args = [:xtrim, key, 'MAXLEN', (approximate ? '~' : nil), maxlen].compact
+      def xtrim(key, len_or_id, strategy: 'MAXLEN', approximate: false, limit: nil)
+        strategy = strategy.to_s.upcase
+
+        args = [:xtrim, key, strategy]
+        args << '~' if approximate
+        args << len_or_id
+        args.concat(['LIMIT', limit]) if limit
         send_command(args)
       end
 
--- a/test/lint/streams.rb
+++ b/test/lint/streams.rb
@@ -121,6 +121,62 @@ module Lint
       assert_equal 0, redis.xtrim('s1', 2, approximate: true)
     end
 
+    def test_xtrim_with_limit_option
+      omit_version('6.2.0')
+
+      begin
+        original = redis.config(:get, 'stream-node-max-entries')['stream-node-max-entries']
+        redis.config(:set, 'stream-node-max-entries', 1)
+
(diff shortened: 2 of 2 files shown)


lib/redis/commands/streams.rb
```diff
@@ -65,14 +65,31 @@ def xadd(key, entry, approximate: nil, maxlen: nil, nomkstream: nil, id: '*')
...
+      #   @param limit       [Integer] maximum count of entries to be evicted
       #
       # @return [Integer] the number of entries actually deleted
-      def xtrim(key, maxlen, approximate: false)
-        args = [:xtrim, key, 'MAXLEN', (approximate ? '~' : nil), maxlen].compact
+      def xtrim(key, len_or_id, strategy: 'MAXLEN', approximate: false, limit: nil)
+        strategy = strategy.to_s.upcase
+        unless %w[MAXLEN MINID].include?(strategy)
+          raise ArgumentError, "strategy must be MINID or MAXLEN"
+        end
```
byroot, line 89: I'm not convinced all this validation is worth it, might as well directly pass it to the server.

lib/redis/commands/streams.rb
```diff
@@ -65,14 +65,31 @@ def xadd(key, entry, approximate: nil, maxlen: nil, nomkstream: nil, id: '*')
...
+
+        args = [:xtrim, key, strategy, (approximate ? '~' : nil), len_or_id].compact
```
byroot, line 91:
```suggestion
        args = [:xtrim, key, strategy]
        args << '!' if approximate
        args << len_or_id
```

I know the code was like this, but I'd rather add to the array only when necessary than to compact nil elements. It's less error prone.

lib/redis/commands/streams.rb
wcmonty: Apparently, we are going to have to either compact/handle the nils in this test case or change the behavior:
https://github.com/redis/redis-rb/blob/master/test/lint/streams.rb#L131 . If we don't then, the we receive a TypeError with this message `Unsupported command argument type: NilClass`.

A couple of options:
- Change the behavior so that we throw the TypeError instead of a Redis::CommandError
- Leave as-is, and compact one
- Combine your suggestion above with a compact before we call `send_command`

What's your preference?

lib/redis/commands/streams.rb
wcmonty: Apparently, we are going to have to either compact/handle the nils in this test case or change the behavior:
https://github.com/redis/redis-rb/blob/master/test/lint/streams.rb#L131 . If we don't then, the we receive a TypeError with this message `Unsupported command argument type: NilClass`.

A couple of options:
- Change the behavior so that we throw the TypeError instead of a Redis::CommandError
- Leave as-is, and compact one
- Combine your suggestion above with a compact before we call `send_command`

What's your preference?

byroot: > or change the behavior:
> https://github.com/redis/redis-rb/blob/master/test/lint/streams.rb#L131

Yeah, that is inconsistent with the rest of the library, just remove that part of the test.

wcmonty: Updated and squashed into 5d7ceea00524d287affb272f8afc791c803b66bc

lib/redis/commands/streams.rb
wcmonty: Updated in 5d7ceea00524d287affb272f8afc791c803b66bc

lib/redis/commands/streams.rb
```diff
@@ -65,15 +65,31 @@ def xadd(key, entry, approximate: nil, maxlen: nil, nomkstream: nil, id: '*')
...
-        args = [:xtrim, key, 'MAXLEN', (approximate ? '~' : nil), maxlen].compact
-        send_command(args)
+      def xtrim(key, len_or_id, strategy: 'MAXLEN', approximate: false, limit: nil)
+        strategy = strategy.to_s.upcase
+
+        args = [:xtrim, key, strategy]
+        args << '~' if approximate
+        args << len_or_id
+        args.concat(['LIMIT', limit]) if limit
+        send_command(args.compact)
```
byroot, line 92:
```suggestion
        send_command(args)
```

wcmonty: Sorry, that slipped through. Updated in 43211594378dc98749f699fce4fa45fbf0343adc

byroot:
The CI failures seem legit. I assume there another test passing nil as argument somewhere.

wcmonty:
So the CI failures were because I expected that the ensure would be skipped in cases where were were skipping the spec. Seems like that's common enough to have it's [own Rubocop cop](https://docs.rubocop.org/rubocop-minitest/cops_minitest.html#minitestskipensure). I pushed up d7926fea412f12b5e9298d8cb445197326265912. If that looks good, I'll squash the changes.

byroot (approved):

wcmonty:
Alright, I squashed the commits. Thanks for working with me on this.

PR #1179, opened 2023-02-02
Add change to default client timeout to CHANGELOG

dbackeus:
After deploying a redis gem upgrade (among other things) today we started getting `Redis::TimeoutError: RedisClient::ReadTimeoutError` every now and then. Took us quite a while to figure out that the default timeout had changed from 5 seconds in 4.x to 1 second in 5.x.

This change deserves a changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 
 # 5.0.0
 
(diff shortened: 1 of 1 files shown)


byroot:
Thanks.

PR #1189, opened 2023-05-01
✨ Add ZMPOP

JerrodCarpenter:
I noticed some 7.0 commands were missing and thought it might be fun to add them. I'm not super familiar with the code base standards, so I'm sure I stepped in it with a couple of decisions, happy to change whatever.

Also, and this is embarrassing, but I couldn't figure out how to run tests locally. 

**Example:**
```ruby
irb(main):001:0> redis = Redis.new(url: "redis://localhost:6379/2")
=> #<Redis client v5.0.6 for redis://localhost:6379/2>
irb(main):002:0> redis.zadd("myzset", [[1, "one"], [2, "two"], [3, "three"]])
=> 3
irb(main):003:0> redis.zmpop("myzset", min_or_max: "MAX", count: 2)
=> ["myzset", [["three", 3.0], ["two", 2.0]]]
```

### DIFF
--- a/lib/redis/commands/sorted_sets.rb
+++ b/lib/redis/commands/sorted_sets.rb
@@ -167,6 +167,42 @@ class Redis
         end
       end
 
+      # Removes and returns up to count members with scores in the sorted set stored at key.
+      #
+      # @example Popping a member
+      #   redis.zmpop('zset')
+      #   #=> ['zset', ['a', 1.0]]
+      # @example With count option
+      #   redis.zmpop('zset', count: 2)
+      #   #=> ['zset', [['a', 1.0], ['b', 2.0]]
+      #
+      # @params key [String, Array<String>] one or more keys with sorted sets
+      # @params modifier [String]
+      #  - when `"MIN"` - the elements popped are those with lowest scores
+      #  - when `"MAX"` - the elements popped are those with the highest scores
+      # @params count [Integer] a number of members to pop
+      #
+      # @return [Array<String, Array<String, Float>>] list of popped elements and scores
+      def zmpop(*keys, modifier: "MIN", count: nil)
+        raise ArgumentError, "Pick either MIN or MAX" unless modifier == "MIN" || modifier == "MAX"
+
+        args = [:zmpop, keys.size, *keys, modifier]
+
+        if count
+          args << "COUNT"
+          args << Integer(count)
+        end
+
+        send_command(args) do |response|
+          response&.map do |entry|
+            case entry
+            when String then entry
+            when Array then entry.map { |pair| FloatifyPairs.call(pair) }.flatten(1)
+            end
+          end
+        end
+      end
+
       # Removes and returns up to count members with the highest scores in the sorted set stored at keys,
       #   or block until one is available.
       #
--- a/lib/redis/distributed.rb
+++ b/lib/redis/distributed.rb
@@ -694,6 +694,13 @@ class Redis
       node_for(key).zmscore(key, *members)
     end
 
+    # Iterate over keys, removing members from the first non empty sorted set found.
+    def zmpop(*keys, modifier: "MIN", count: nil)
+      ensure_same_node(:zmpop, keys) do |node|
+        node.zmpop(*keys, modifier: modifier, count: count)
+      end
+    end
+
     # Return a range of members in a sorted set, by index, score or lexicographical ordering.
     def zrange(key, start, stop, **options)
       node_for(key).zrange(key, start, stop, **options)
--- a/test/lint/sorted_sets.rb
+++ b/test/lint/sorted_sets.rb
@@ -479,6 +479,20 @@ module Lint
       assert_equal [['d', 3.0]], r.zrange('foo', 0, -1, with_scores: true)
     end
 
+    def test_zmpop
+      target_version('7.0') do
+        assert_nil r.zmpop('foo')
+
+        r.zadd('foo', %w[0 a 1 b 2 c 3 d])
+        assert_equal ['foo', [['a', 0.0]]], r.zmpop('foo')
+        assert_equal ['foo', [['b', 1.0], ['c', 2.0], ['d', 3.0]]], r.zmpop('foo', count: 4)
+
+        r.zadd('foo', %w[0 a 1 b 2 c 3 d])
+        r.zadd('foo2', %w[0 a 1 b 2 c 3 d])
+        assert_equal ['foo', [['d', 3.0]]], r.zmpop('foo', 'foo2', modifier: "MAX")
+      end
+    end
+
     def test_zremrangebylex
       r.zadd('foo', %w[0 a 0 b 0 c 0 d 0 e 0 f 0 g])
       assert_equal 5, r.zremrangebylex('foo', '(b', '[g')


lib/redis/commands/sorted_sets.rb
```diff
@@ -167,6 +167,41 @@ def zpopmin(key, count = nil)
...
+      # @example With count option
+      #   redis.zmpop('zset', count: 2)
+      #   #=> ['zset', [['a', '1'], ['b', '2']]
+      #
+      # @params key [String, Array] a key of the sorted set(s)
+      # @params min_or_max [String] to pop from min or max
+      # @params count [Integer] a number of members
+      #
+      # @return [Array<String, Array<String, Float>>] list of popped elements and scores
+      def zmpop(*keys, min_or_max: "MIN", count: nil)
```
JerrodCarpenter, line 184: Feel free to suggest whatever here.

lib/redis/commands/sorted_sets.rb
```diff
@@ -167,6 +167,41 @@ def zpopmin(key, count = nil)
...
+
+        send_command(args) do |response|
+          response&.map do |entry|
+            case entry
+            when String then entry
+            when Array then entry.map { |pair| FloatifyPairs.call(pair) }.flatten(1)
+            end
+          end
+        end
+      end
```
fatkodima, line 203:
```suggestion
      # Removes and returns up to count members with scores from the first non-empty sorted-set in the provided list of key names.
      #
      # @example Popping a member
      #   redis.zmpop('zset')
      #   #=> ['zset', ['a', 1.0]]
      # @example With count option
      #   redis.zmpop('zset', count: 2)
      #   #=> ['zset', [['a', 1.0], ['b', 2.0]]
      #
      # @params key [String, Array<String>] one or more keys with sorted sets
      # @params modifier [String]
      #  - when `"MIN"` - the elements popped are those with the lowest scores
      #  - when `"MAX"` - the elements popped are those with the highest scores
      # @params count [Integer] a number of members to pop
      #
      # @return [Array<String, Array<String, Float>>] list of popped elements and scores
      def zmpop(*keys, modifier: "MIN", count: 1)
        raise ArgumentError, "Pick either MIN or MAX" unless modifier == "MIN" || modifier == "MAX"

        args = [:zmpop, keys.size, *keys, modifier]

        if count
          args << "COUNT"
          args << Integer(count)
        end

        send_command(args) do |response|
          response&.map do |entry|
            case entry
            when String then entry
            when Array then FloatifyPairs.call(entry)
            end
          end
        end
      end
```

JerrodCarpenter: I included most of these suggestions. Your line 235 does not work:

```ruby
irb(main):005:0> redis.zmpop("myzset", modifier: "MAX", count: 2)
=> ["myzset", [[["three", "3"], ["two", "2"]]]]
```

fatkodima: Thanks. I haven't tested it before suggesting, so I was thinking that it should work 😎

fatkodima: You forgot to change the command description to the suggested (looks more clear imo) and to remove `number_of_keys` variable - `args` is already an array (because of `*args`).

JerrodCarpenter: Updated, appreciate it.

JerrodCarpenter:
Not sure if something's up... but I can't seem to get tests to run locally, they change indefinitely:
```ruby
➜  redis-rb git:(master) ✗ make test
Run options: --seed 4443

# Running:

..................................................................................................................................................................................................................
```

And:
```ruby
➜  redis-rb git:(master) ✗ bundle exec rake test:redis
Run options: --seed 31572

# Running:

...................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................Error running mock server: OpenSSL::SSL::SSLError: SSL_accept returned=1 errno=0 peeraddr=127.0.0.1:61251 state=error: tlsv1 alert unknown ca
/Users/jerrod.carpenter/.asdf/installs/ruby/3.2.0/lib/ruby/3.2.0/openssl/ssl.rb:527:in `accept'
/Users/jerrod.carpenter/.asdf/installs/ruby/3.2.0/lib/ruby/3.2.0/openssl/ssl.rb:527:in `accept'
/Users/jerrod.carpenter/Projects/redis-rb/test/support/redis_mock.rb:46:in `block in run'
/Users/jerrod.carpenter/Projects/redis-rb/test/support/redis_mock.rb:38:in `loop'
/Users/jerrod.carpenter/Projects/redis-rb/test/support/redis_mock.rb:38:in `run'
/Users/jerrod.carpenter/Projects/redis-rb/test/support/redis_mock.rb:30:in `block in start'
..Error running mock server: OpenSSL::SSL::SSLError: SSL_accept returned=1 errno=0 peeraddr=127.0.0.1:61260 state=error: tlsv1 alert unknown ca
/Users/jerrod.carpenter/.asdf/installs/ruby/3.2.0/lib/ruby/3.2.0/openssl/ssl.rb:527:in `accept'
/Users/jerrod.carpenter/.asdf/installs/ruby/3.2.0/lib/ruby/3.2.0/openssl/ssl.rb:527:in `accept'
/Users/jerrod.carpenter/Projects/redis-rb/test/support/redis_mock.rb:46:in `block in run'
/Users/jerrod.carpenter/Projects/redis-rb/test/support/redis_mock.rb:38:in `loop'
/Users/jerrod.carpenter/Projects/redis-rb/test/support/redis_mock.rb:38:in `run'
/Users/jerrod.carpenter/Projects/redis-rb/test/support/redis_mock.rb:30:in `block in start'
................................................
```

casperisfine:
Not too sure what your local error is, but the CI one is legit:

```ruby
  1) Error:
TestDistributedCommandsOnSortedSets#test_zmpop:
NoMethodError: undefined method `zmpop' for #<Redis client v5.0.6 for redis://127.0.0.1:6381/15>
    /home/runner/work/redis-rb/redis-rb/test/lint/sorted_sets.rb:484:in `block in test_zmpop'
    /home/runner/work/redis-rb/redis-rb/test/helper.rb:159:in `target_version'
    /home/runner/work/redis-rb/redis-rb/test/lint/sorted_sets.rb:483:in `test_zmpop'
```

You need to define how this command work on `Redis::Distributed` (in `distributed.rb`).

lib/redis/commands/sorted_sets.rb
```diff
@@ -167,6 +167,42 @@ def zpopmin(key, count = nil)
...
+      # @return [Array<String, Array<String, Float>>] list of popped elements and scores
+      def zmpop(*keys, modifier: "MIN", count: 1)
+        raise ArgumentError, "Pick either MIN or MAX" unless modifier == "MIN" || modifier == "MAX"
+
+        args = [:zmpop, keys.size, *keys, modifier]
+
+        if count
+          args << "COUNT"
+          args << Integer(count)
+        end
```
casperisfine, line 194:
```suggestion
        args << "COUNT" << Integer(count) if count
```

But also, the argument is optional on the redis server side (behave like `1` if not provided), but you default it to 1 here.

I'd day it would be more correct to default it to `nil`.

lib/redis/commands/sorted_sets.rb
```diff
@@ -167,6 +167,42 @@ def zpopmin(key, count = nil)
...
+      #   #=> ['zset', [['a', 1.0], ['b', 2.0]]
+      #
+      # @params key [String, Array<String>] one or more keys with sorted sets
+      # @params modifier [String]
+      #  - when `"MIN"` - the elements popped are those with lowest scores
+      #  - when `"MAX"` - the elements popped are those with the highest scores
+      # @params count [Integer] a number of members to pop
+      #
+      # @return [Array<String, Array<String, Float>>] list of popped elements and scores
+      def zmpop(*keys, modifier: "MIN", count: 1)
```
casperisfine, line 186:
```suggestion
      def zmpop(*keys, modifier: "MIN", count: nil)
```

lib/redis/distributed.rb
```diff
@@ -694,6 +694,11 @@ def zmscore(key, *members)
       node_for(key).zmscore(key, *members)
     end
 
+    # Iterate over keys, removing members from the first non empty sorted set found.
+    def zmpop(*keys, modifier: "MIN", count: nil)
+      keys.find { |key| node_for(key).zmpop(key, modifier: modifier, count: count) }
```
JerrodCarpenter, line 699: I'm not sure if this is right, it seems to make sense? The keys are on different nodes, so we can't really use zmpop otherwise we're assuming they're on the same node right? Would it be better to just raise not implemented?

casperisfine: I'd say do what `zpopmin` and `zpopmax` do, which is to call `ensure_same_node`.

lib/redis/distributed.rb
```diff
@@ -694,6 +694,13 @@ def zmscore(key, *members)
       node_for(key).zmscore(key, *members)
     end
 
+    # Iterate over keys, removing members from the first non empty sorted set found.
+    def zmpop(*keys, modifier: "MIN", count: nil)
+      ensure_same_node(:zmpop, keys) do |node|
+        node.zmpop(key, modifier: modifier, count: count)
```
casperisfine, line 700:
```suggestion
        node.zmpop(*keys, modifier: modifier, count: count)
```

casperisfine (changes requested):
Apologies for the delay I'm traveling right now.

One small error, other than that LGTM.

byroot:
Merged as https://github.com/redis/redis-rb/commit/5904289e8aba1dc1f24a0723763205887ad50957

The failure were because `foo` and `foo2` wouldn't hash to the same node. Fixed by using `{1}foo` and `{1}foo2` instead.

PR #1194, opened 2023-05-27
Remove unneeded requires

fatkodima:
These were left after the logger was removed in version 5.0.

I am curious, what was the reason to remove it?


byroot:
Nice catch. Thank you!

PR #1196, opened 2023-05-30
🚨 Add rubocop-performance linting

JerrodCarpenter:
This may be a tad presumptuous.. but I thought it might be nice to add `rubocop-performance` for the following reasons:
* The gem is marshaling a lot of data and there are risks of us adding some inefficiencies as we add new commands.
* Increasing confidence when reviewing PRs.

Really, I know I may be over stepping here, no worries if I'm way off track.


byroot:
My problem here  is mostly than many cops in rubocop-performance are a bit questionable. Some were added a long time ago based on questionable micro benchmarks or weren't re-considered on more modern rubies.

That said there may be value here, so I'm OK with the general idea, but I'll have to review the cops, and I think I'll disable it for the `test` directory.

Also the correction it did broke CI.

JerrodCarpenter:
> My problem here is mostly than many cops in rubocop-performance are a bit questionable. Some were added a long time ago based on questionable micro benchmarks or weren't re-considered on more modern rubies.

hmmm fair enough

> That said there may be value here, so I'm OK with the general idea, but I'll have to review the cops, and I think I'll disable it for the `test` directory.
> 
> Also the correction it did broke CI.

I'll take a look at CI, might be a little while if that's okay.

Issue #1187, opened 2023-03-20
Adding support for Redis 7.0 commands


SoulPancake:
Support for LMPOP, ZMPOP and other Redis 7.0 commands.
We can create a new tracker for those here if that's cool

mkdynamic:
Would also be interested in the sharded pub/sub commands in Redis 7.0, SSUBSCRIBE, SPUBLISH and SUBSCRIBE:
https://redis.io/docs/manual/pubsub/#sharded-pubsub

JerrodCarpenter:
I plan on adding some of these one by one.. Here's a start: https://github.com/redis/redis-rb/pull/1189

Goltergaul:
any plans to add exiry commands related to hashes like for example HPEXPIRE that come with 7.4?

byroot:
As you can see in all linked PRs, commands are added whenever someone who needs it makes a PR for it.

Adding support for a command is pretty trivial so you should try your hand at it.

Combos93:
Hello, guys.
Yesterday i started to implement HEXPIRE, HTTL, HPTTL commands. If you don't mind

vladvildanov:
Hey, folks! After closing #1384 #1385 I think it's all in terms of 7.0 compatibility, please re-open if something missed

PR #1201, opened 2023-06-10
✨ Add SINTERCARD

JerrodCarpenter:
* Adding [SINTERCARD](https://redis.io/commands/sintercard/)
* https://github.com/redis/redis-rb/issues/1187

```ruby
=> #<Redis client v5.0.6 for redis://localhost:6379/2>
irb(main):002:0> redis.sadd("key1", "a")
=> 1
irb(main):003:0> redis.sadd("key1", "b", "c", "d", "e")
=> 4
irb(main):004:0> redis.sadd("key2", "c", "d", "f")
=> 3
irb(main):005:0> redis.sinter("key1", "key2")
=> ["c", "d"]
irb(main):006:0> redis.sintercard("key1", "key2")
=> 2
irb(main):007:0> redis.sintercard("key1", "key2", limit: 1)
=> 1
```


lib/redis/distributed.rb
```diff
@@ -648,6 +648,15 @@ def sinterstore(destination, *keys)
       end
     end
 
+    # This command is similar to SINTER, but instead of returning the result set,
+    #   it returns just the cardinality of the result.
+    def sintercard(*keys, limit: nil)
+      keys.flatten!(1)
```
byroot, line 654: If you support both `sintercard("key1", "key2")` and `sintercard(["key1", "key2"])` that should be tested.

byroot (changes requested):

lib/redis/commands/sets.rb
```diff
@@ -154,6 +154,19 @@ def sinterstore(destination, *keys)
         send_command([:sinterstore, destination].concat(keys))
       end
 
+      # This command is similar to SINTER, but instead of returning the result set,
+      #   it returns just the cardinality of the result.
+      #
+      # @param [String, Array<String>] keys keys pointing to sets to intersect
+      # @param [Interger] limits number of matches searched for
+      # @return [Array<String>] cardinality of the intersection
```
roharon, line 162: Isn't it return Integer type?
Seems to need changing it
```suggestion
      # @return [Integer] cardinality of the intersection
```

PR #1205, opened 2023-07-05
Replace docs on Inch CI with ones on rubydoc.info

tnir:
Inch CI has not been in service for many years and so `https://inch-ci.org/github/redis/redis-rb` is not available. https://rubydoc.info/gems/redis (published version) and https://rubydoc.info/github/redis/redis-rb (master HEAD) are now available.


byroot:
Thanks.

PR #1217, opened 2023-08-16
Support role being specified as a string

dhruvCW:
sometimes the redis options are provided as a hash with string values.

This is true in my case where the configurations are read as a json object from an environment variable. 

 Given that only the role is expected to be a symbol, It is a much better user experience if the conversion was handled internally by the library instead of requiring the user to handle it themselves.

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -152,6 +152,9 @@ class Redis
         end
       end
 
+      # Role is expected to be a symbol
+      options[:role] = options[:role].to_sym if options.key?(:role)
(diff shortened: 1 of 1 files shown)


byroot:
I could be open to this, but this would need to be done in `redis-client`, not here.

dhruvCW:
>I could be open to this, but this would need to be done in redis-client, not here.

awesome, wasn't really sure if that was something you might be open to or not. Will open a PR there 👍

PR #1222, opened 2023-09-13
Fix redis-clustering gem to pass the test with latest dependencies

supercaracal:
* Since redis-cluster-client changed a portion of internal implementation in `0.6.1`, redis-rb should adapt codes depended on it.
* A portion of test cases is a bit unstable against redis server minor changes, so we should keep it simple.

### DIFF
--- a/cluster/lib/redis/cluster/client.rb
+++ b/cluster/lib/redis/cluster/client.rb
@@ -28,11 +28,11 @@ class Redis
       ruby2_keywords :initialize if respond_to?(:ruby2_keywords, true)
 
       def id
-        @router.node.node_keys.join(' ')
+        @router.node_keys.join(' ')
       end
 
       def server_url
-        @router.node.node_keys
+        @router.node_keys
       end
 
       def connected?
--- a/cluster/redis-clustering.gemspec
+++ b/cluster/redis-clustering.gemspec
@@ -47,5 +47,5 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.7.0'
 
   s.add_runtime_dependency('redis', s.version)
-  s.add_runtime_dependency('redis-cluster-client', '>= 0.3.7')
+  s.add_runtime_dependency('redis-cluster-client', '>= 0.6.1')
 end
--- a/cluster/test/commands_on_keys_test.rb
+++ b/cluster/test/commands_on_keys_test.rb
@@ -47,7 +47,6 @@ class TestClusterCommandsOnKeys < Minitest::Test
   def test_object
     redis.lpush('mylist', 'Hello World')
     assert_equal 1, redis.object('refcount', 'mylist')
-    assert_equal 'quicklist', redis.object('encoding', 'mylist')
     assert(redis.object('idletime', 'mylist') >= 0)
 
     redis.set('foo', 1000)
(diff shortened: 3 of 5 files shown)


cluster/lib/redis/cluster/client.rb
```diff
@@ -28,11 +28,11 @@ def initialize(*)
       ruby2_keywords :initialize if respond_to?(:ruby2_keywords, true)
 
       def id
-        @router.node.node_keys.join(' ')
+        @router.node_keys.join(' ')
```
supercaracal, line 31: https://github.com/redis-rb/redis-cluster-client/blob/9b5fc6644a9cffab169bca517552120519bad7e4/lib/redis_client/cluster/router.rb#L208-L210

cluster/redis-clustering.gemspec
```diff
@@ -47,5 +47,5 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.7.0'
 
   s.add_runtime_dependency('redis', s.version)
-  s.add_runtime_dependency('redis-cluster-client', '>= 0.3.7')
+  s.add_runtime_dependency('redis-cluster-client', '>= 0.6.1')
```
supercaracal, line 50: https://github.com/redis-rb/redis-cluster-client/releases

cluster/test/commands_on_keys_test.rb
```diff
@@ -47,7 +47,7 @@ def test_migrate
   def test_object
     redis.lpush('mylist', 'Hello World')
     assert_equal 1, redis.object('refcount', 'mylist')
-    assert_equal 'quicklist', redis.object('encoding', 'mylist')
+    assert_equal 'listpack', redis.object('encoding', 'mylist')
```
supercaracal, line 50: It looks like internal changes for Redis 7.2.

byroot: Yes, it's best to either remove that test or to allow both values.

There is no point testing Redis internal implementation.

cluster/test/commands_on_server_test.rb
```diff
@@ -41,12 +41,8 @@ def test_client_kill
...
   def test_client_list
     a_client_info = redis.client(:list).first
-    actual = a_client_info.keys.sort
-    expected = %w[addr age cmd db events fd flags id idle multi name obl oll omem psub qbuf qbuf-free sub]
-    expected << 'user' << 'argv-mem' << 'tot-mem' if version >= '6'
-    expected << 'laddr' << 'redir' if version >= '6.2'
-    expected << "multi-mem" << "rbp" << "rbs" << "resp" << "ssub" if version >= '7.0'
-    assert_equal expected.sort, actual.sort
+    assert_instance_of Hash, a_client_info
+    assert_includes a_client_info, 'addr'
```
supercaracal, line 45: * It looks like server side changes for reply type from Array to Hash.
* I simplified this test case rather than meticulous assertion because it is weak against server side minor changes.

cluster/test/commands_on_keys_test.rb
```diff
@@ -47,7 +47,6 @@ def test_migrate
   def test_object
     redis.lpush('mylist', 'Hello World')
     assert_equal 1, redis.object('refcount', 'mylist')
-    assert_equal 'quicklist', redis.object('encoding', 'mylist')
```
supercaracal, line 50: It was changed to 'listpack' in Redis 7.2. We have no advantages for testing redis minor behavior, so just removed.

PR #1225, opened 2023-09-29
Update default timeout docs

afinzel:
Update the documentation to reflect the timeout is now 1 second.  This pr, https://github.com/redis-rb/redis-client/commit/90d219fe59dc124ed9157c353e4e9d0c75b0eb67#diff-dd62159b97639b7e6aca9f5c209b2ff5a74924bf38699581dcdc5074e2e62bbfL7 set it to 1 second.  It is also documented in the change-logs as being set to 1 second for version 5


afinzel:
Thank you for merging so fast.

PR #1236, opened 2023-11-15
Pass watch: kwarg to #multi calls on clients

KJTsanaktsidis:
Both RedisClient and RedisClient::Cluster have #multi methods that accept a watch: kwarg to watch keys as part of the transaction. See:

* https://github.com/redis-rb/redis-client/blob/ba8848b195404ab337daa94647f600b1a4506f0f/lib/redis_client.rb#L419
* https://github.com/redis-rb/redis-cluster-client/blob/c00ee01d087471b690f44ecbac8d0cea005783fd/lib/redis_client/cluster/transaction.rb#L36

However, the subclasses `Redis::Client` and `Redis::Cluster::Client` in this gem(s) don't forward these keyword arguments, so you can't take advantage of them.

This PR simply delegates the `watch` argument.

### DIFF
--- a/cluster/lib/redis/cluster/client.rb
+++ b/cluster/lib/redis/cluster/client.rb
@@ -70,8 +70,8 @@ class Redis
         handle_errors { super(&block) }
       end
 
-      def multi(&block)
-        handle_errors { super(&block) }
+      def multi(watch: nil, &block)
+        handle_errors { super(watch: watch, &block) }
       end
 
       private
--- a/lib/redis/client.rb
+++ b/lib/redis/client.rb
@@ -111,7 +111,7 @@ class Redis
       Client.translate_error!(error)
     end
 
-    def multi
+    def multi(watch: nil)
       super
     rescue ::RedisClient::Error => error
       Client.translate_error!(error)


byroot:
Yeah, I guess it's a nice shortcut.

KJTsanaktsidis:
That's got to be the fastest merge I've ever had! thanks.

> Yeah, I guess it's a nice shortcut.

It's actually a little more important than a shortcut for cluster mode - the watch/unwatch machinery in `Redis::Commands::Transactions` doesn't work properly with redis-cluster-client because it doesn't store transaction-related state (like what slot we have to ensure everything is in) outside of a call to `#multi`. So this pattern doesn't work.

```
$redis.watch('{hash}key1', '{hash}key2') do
    $redis.get('{hash}key1')
    # oops, don't like what I read out of key1, let's not do the transaction after all.
    $redis.unwatch
end
``` 

Using the `multi(watch: ...)` pattern does work though (once https://github.com/redis-rb/redis-cluster-client/pull/294 gets merged).

I do want to make `Redis::Cluster#watch do ... end` work properly though if possible, I think that needs a little more surgery though so starting with this seemed easier.

PR #1238, opened 2023-11-17
Cluster: Preserve message and suberrors when raising CommandErrorCollection

KJTsanaktsidis:
The constructor signature for CommandErrorCollection is (errors, message). That means if you call "raise CommandErrorCollection, 'foobar'", that actually winds up setting the _errors_ field to 'foobar', _NOT_ the message. This leads to incredibly undescriptive "Command errors were replied on any node" errors getting emitted if there are errors using pipelining.

Fix this by:
1) explicitly constructing the error, with the arguments in the right order
2) mapping the sub-errors into redis-rb errors too - presumably there's an :errors attr_reader because somebody will find that useful.

### DIFF
--- a/cluster/lib/redis/cluster/client.rb
+++ b/cluster/lib/redis/cluster/client.rb
@@ -84,7 +84,14 @@ class Redis
             raise ERROR_MAPPING.fetch(node_error.class), node_error.message, node_error.backtrace
           end
         end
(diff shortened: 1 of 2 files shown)


byroot:
I'll wait for @supercaracal 's opinion.

supercaracal (approved):
Sorry for my late reply. I agree with this fix.

byroot:
Not late at all. Thank you for maintaining clustering.

PR #1242, opened 2023-12-09
Support `BIT/BYTE` for `bitcount` and `bitpos` for redis 7+

fatkodima:
Reference https://github.com/redis/redis-rb/issues/1187

Redis 7.0 release notes - https://raw.githubusercontent.com/redis/redis/7.0/00-RELEASENOTES

I am not very happy with the `bytes_range` parameter name. Maybe you can suggest something better? Maybe `use_bytes` or `offset_unit` etc...

### DIFF
--- a/lib/redis/commands/bitmaps.rb
+++ b/lib/redis/commands/bitmaps.rb
@@ -27,9 +27,13 @@ class Redis
       # @param [String] key
       # @param [Integer] start start index
       # @param [Integer] stop stop index
+      # @param [String, Symbol] scale the scale of the offset range
+      #     e.g. 'BYTE' - interpreted as a range of bytes, 'BIT' - interpreted as a range of bits
       # @return [Integer] the number of bits set to 1
-      def bitcount(key, start = 0, stop = -1)
-        send_command([:bitcount, key, start, stop])
+      def bitcount(key, start = 0, stop = -1, scale: nil)
+        command = [:bitcount, key, start, stop]
+        command << scale if scale
+        send_command(command)
       end
 
       # Perform a bitwise operation between strings and store the resulting string in a key.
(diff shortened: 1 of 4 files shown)


byroot:
What about: `scale: :bit` or `index: :byte`?

fatkodima:
👍 for `:scale`.

lib/redis/commands/bitmaps.rb
```diff
@@ -27,9 +27,13 @@ def getbit(key, offset)
...
       # @param [Integer] start start index
       # @param [Integer] stop stop index
+      # @param [String, Symbol] scale the scale of the offset range
+      #     e.g. 'BYTE' - interpreted as a range of bytes, 'BIT' - interpreted as a range of bites
       # @return [Integer] the number of bits set to 1
-      def bitcount(key, start = 0, stop = -1)
-        send_command([:bitcount, key, start, stop])
+      def bitcount(key, start = 0, stop = -1, scale: 'BYTE')
+        command = [:bitcount, key, start, stop]
+        command << 'BIT' if scale.to_s.upcase == 'BIT'
```
byroot, line 35:
```suggestion
        command << 'BIT' if scale.
```
I think we should append it unless it's not set. Because here if I pass `scale: "lol"`, it will silently do nothing.

With `scale: nil` in the signature.

fatkodima: Agreed, thanks.

lib/redis/commands/bitmaps.rb
```diff
@@ -27,9 +27,13 @@ def getbit(key, offset)
       # @param [String] key
       # @param [Integer] start start index
       # @param [Integer] stop stop index
+      # @param [String, Symbol] scale the scale of the offset range
+      #     e.g. 'BYTE' - interpreted as a range of bytes, 'BIT' - interpreted as a range of bites
```
byroot, line 31:
```suggestion
      #     e.g. 'BYTE' - interpreted as a range of bytes, 'BIT' - interpreted as a range of bits
```

lib/redis/commands/bitmaps.rb
```diff
@@ -27,9 +27,13 @@ def getbit(key, offset)
...
       # @return [Integer] the number of bits set to 1
-      def bitcount(key, start = 0, stop = -1)
-        send_command([:bitcount, key, start, stop])
+      def bitcount(key, start = 0, stop = -1, scale: nil)
+        command = [:bitcount, key, start, stop]
+        command << scale.to_s.upcase if scale
```
byroot, line 35: I wouldn't even bother with the `to_s.upcase`, we can directly pass it to redis in lower case, it's no problem.

PR #1243, opened 2023-12-16
Update README to remove mutex language

awiddersheim:
This change removes references to their being a `mutex` protecting the underlying Redis connection from concurrent access. This was removed in `5.x`[1].

[1]: https://github.com/redis/redis-rb/issues/1070

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,10 +74,9 @@ available on [RubyDoc.info][rubydoc].
 ## Connection Pooling and Thread safety
 
 The client does not provide connection pooling. Each `Redis` instance
-has one and only one connection to the server, and use of this connection
-is protected by a mutex.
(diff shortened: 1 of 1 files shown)


awiddersheim:
Seems like based on the plans laid out in https://github.com/redis/redis-rb/issues/1070, folks should be gravitating toward using `redis-client` instead of this gem in the future. That still fair to say?

byroot:
I never had any plan to remove the mutex from `redis-rb`, it would break way too many apps out there. It's still there: https://github.com/redis/redis-rb/blob/935f64f18d7c37386e9b01db71683f8f3b868cd2/lib/redis.rb#L146-L148

awiddersheim:
Thanks @byroot.

PR #1244, opened 2024-01-08
Handle unknown subclasses of RedisClient::Error in clustering

KJTsanaktsidis:
The Redis::Client code already knows how to crawl up through exception ancestors and match that against the hash; call into that code from Redis::Cluster::Client as well. `

Otherwise, you can get KeyError raised from e.g. pipeline blocks instead of the real error that happened.


KJTsanaktsidis:
I don't think I'm responsible for the rubygems problems or the truffleruby crash :/ so I think this should be OK to look at? 🤷 

https://github.com/rubygems/rubygems/issues/6197 - I wonder if we need to specify a desired version of rubygems to install explicitly if we want to keep running rubocop on 2.6

supercaracal (approved):

byroot:
Yeah, we can figure out the rubocop issue later.

PR #1251, opened 2024-01-30
Add PEXPIRETIME

wholien:
- add [PEXPIRETIME](https://redis.io/commands/pexpiretime/)
- https://github.com/redis/redis-rb/issues/1187

CleanShot 2024-01-30 at 13 34 01@2x


wholien:
@byroot quick question: when will a new release of this gem be available? I'd like to be able to use the PEXPIRETIME / EXPIRETIME commands that were added. Thanks!

byroot:
There's no specific schedule, I release when there is an important patch or enough change to justify taking the time, in the meantime users can point their Gemfile at the repo if there is really something they direly need.

There was a couple features on master and the last release was a few months ago, so I've cut a 5.1.0.

PR #1262, opened 2024-04-08
Fixed small README issues, replace single with double quotes

OlegChuev:
- Fixed small README issues
- Replaced single quotes (') with double quotes (") to follow the same approach everywhere in the docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ See [RubyDoc.info][rubydoc] for the API docs of the latest published gem.
 
 Install with:
 
-```
-$ gem install redis
+```bash
+gem install redis
 ```
 
 You can connect to Redis by instantiating the `Redis` class:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,8 +8,8 @@ See [RubyDoc.info][rubydoc] for the API docs of the latest published gem.
 
 Install with:
 
-```
-$ gem install redis
```
byroot, line 12: This is what Bundelr generate when running `bundle gem`, I don't see why we'd remove it.

OlegChuev: @byroot 

Whenever you copy a command directly from GitHub and paste it into the terminal, it's terribly annoying to have to delete the `$` sign each time. Personally, I find the dollar sign useless here, as it only wastes time to remove it and doesn't represent anything

Screenshot 2024-04-08 at 17 21 22


Screenshot 2024-04-08 at 17 23 13

byroot:
Thanks for the PR, but I don't see much value in these changes.

PR #1263, opened 2024-04-10
add flag to disable raising exceptions with pipelined

philippeboyd:
Linked to https://github.com/redis-rb/redis-client/pull/187
Depends on https://github.com/redis-rb/redis-cluster-client/pull/343

Modifying the pipelined method in lib/redis_client.rb to include an optional exception parameter, which determines whether an exception should be raised when an error occurs during pipelining.

This gives better flexibility for error management when we want to retry only failed commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,28 @@ end
 # => ["OK"]
 ```
 
+### Exception management
+
+The `exception` flag in the `#pipelined` is a feature that modifies the pipeline execution behavior. When set
+to `false`, it doesn't raise an exception when a command error occurs. Instead, it allows the pipeline to execute all
+commands, and any failed command will be available in the returned array. (Defaults to `true`)
+
+```ruby
+results = redis.pipelined(exception: false) do |pipeline|
+  pipeline.set('key1', 'value1')
+  pipeline.lpush('key1', 'something') # This will fail
+  pipeline.set('key2', 'value2')
+end
+# results => ["OK", #<RedisClient::WrongTypeError: WRONGTYPE Operation against a key holding the wrong kind of value>, "OK"]
+
+results.each do |result|
+  if result.is_a?(Redis::CommandError)
+    # Do something with the failed result
+  end
+end
+```
+
+
 ### Executing commands atomically
 
 You can use `MULTI/EXEC` to run a number of commands in an atomic
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add `exception` flag in `pipelined` allowing failed commands to be returned in the result array when set to `false`.
+
 # 5.1.0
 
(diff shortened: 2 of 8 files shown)


Gemfile
```diff
@@ -10,3 +10,6 @@ gem 'rubocop', '~> 1.25.1'
 gem 'mocha'
 
 gem 'hiredis-client'
+
+# TODO remove before merging
+gem 'redis-client', git: 'https://github.com/redis-rb/redis-client.git', branch: 'master'
```
casperisfine, line 15: Note; we'll need to bump the `redis-client` version requirement in `gemspec`.

README.md
```diff
@@ -191,6 +191,28 @@ end
...
+```ruby
+results = redis.pipelined(exception: false) do |pipeline|
+  pipeline.set('key1', 'value1')
+  pipeline.lpush('key1', 'something') # This will fail
+  pipeline.set('key2', 'value2')
+end
+# results => ["OK", #<RedisClient::WrongTypeError: WRONGTYPE Operation against a key holding the wrong kind of value>, "OK"]
+
+results.each do |result|
+  if result.is_a?(RedisClient::CommandError)
```
casperisfine, line 209:
```suggestion
  if result.is_a?(Redis::CommandError)
```

cluster/redis-clustering.gemspec
```diff
@@ -47,5 +47,5 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.7.0'
 
   s.add_runtime_dependency('redis', s.version)
-  s.add_runtime_dependency('redis-cluster-client', '>= 0.7.0')
+  s.add_runtime_dependency('redis-cluster-client', '>= 0.7.0', "< 0.7.10")
```
byroot, line 50: Why `< 0.7.10` ?

If we need a new release of it, it can wait. cc @supercaracal

supercaracal: `0.8.0` has released.

* https://github.com/redis-rb/redis-cluster-client/releases/tag/v0.8.0
* https://rubygems.org/gems/redis-cluster-client/versions/0.8.0

supercaracal:
Several errors and failures in test cases of the cluster feature will be fixed with the following PR.

* #1255

Although there are some discussion about the transaction feature for the cluster client, the issue is a matter on the redis-cluster-client side. So redis-rb can ignore the discussion at now.

cluster/redis-clustering.gemspec
supercaracal: `0.7.11` was included some changes for the transaction feature. We can get rid of the restriction if the following PR is merged.

* #1255

philippeboyd: @byroot yeah sorry I was waiting for that specific fix for the watch command, hence the pinning.
I just rebased my branch, and PR is updated.

byroot (approved):

PR #1275, opened 2024-05-02
Fix a failure of a test case for a cluster client

supercaracal:
The version `0.9.0` of redis-cluster-client was released. It made `MSET`, `MGET` and `DEL` commands able to pass multiple keys without a hashtag by using pipelining. Although the performance is degradated, I think it may be useful in some use cases. So, because the behavior was changed, This pull request fixes a failure of a test case for a cluster client.

### DIFF
--- a/cluster/test/commands_on_keys_test.rb
+++ b/cluster/test/commands_on_keys_test.rb
@@ -18,10 +18,7 @@ class TestClusterCommandsOnKeys < Minitest::Test
   def test_del
     set_some_keys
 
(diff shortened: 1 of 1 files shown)


supercaracal:
```
  1) Failure:
TestClusterClientReplicas#test_client_can_command_with_replica [cluster/test/client_replicas_test.rb:19]:
--- expected
+++ actual
@@ -1,3 +1 @@
-# encoding: US-ASCII
-#    valid: true
-"1"
+nil
```

The above failure isn't related this pull request. It may be flaky. Please ignore it on this pull request. I'll look into and work on it if it occurs frequently.

PR #1283, opened 2024-06-03
Fix a memoization bug for cluster client

### LINKED ISSUE #1282: redis-clustering: NoMethodError: undefined method `node_keys' for nil:NilClass
With [a recent redis-cluster-client commit](https://github.com/redis-rb/redis-cluster-client/commit/c1b7dd7cd135aad86355c817f8c0a1d3bdb5ee07), `@router` can be `nil`, so https://github.com/redis/redis-rb/blob/v5.2.0/cluster/lib/redis/cluster/client.rb#L56 is failing. 

```
#<NoMethodError: undefined method `node_keys' for nil:NilClass

        @router.node_keys.join(' ')
               ^^^^^^^^^^>
/var/web/canvas/shared/bundle/ruby/3.1.0/gems/redis-clustering-5.2.0/lib/redis/cluster/client.rb:56:in `id'
/var/web/canvas/shared/bundle/ruby/3.1.0/gems/redis-5.2.0/lib/redis.rb:111:in `id'
/var/web/canvas/shared/bundle/ruby/3.1.0/gems/redis-5.2.0/lib/redis.rb:115:in `inspect'
(irb):8:in `<top (required)>'
```
is a full backtrace.

Note that in [redis-cluster-client's inspect method](https://github.com/redis-rb/redis-cluster-client/commit/c1b7dd7cd135aad86355c817f8c0a1d3bdb5ee07#diff-737532d46e36c72be00bbbbce4de81b98dc9cdd82466bd1a0b6058b9c77f8da0R28), it checks if `@router` is `nil` before using it now.

supercaracal:
Closes #1282

I fixed redis-cluster-client to be able to delay the initialization to the first query.

* https://github.com/redis-rb/redis-cluster-client/pull/364
* https://github.com/redis-rb/redis-cluster-client/releases/tag/v0.10.0

But I've forgotten to fix it on redis-clustering side. So this pull request fixes it.

### DIFF
--- a/cluster/lib/redis/cluster/client.rb
+++ b/cluster/lib/redis/cluster/client.rb
@@ -54,11 +54,11 @@ class Redis
       ruby2_keywords :initialize if respond_to?(:ruby2_keywords, true)
 
       def id
-        @router.node_keys.join(' ')
+        server_url.join(' ')
       end
 
       def server_url
-        @router.node_keys
+        @router.nil? ? @config.startup_nodes.keys : router.node_keys
       end
 
       def connected?
@@ -115,9 +115,9 @@ class Redis
         end
 
         handle_errors do
-          RedisClient::Cluster::OptimisticLocking.new(@router).watch(keys) do |c, slot, asking|
+          RedisClient::Cluster::OptimisticLocking.new(router).watch(keys) do |c, slot, asking|
             transaction = Redis::Cluster::TransactionAdapter.new(
-              self, @router, @command_builder, node: c, slot: slot, asking: asking
+              self, router, @command_builder, node: c, slot: slot, asking: asking
             )
 
             result = yield transaction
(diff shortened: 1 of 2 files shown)


cluster/lib/redis/cluster/client.rb
```diff
@@ -54,11 +54,11 @@ def initialize(*)
       ruby2_keywords :initialize if respond_to?(:ruby2_keywords, true)
 
       def id
-        @router.node_keys.join(' ')
+        server_url.join(' ')
       end
 
       def server_url
-        @router.node_keys
+        @router.nil? ? @config.startup_nodes.keys : router.node_keys
```
corrupt952, line 61: Hi! 👋

This is related to a question I asked in https://github.com/redis-rb/redis-cluster-client/pull/364/files#r1724410379 . 

I noticed that here too, the `@router` variable is being used directly instead of calling the `router` method. What's the reasoning behind this approach?

It seems that calling `router` would eliminate the need for the nil check. Is there a specific advantage to using `@router` directly in this context?

Also, I am waiting for this fix to be released!

Thanks in advance for your help!

supercaracal: I thought that it would be better to delay initialization if any commands aren't sent to servers in the method.

PR #1288, opened 2024-08-17
Support Redis.default_configuration = { ... }

sebyx07:
problem: I have a rails app + sidekiq + anycable + rails cache. And i want all of them to use redis with hiredis + a different db

solution:

```ruby
redis_uri = URI.parse(ENV["REDIS_URL"])

Redis.default_configuration = { 
  host: redis_uri.host, port: port = redis_uri.port, driver: :hiredis
}

# for example sidekiq

Sidekiq.configure_server do |c|
  c.redis = { db: 0 }
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,16 @@ server or a different port, try:
 
 ```ruby
 redis = Redis.new(host: "10.0.1.1", port: 6380, db: 15)
+
+# or you can globally configure the connection information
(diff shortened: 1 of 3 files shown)


byroot:
I feel this is just as easily done with a simple hash merge in your app. Not really convinced of the usefulness of this.

PR #1289, opened 2024-08-23
Forward blocks on Redis#call

mlarraz:
This matches the behavior of RedisClient#call


byroot:
Hum, why not, but maybe a test would be good.

byroot:
I tried to squash your commits and GitHub didn't like it, but it's effectively merged: https://github.com/redis/redis-rb/commit/a6759452a897911d69ca8b05645cf7a31ceaffac

PR #1300, opened 2024-12-09
Add :with_score option to zrank and zrevrank

Hugo-Hache:
[Option](https://redis.io/docs/latest/commands/zrank) was introduced with Redis 7.2.0

### DIFF
--- a/lib/redis/commands.rb
+++ b/lib/redis/commands.rb
@@ -83,12 +83,14 @@ class Redis
       end
     }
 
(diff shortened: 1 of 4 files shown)


byroot:
Looks good, it's just missing the server version checks in the test so it doesn't fail CI against older servers.

See the `target_version` method in that same test file.

Hugo-Hache:
Found it thanks. Should be fixed.

PR #1309, opened 2025-02-18
Support the MINID option for XADD

### LINKED ISSUE #1308: Support MINID for XADD
Happy to send an MR for this, but would ideally like to get agreement on the API first.

XTRIM support for MINID was added as a `strategy` option, with the value argument renamed to `id_or_len` - but it's a positional argument so no-one should have noticed.

Currently `xadd` has a *named* `maxlen` argument, and I don't really want to apply the same `strategy`/`id_or_len` change because that would be a breaking change. 

So is it OK if I add a new `minid` or `min_id` argument to `xadd`?

willbryant:
Closes #1308.

### DIFF
--- a/lib/redis/commands/streams.rb
+++ b/lib/redis/commands/streams.rb
@@ -41,18 +41,25 @@ class Redis
       # @param opts  [Hash]   several options for `XADD` command
       #
       # @option opts [String]  :id          the entry id, default value is `*`, it means auto generation
-      # @option opts [Integer] :maxlen      max length of entries
-      # @option opts [Boolean] :approximate whether to add `~` modifier of maxlen or not
+      # @option opts [Integer] :maxlen      max length of entries to keep
+      # @option opts [Integer] :minid       min id of entries to keep
+      # @option opts [Boolean] :approximate whether to add `~` modifier of maxlen/minid or not
       # @option opts [Boolean] :nomkstream  whether to add NOMKSTREAM, default is not to add
       #
       # @return [String] the entry id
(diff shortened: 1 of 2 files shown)


lib/redis/commands/streams.rb
```diff
@@ -41,18 +41,25 @@ def xinfo(subcommand, key, group = nil)
...
       # @option opts [Boolean] :nomkstream  whether to add NOMKSTREAM, default is not to add
       #
       # @return [String] the entry id
-      def xadd(key, entry, approximate: nil, maxlen: nil, nomkstream: nil, id: '*')
+      def xadd(key, entry, approximate: nil, maxlen: nil, minid: nil, nomkstream: nil, id: '*')
         args = [:xadd, key]
         args << 'NOMKSTREAM' if nomkstream
         if maxlen
+          raise ArgumentError, "can't supply both maxlen and minid" if minid
+
```
byroot, line 55: I had to insert this blank line to appease rubocop.

test/lint/streams.rb
```diff
@@ -89,6 +89,16 @@ def test_xadd_with_maxlen_and_approximate_option
       assert_match ENTRY_ID_FORMAT, actual
     end
 
+    def test_xadd_with_minid_and_approximate_option
+      omit_version('6.2.0')
```
byroot, line 93: With newer commands and argument, we have to specify which version introduced them.

lib/redis/commands/streams.rb
willbryant: My old nemesis.

PR #1312, opened 2025-05-14
Fix `without_reconnect` to wrap redis-client exceptions

byroot:


ChrisBr (approved):
Thank you!

stanhu:
@byroot I think this introduced an unintended regression in ActionCable: https://github.com/rails/rails/issues/55919.

stanhu:
Oh, this was already fixed in https://github.com/rails/rails/pull/55359. Thanks!

PR #1317, opened 2025-07-21
Test Cluster against multiple Ruby versions

marcogregorius:
The `redis-clustering` gem currently only runs on Ruby 2.7 on CI.

This adds a matrix of Ruby versions from 2.7 up to 3.4.


byroot:
This was done on purpose because the clustering test suite was quite flaky at the time.

Seems like it's better now.

PR #1324, opened 2025-10-30
Add ability to set TTL for hash fields, `HEXPIRE`

vietqhoang:
Issue: #1187 

Introduced in 7.4.0, https://redis.io/docs/latest/commands/hexpire/

### DIFF
--- a/test/lint/hashes.rb
+++ b/test/lint/hashes.rb
@@ -227,5 +227,28 @@ module Lint
       expected = ['0', [%w[f1 Jack], %w[f2 33]]]
       assert_equal expected, redis.hscan('foo', 0)
     end
+
+    def test_hexpire
+      target_version "7.4.0" do
+        r.hset("foo", "f1", "v2")
+
+        assert_equal [1], r.hexpire("foo", 4, "f1")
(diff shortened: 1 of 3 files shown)


test/lint/hashes.rb
```diff
@@ -227,5 +227,18 @@ def test_hscan
...
     end
+
+    def test_hexpire
+      target_version "7.4.0" do
+        r.hset("foo", "f1", "v2")
+
+        assert_equal [1], r.hexpire("foo", 1, "f1")
+        assert r.hexists("foo", "f1")
+
+        sleep(1)
```
byroot: Sleeping in test suite is always a bad idea.

A better way to test `hexpire` is to implement `httl`, and use it to check the TTL changed.

vietqhoang:
Sorry, didn't catch you were also editing.

Just want to note the change added to distributed

https://github.com/vietqhoang/redis-rb/blob/2c4eaac1c1ef07a6b33bfe2ac4d1945912b09c8c/lib/redis/distributed.rb#L925-L927

The `ttl` argument is unnecessary since `httl` doesn't intake `ttl`. Tests still passed...

PR #1346, opened 2026-06-04
Added initial support for JSON module (JSON.SET/GET)

vladvildanov:
This PR introduces support for Redis modules, starting with RedisJSON. Beyond the two commands themselves, the goal is to lay down the reusable pattern every future module command will follow: where module command code lives, how it's namespaced, how replies are kept protocol-safe, and how the module test suite is wired into CI.

  **Test infrastructure**
  - New decoupled test:modules group (test/modules/, shared test/lint/json.rb), folded into the default rake test aggregate after the standalone/distributed/sentinel suites.
  - New Helper::Modules client + omit_unless_module("ReJSON") helper (checks MODULE LIST, not the version number).
  - Module tests run on a configurable port (REDIS_MODULES_PORT, default 6381).
  
   **Docker / CI — version-aware module testing**
  - Module tests run inside the existing test matrix (Ruby × Redis version × driver), not a separate job.
  - On Redis ≥ 8.0, modules are in core → tests run against the regular standalone instance; no extra infrastructure.
  - On Redis 7.2.x / 7.4.x, a dedicated Redis Stack service (compose modules profile, port 6383) is booted with the version-matched image — 7.2 → rs-7.2.0-v20, 7.4 → rs-7.4.0-v8 — and the module client points
  at it.
  - makefile gains start_modules / stop_modules shims.

---

> [!NOTE]
> **Low Risk**
> New optional API surface and test/docker wiring; no changes to core Redis command paths or auth.
> 
> **Overview**
> Adds **RedisJSON** client support via `json_set` and `json_get` under `Redis::Commands::Modules::Json`, wired into `Redis` and `Redis::Distributed`. Values are encoded with `JSON.generate` on write and parsed on read, with `raw:`, `nx:`, and `xx:` options; conditional sets use the existing `BoolifySet` reply shaping.
> 
> Introduces a **`test:modules`** suite (`test/modules/`, shared `Lint::Json`), `Helper::Modules` / `Helper::DistributedModules`, and `require_module("ReJSON")` so missing module support fails loudly. CI runs module tests in the existing matrix: Redis **≥ 8** hits standalone on **6381**; **7.2/7.4** boot a Redis Stack service on **6383** (`docker compose` `modules` profile, version-matched `REDIS_STACK_VERSION`). Rake default `test` and GitHub Actions now include `test:modules`; the makefile adds `start_modules` / `stop_modules`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7cf316ea1ac5a2aeed5126116933dd15b4dd4414. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -47,6 +47,12 @@ jobs:
           - { ruby_distribution: TruffleRuby, ruby_version: "4.0" }
     env:
       REDIS_VERSION: ${{ matrix.redis_version }}
+      # RedisJSON et al. ship in Redis core only from 8.0. On the 7.2/7.4 rows the module tests
+      # run against a Redis Stack instance (port 6383) booted from the version-matched rs- tag;
+      # on >= 8 REDIS_STACK_VERSION is empty (no Stack booted) and the module tests run against
+      # the standalone instance (port 6381).
+      REDIS_STACK_VERSION: ${{ startsWith(matrix.redis_version, '7.2') && 'rs-7.2.0-v20' || startsWith(matrix.redis_version, '7.4') && 'rs-7.4.0-v8' || '' }}
+      REDIS_MODULES_PORT: ${{ (startsWith(matrix.redis_version, '7.2') || startsWith(matrix.redis_version, '7.4')) && '6383' || '6381' }}
       DRIVER: ${{ matrix.driver == 'hiredis' && 'hiredis' || 'ruby' }}
       LOW_TIMEOUT: "0.14"
       # TruffleRuby defaults to a throughput-tuned JIT mode that warms up slowly;
@@ -79,8 +85,13 @@ jobs:
       - name: Boot Redis (all topologies)
         run: docker compose --profile all up -d --wait
 
-      - name: Run standalone, distributed, sentinel suites
-        run: bundle exec rake test:redis test:distributed test:sentinel
+      # Only on Redis < 8, where modules aren't in core; on >= 8 the module tests use standalone.
+      - name: Boot Redis Stack (module tests on Redis < 8)
+        if: env.REDIS_STACK_VERSION != ''
+        run: docker compose --profile modules up -d --wait
+
+      - name: Run standalone, distributed, sentinel, module suites
+        run: bundle exec rake test:redis test:distributed test:sentinel test:modules
 
       - name: Run cluster suite
         env:
@@ -90,4 +101,4 @@ jobs:
 
       - name: Tear down Redis
         if: always()
-        run: docker compose --profile all down -v
+        run: docker compose --profile all --profile modules down -v
--- a/lib/redis/commands/modules/json.rb
+++ b/lib/redis/commands/modules/json.rb
@@ -16,19 +16,30 @@ class Redis
     module Json
       # Set the JSON value at +path+ in the document stored under +key+.
       #
+      # By default +value+ is a Ruby object that is serialized to JSON text with
+      # JSON.generate. Pass +raw: true+ to send an already-encoded JSON string through
+      # untouched (no re-serialization), which avoids double-encoding pre-built JSON.
+      #
       # @example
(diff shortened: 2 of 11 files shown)


.github/workflows/test.yaml
```diff
@@ -90,4 +101,4 @@ jobs:
 
       - name: Tear down Redis
         if: always()
-        run: docker compose --profile all down -v
+        run: docker compose --profile all --profile modules down -v
```
petyaslavova, line 104: Shouldn't `all` cover also `modules`

test/lint/json.rb
```diff
@@ -0,0 +1,94 @@
+# frozen_string_literal: true
+
+module Lint
+  module Json
+    def setup
+      super
+      omit_unless_module("ReJSON")
```
petyaslavova: Maybe it will be a good idea here to raise some error when the module is missing. When we run the module tests we should expect to have all modules we want to test available in the test server.
If we just skip those tests when the module is not available, there might be some major problem that would be missed because the tests will just be skipped.

lib/redis/commands/modules/json.rb
```diff
@@ -0,0 +1,83 @@
...
+    module Json
+      # Set the JSON value at +path+ in the document stored under +key+.
+      #
+      # By default +value+ is a Ruby object that is serialized to JSON text with
+      # JSON.generate. Pass +raw: true+ to send an already-encoded JSON string through
+      # untouched (no re-serialization), which avoids double-encoding pre-built JSON.
+      #
+      # @example
+      #   redis.json_set("doc", "$", { "a" => 1, "nested" => { "b" => 2 } })
+      #     # => true
```
petyaslavova: There is no test validating the return result of the regular set. 
The json_set return type is only validated when `nx` and `xx` are provided.
In these two examples `true` is shown as return value, but at the end of the docstring is stated that for those two we should have raw +"OK"+ reply.

lib/redis/commands/modules/json.rb
```diff
@@ -0,0 +1,83 @@
...
+      #   when +raw+ is true
+      # @param [Boolean] nx only set when the path does not already exist
+      # @param [Boolean] xx only set when the path already exists
+      # @param [Boolean] raw treat +value+ as an already-encoded JSON string and send it as-is
+      # @return [Boolean, String] when +nx+ or +xx+ is given, +true+ on success and +false+
+      #   when the condition was not met; otherwise the raw +"OK"+ reply
+      def json_set(key, path, value, nx: false, xx: false, raw: false)
+        value = ::JSON.generate(value) unless raw
+        args = [:"JSON.SET", key, path, value]
+        args << "NX" if nx
```
petyaslavova, line 50: NX and XX cannot both be true, maybe additional guard for improper command usage would be beneficial - this could also be added to the docstring.

.github/workflows/test.yaml
vladvildanov: I decided to decouple those, the reason is that we do not need Redis Stack on versions >= 8.0, so I just spin up Redis Stack separately in case we're testing with Redis < 8.0

lib/redis/commands/modules/json.rb
vladvildanov: Good catch, the problem that this is a pattern throughout whole library. But I agree we can start validating those.

test/lint/json.rb
vladvildanov: Done

lib/redis/commands/modules/json.rb
vladvildanov: Fixed

lib/redis/commands/modules/json.rb
vladvildanov: Added

petyaslavova (approved):
LGTM.

PR #1349, opened 2026-06-10
Added support for remaining JSON commands

vladvildanov:
> [!NOTE]
> **Low Risk**
> Additive API and tests on the JSON command module; normalization logic is isolated and covered by lint and protocol unit tests.
> 
> **Overview**
> Adds the remaining **RedisJSON** client commands: `json_numincrby`, `json_type`, `json_objkeys`, `json_objlen`, `json_strlen`, `json_strappend`, `json_toggle`, and `json_debug_memory`, with the same JSONPath vs legacy path return shapes as existing array helpers.
> 
> **`json_numincrby`** and **`json_type`** apply reply normalizers (`NumincrbyNormalize`, `TypeNormalize`) so RESP2 (JSON strings / flat arrays) and future RESP3 (native numbers, extra nesting, multi-match legacy paths) yield the same Ruby values. A private **`json_path?`** helper drives that branching.
> 
> **`Redis::Distributed`** forwards each new method to the node for the key. Integration coverage lives in **`test/lint/json.rb`**; **`test/modules/commands_on_json_test.rb`** unit-tests the normalizers against both protocol shapes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bc80ee9ae96122cfd84de0fb10f2014933249d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/redis/commands/modules/json.rb
+++ b/lib/redis/commands/modules/json.rb
@@ -14,6 +14,38 @@ class Redis
     # JSON.NUMINCRBY — DO return differently shaped replies under RESP3 and will need
     # protocol-aware reshaping once RESP3 is supported at this layer.
     module Json
+      # Normalize a JSON.NUMINCRBY reply to a protocol-independent value: an array of numbers for
+      # a JSONPath, a single number for a legacy path. Under RESP2 the result arrives as a
+      # JSON-encoded string ("[3,4]" / "3"); under RESP3 it arrives as native numbers (an array,
+      # even for a legacy path). Both shapes map to the same value.
+      NumincrbyNormalize = lambda do |reply, jsonpath|
+        return nil if reply.nil?
+
+        reply = ::JSON.parse(reply) if reply.is_a?(String) # RESP2 string form
+        if jsonpath
+          Array(reply)
+        elsif reply.is_a?(Array)
+          reply.first # RESP3 wraps even a legacy reply in an array
+        else
+          reply
+        end
+      end
+
+      # Normalize a JSON.TYPE reply to a protocol-independent value: an array of type strings for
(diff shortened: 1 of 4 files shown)


lib/redis/commands/modules/json.rb
```diff
@@ -14,6 +14,38 @@ module Commands
...
+      # JSON-encoded string ("[3,4]" / "3"); under RESP3 it arrives as native numbers (an array,
+      # even for a legacy path). Both shapes map to the same value.
+      NumincrbyNormalize = lambda do |reply, jsonpath|
+        return nil if reply.nil?
+
+        reply = ::JSON.parse(reply) if reply.is_a?(String) # RESP2 string form
+        if jsonpath
+          Array(reply)
+        elsif reply.is_a?(Array)
+          reply.first # RESP3 wraps even a legacy reply in an array
```
petyaslavova: After digging into this command it seems that legacy paths also can match multiple values.
For example, `JSON.NUMINCRBY doc '..a' 2` should update all matching numeric `a` values; RESP2 will return the legacy scalar result `"7"` (the last changed scalar value), but RESP3 returns the native match array `[nil, 4, 7, nil]`. With `reply.first`, this would normalize to `nil` instead of the RESP2-compatible `"7"`/`7`.
Maybe, also a unit test covering such case can also be added.

petyaslavova (changes requested):

lib/redis/commands/modules/json.rb
vladvildanov: Good catch

petyaslavova (approved):
LGTM.

PR #1351, opened 2026-06-15
Added support for RESP3 protocol

vladvildanov:
# Summary

redis-rb has always pinned the underlying redis-client to RESP2, even though redis-client defaults to RESP3 and converts native types at the parser level. This PR makes the protocol configurable, defaults to RESP3, and ensures the change is safe: existing reply shapes are preserved across both protocols, and older servers transparently fall back to RESP2.

Because the default flips and a couple of commands intentionally change their return type under RESP3, this is intended for a major release. **RESP2 remains fully supported via protocol: 2**.

## Configuration
  
**protocol**: is now honored by every client type and defaults to **3**:

```
Redis.new                              # RESP3 (default)
Redis.new(protocol: 2)                 # opt back into RESP2
Redis::Cluster.new(..., protocol: 2)
Redis.new(sentinels: [...], protocol: 2)
```

The hard-pinned protocol: 2 was removed from `Redis::Client.config/.sentinel, Redis::Cluster::Client.config/.sentinel`, and `Redis::Cluster#initialize_client`. Added a `Redis::Client#protocol` reader.

## Backward compatibility — reply shapes

redis-client's RESP3 parser already returns native `Hash/Float/true/false/Array`, so most of redis-rb's reshape lambdas needed no change. Three were made shape-detecting so they produce identical Ruby values on either protocol (in lib/redis/commands.rb):

- Hashify — pass a native Hash through (RESP3 HGETALL/CONFIG GET/XINFO); previously it mangled it because Hash responds to each_slice.
- Pairify / FloatifyPairs — detect RESP3's nested [[k, v], …] vs RESP2's flat [k, v, …] (HRANDFIELD WITHVALUES, all *WITHSCORES, ZPOPMAX/MIN, ZMPOP, ZSCAN, ZRANDMEMBER, …).

Streams, cluster hashers, Floatify, Boolify, and BoolifySet were verified compatible without changes.

## Intentional behavior change — GEO
  
`GEOPOS and GEOSEARCH/GEORADIUS … WITHCOORD` return coordinates as strings under RESP2 but as native doubles under RESP3 (distances stay strings, hashes stay integers). We deliberately surface the native RESP3 type rather than coercing back to strings. Dedicated per-protocol tests document this (*_on_resp2 / *_on_resp3) in `test/redis/commands_on_geo_test.rb` and `cluster/test/commands_on_geo_test.rb`.

## RESP2 fallback for older servers

Servers without RESP3 (notably Redis < 6.0, which has no HELLO command) reject the HELLO 3 handshake. Since we now default to RESP3, the client transparently falls back to RESP2 so those users aren't broken — no protocol: 2 needed. Trigger is `Redis::Client.resp3_unsupported?` (`RedisClient::UnsupportedServer or a NOPROTO reply`). Implemented per client type:

  - standalone / distributed — `Redis::Client#ensure_connected` downgrades the config and retries the connect (this catches the error before translation mangles it).
  - sentinel / cluster — `Redis#with_protocol_fallback` (around send_command/send_blocking_command) rebuilds `@client` at protocol: 2 and retries; `Redis::Cluster::Client#handle_errors` re-raises the RESP3-unsupported error untranslated so it reaches the fallback.

## Test infrastructure
  
- New `PROTOCOL` env var (default 3) threaded into every test client (`test/helper.rb`, `cluster/test/helper.rb`); also fixed a pre-existing `ENV["conn"]` typo in the distributed helpers.
- CI now runs the full suite under RESP3 first, then RESP2 (`.github/workflows/test.yaml`).
- Fallback unit tests using a HELLO-rejecting mock (`test/redis/internals_test.rb`).

## Breaking changes
  
- Default protocol is now RESP3 (set protocol: 2 to restore RESP2).
- Under RESP3, GEO coordinate commands return Float instead of String.

---

> [!NOTE]
> **High Risk**
> Major-version breaking default (RESP3) changes wire handshake and GEO coordinate types for all new connections; fallback rebuilds the live client and must stay correct across sentinel/cluster and concurrent use behind the monitor.
> 
> **Overview**
> **Makes RESP3 the default** for standalone, sentinel, distributed, and cluster clients (`protocol: 3`, overridable with `protocol: 2`). Removes the previous hard-coded RESP2 defaults in `Redis::Client` and `Redis::Cluster::Client`.
> 
> **Adds a single RESP3→RESP2 fallback** in `Redis#with_protocol_fallback` (via `#synchronize`): on `HELLO` / `NOPROTO` / unsupported-server errors, the client rebuilds `@client` at protocol 2, warns once, and retries. Lower layers re-raise those errors untranslated (`Redis::Client#resp3_unsupported?`, cluster `handle_errors`) so fallback applies to commands, pipelines, `multi`/`watch`, pub/sub, and `without_reconnect`, with `inherit_socket` preserved on rebuild.
> 
> **Keeps wrapped command return shapes aligned with 5.x** by updating reshape helpers (`Hashify`, `Pairify`, `FloatifyPairs`, `sentinel`) for RESP3 maps and nested arrays; **GEO `WITHCOORD` / `GEOPOS` intentionally return `Float` under RESP3** (strings under RESP2).
> 
> **Docs and CI**: changelog/README, `specs/migration-resp3.md`, `PROTOCOL` env in tests, full suite run under RESP3 then RESP2, and RESP-aware `RedisMock` encoding.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c3624eac25a6effa217844cdf8c539758279011. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -145,25 +145,28 @@ class Redis
     end
   end
 
+  # All access to @client funnels through here: it serializes on @monitor and applies the RESP3
+  # protocol fallback. Routing pipelined/multi/watch (which all call synchronize) through the same
+  # path means they fall back to RESP2 against pre-HELLO servers just like single commands do.
   def synchronize
-    @monitor.synchronize { yield(@client) }
+    @monitor.synchronize do
+      with_protocol_fallback do
+        yield(@client)
+      end
+    end
   end
 
   def send_command(command, &block)
-    @monitor.synchronize do
-      with_protocol_fallback do
-        @client.call_v(command, &block)
-      end
+    synchronize do |client|
+      client.call_v(command, &block)
     end
   rescue ::RedisClient::Error => error
     Client.translate_error!(error)
   end
 
   def send_blocking_command(command, timeout, &block)
-    @monitor.synchronize do
-      with_protocol_fallback do
-        @client.blocking_call_v(timeout, command, &block)
-      end
+    synchronize do |client|
+      client.blocking_call_v(timeout, command, &block)
     end
   end
 
@@ -173,10 +176,14 @@ class Redis
   #
   # Standalone and distributed clients fall back deeper, in Redis::Client#ensure_connected, before
   # the error is translated; sentinel clients (plain RedisClient) and cluster clients surface the
-  # raw error here, where rebuilding @client is the only option.
+  # raw error here, where rebuilding @client is the only option. Because every @client access
+  # (single commands, pipelined, multi, watch) flows through #synchronize, wrapping it here covers
+  # them all, not just send_command.
   #
   # Must be called while holding @monitor: it closes and replaces @client, so it has to be
-  # serialized with the command execution that uses @client.
+  # serialized with the command execution that uses @client. The retried block re-reads @client, so
+  # callers must reference the rebuilt instance (via the synchronize block argument), not a cached
+  # one.
   def with_protocol_fallback
     yield
   rescue ::RedisClient::Error => error
--- a/lib/redis.rb
+++ b/lib/redis.rb
@@ -186,11 +186,12 @@ class Redis
   # notably Redis < 6.0, which has no HELLO command). Rebuild the client as RESP2 once so those
   # servers keep working without the user setting `protocol: 2`.
   #
-  # Standalone and distributed clients fall back deeper, in Redis::Client#ensure_connected, before
-  # the error is translated; sentinel clients (plain RedisClient) and cluster clients surface the
-  # raw error here, where rebuilding @client is the only option. Because every @client access
-  # (single commands, pipelined, multi, watch) flows through #synchronize, wrapping it here covers
-  # them all, not just send_command.
+  # This is the single fallback point for every client type. Each one surfaces the resp3-unsupported
+  # error here untranslated (still a RedisClient::Error): standalone/distributed via
+  # Redis::Client#call_v et al., sentinel via the plain RedisClient (which never translates), and
+  # cluster via Redis::Cluster::Client#handle_errors. Because every @client access — single commands,
+  # pipelined, multi, watch, and the pub/sub socket — flows through #synchronize, wrapping it here
+  # covers them all.
   #
   # Must be called while holding @monitor: it closes and replaces @client, so it has to be
   # serialized with the command execution that uses @client. The retried block re-reads @client, so
@@ -200,6 +201,10 @@ class Redis
     yield
   rescue ::RedisClient::Error => error
     if @options.fetch(:protocol, 3).to_i == 3 && Client.resp3_unsupported?(error)
+      # Fires once per client: after the downgrade @options[:protocol] is 2, so this branch never
+      # re-enters. Passing `protocol: 2` explicitly skips it entirely (and silences this warning).
+      warn("Redis: #{id} does not support RESP3 (the HELLO 3 handshake failed); falling back to " \
+           "RESP2. Pass `protocol: 2` to select RESP2 explicitly and silence this warning.")
       @options = @options.merge(protocol: 2)
       @client.close
       @client = build_client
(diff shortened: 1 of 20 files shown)


test/helper.rb
petyaslavova, line 259: Shouldn't this one also have the protocol set?

test/sentinel/sentinel_test.rb
petyaslavova, line 1: All the expected responses in this file are transformed directly in resp3 format. Shouldn't we still have the RESP2 format tested here? Having the whole test suite executed with both protocols suggests that we should still expect to have both formats validated.

lib/redis/client.rb
```diff
@@ -20,19 +20,42 @@ class Client < ::RedisClient
...
+        # survives, so match it to still trigger the RESP2 fallback for pre-6.0 clusters. Guarded by
+        # defined? because the cluster error class is only loaded with the redis-clustering gem.
+        defined?(::RedisClient::Cluster::InitialSetupError) &&
+          error.is_a?(::RedisClient::Cluster::InitialSetupError) &&
+          (error.message.include?("NOPROTO") || error.message.include?("HELLO command"))
+      end
+
+      # Pin the given config to RESP2 so subsequent (re)connects skip the `HELLO 3` handshake.
+      def downgrade_to_resp2!(config)
+        config.instance_variable_set(:@protocol, 2)
```
petyaslavova: `downgrade_to_resp2!` writes directly to `@protocol` via instance_variable_set — a private field of `RedisClient::Config`.

The sentinel and cluster paths avoid this by closing and rebuilding `@client` in `with_protocol_fallback`. 
Can't we propagate the UnsupportedServer/NOPROTO error untranslated from `ensure_connected` so `with_protocol_fallback` handles standalone too — making `downgrade_to_resp2!` unnecessary?

lib/redis.rb
```diff
@@ -145,22 +157,56 @@ def initialize_client(options)
...
+  # serialized with the command execution that uses @client. The retried block re-reads @client, so
+  # callers must reference the rebuilt instance (via the synchronize block argument), not a cached
+  # one.
+  def with_protocol_fallback
+    yield
+  rescue ::RedisClient::Error => error
+    if @options.fetch(:protocol, 3).to_i == 3 && Client.resp3_unsupported?(error)
+      @options = @options.merge(protocol: 2)
+      @client.close
+      @client = build_client
```
petyaslavova, line 215: Maybe it's worth adding a warning message here, so that users would know that the protocol has changed under the hood.

test/helper.rb
vladvildanov: Fixed

lib/redis/client.rb
vladvildanov: Good suggestion, changed it the way it bubbles before the translation, so a single way to rebuild a client for all types of clients

lib/redis.rb
vladvildanov: Added

lib/redis.rb
```diff
@@ -203,6 +206,11 @@ def with_protocol_fallback
       @options = @options.merge(protocol: 2)
       @client.close
       @client = build_client
+      # Warn only once the RESP2 client is actually in place — if the rebuild itself raises we
+      # haven't really fallen back. Fires once per client: @options[:protocol] is now 2, so this
+      # branch never re-enters. Passing `protocol: 2` explicitly skips it (and silences this).
+      warn("Redis: #{id} does not support RESP3 (the HELLO 3 handshake failed); falling back to " \
```
petyaslavova: What is included in the `id`? Can this expose sensitive information in the logs?

vladvildanov: Fixed, removed id entirely. For now it doesn't expose any sensitive data, but this comes from downstream driver and we cannot guarantee it wouldn't in a future

petyaslavova (approved):
LGTM.

PR #1356, opened 2026-06-25
Added support for Redis Query Engine

vladvildanov:
## Summary

This PR adds first-class support for the `Redis Query Engine` and aligns it with the module-command structure established by the JSON module. It started from an earlier standalone RQE implementation and was adapted to fit the current `lib/redis/commands/modules/**` layout, then extended with reply reshaping, protocol compatibility, broad test coverage, and runnable examples.

Kudos to @bsbodden for the initial RQE implementation 

All FT.* commands are available on any Redis (and, by inheritance, Redis::Cluster) instance. The change is purely additive — no existing behavior changes.

### What's included

30 ft_* command methods (lib/redis/commands/modules/search/miscellaneous.rb) covering:
- **Index lifecycle**: `ft_create, ft_alter, ft_dropindex, ft_info`
- **Querying**: `ft_search, ft_aggregate, ft_cursor_read/ft_cursor_del, ft_explain, ft_profile`
- **Hybrid (lexical + vector)**: `ft_hybrid_search`
- **Suggestions**: `ft_sugadd/ft_sugget/ft_suglen/ft_sugdel`
- **Dictionaries, synonyms, aliases, tag values, spellcheck, config**

A high-level builder/object layer (the recommended entry point):
- **Index** — bundles schema + prefix + connection; create_index returns it and it exposes `#add/#search/#aggregate/#hybrid_search/#info/#drop/…`
- **Schema / Field** (Text/Numeric/Tag/Geo/GeoShape/Vector incl. FLAT/HNSW/SVS‑VAMANA)
- **Query** (fluent + predicate DSL), IndexDefinition/IndexType
- `AggregateRequest + Reducers + Asc/Desc + Cursor`
- **Hybrid builders**: `HybridSearchQuery, HybridVsimQuery, HybridQuery, CombineResultsMethod, HybridPostProcessingConfig, HybridCursorQuery`

Typed, protocol-independent result objects (result.rb): `SearchResult/Document, AggregateResult, and HybridResult`, produced by a `ResultParser` that normalizes both RESP2 (flat arrays) and RESP3 (native maps) to the same Ruby objects. `FT.INFO/FT.CONFIG GET/FT.SYNDUMP/FT.SPELLCHECK` return plain Hashes.

Both styles interoperate — you can call a raw ft_* method with a hand-built query string and still get a reshaped result back.

### Layering

`Index (high-level)  →  Builders (Schema/Query/AggregateRequest/Hybrid*)  →  ft_* commands  →  ResultParser (RESP2/RESP3 → objects)`.
 
See `specs/query-engine.md` for the full design and an abstractions-vs-raw-commands guide.

### Behavior notes

- Default **dialect = 2** (Search::DEFAULT_DIALECT). Applied on every path — `Query, AggregateRequest`, and the raw `ft_search/ft_aggregate` string forms — overridable per query.
- **Protocol**: works under both RESP3 (the gem default) and RESP2.
- `Redis::Distributed`: not supported (FT.* is index-scoped, not key-shardable) — methods are simply absent. `Redis::Cluster` inherits the commands via the Commands mixin; the module test suite exercises standalone only (consistent with the JSON module).

### Examples & docs

- `examples/search/` — 10 runnable examples, Index-API-first (quickstart, ft_queries, range, aggregations, geo, vector_similarity, with_hashes, with_json, hybrid), plus search_raw_ft_commands.rb showing the low-level alternative.
- `specs/query-engine.md` — architecture + usage guide.

---

> [!NOTE]
> **Medium Risk**
> Large new surface area touching command encoding and reply parsing across RESP2/RESP3; regressions would affect any app adopting search, but existing clients are unchanged if they do not call FT APIs.
> 
> **Overview**
> Adds **first-class Redis Query Engine (RediSearch / `FT.*`)** support to redis-rb, wired through `lib/redis/commands/modules/search` and mixed into the main `Redis` client (and cluster via the same mixin).
> 
> Callers get roughly **30 `ft_*` methods** (create/search/aggregate/hybrid, cursors, suggestions, synonyms, dictionaries, aliases, spellcheck, config, etc.) plus a **high-level `Index` API** via `create_index` — schema builders, fluent `Query` / predicate DSL, `AggregateRequest` + reducers, and **hybrid search** builders for `FT.HYBRID`. Replies are normalized by a **`ResultParser`** into `SearchResult`/`Document`, `AggregateResult`, and `HybridResult` for both RESP2 and RESP3; **dialect 2** is the default on search/aggregate paths.
> 
> **Docs and CI:** `CHANGELOG` documents the feature; **`examples/search/`** adds runnable quickstart, geo, vector, hybrid, and raw-command examples; CI adds an **unstable Redis** matrix row for Query Engine coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 659d885199d09b997172d753a2f67c39ee23769e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/redis/commands/modules/search/aggregation.rb
```diff
@@ -0,0 +1,360 @@
...
+        # Add one +APPLY+ step per expression.
+        #
+        # @param expressions [Hash{String, Symbol => String}] map of result alias => expression
+        # @return [self]
+        def apply(expressions)
+          expressions.each do |as, expression|
+            @steps << ["APPLY", expression, "AS", as.to_s]
+          end
+          self
+        end
```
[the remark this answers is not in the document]
vladvildanov: This is not a bug, test case provided

lib/redis/commands/modules/search/miscellaneous.rb
```diff
@@ -0,0 +1,626 @@
...
+        args.concat(post_processing.build_args) if post_processing
+        if params_substitution
+          args << "PARAMS" << params_substitution.size * 2
+          params_substitution.each do |key, value|
+            args << key.to_s << value
+          end
+        end
+        args.concat(["TIMEOUT", timeout]) if timeout
+        args.concat(cursor.build_args) if cursor
+        send_command(args) { |reply| ResultParser.hybrid(reply) }
```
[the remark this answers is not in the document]
vladvildanov: Hybrid doesn't accept DIALECT argument

lib/redis/commands/modules/search/result.rb
```diff
@@ -0,0 +1,465 @@
...
+          return reply if reply.nil?
+
+          cursor = nil
+          body = reply
+          # WITHCURSOR wraps the aggregate body and the cursor id in a 2-element array. The
+          # body is a flat array under RESP2 and a native map under RESP3.
+          if reply.is_a?(Array) && reply.length == 2 && reply[1].is_a?(Integer) &&
+             (reply[0].is_a?(Array) || reply[0].is_a?(Hash))
+            body = reply[0]
+            cursor = reply[1]
```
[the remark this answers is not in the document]
vladvildanov: I checked, it's always an integer regardless of protocol

lib/redis/commands/modules/search/index_definition.rb
```diff
@@ -0,0 +1,112 @@
...
+        # @raise [ArgumentError] if +index_type+ is given but is not {IndexType::HASH} or {IndexType::JSON}
+        def initialize(
+          prefix: [], filter: nil, language_field: nil, language: nil,
+          score_field: nil, score: 1.0, payload_field: nil, index_type: nil
+        )
+          @args = []
+          @index_type = index_type
+          # Array() makes the prefix handling nil-safe (nil -> []) and wraps a lone String prefix
+          # into a one-element list, so PREFIX always emits the correct count.
+          @prefixes = Array(prefix)
```
petyaslavova: In case a list with prefixes is provided, the call to Arrays(prefix) will return the same list instance, and if the caller later changes the values in the original prefix instance - the changes will affect the current object as well. Maybe it worths duplicating the input list to ensure that the content of the IndexDefinition won't change due to some external object update.

lib/redis/commands/modules/search/aggregation.rb
```diff
@@ -0,0 +1,360 @@
...
+      end
+
+      # Represents an +FT.CURSOR+ read for paging through aggregation results.
+      class Cursor
+        # @return [Integer] the cursor id
+        # @return [Integer] the +MAXIDLE+ in milliseconds (0 to omit)
+        # @return [Integer] the +COUNT+ (batch size, 0 to omit)
+        attr_accessor :cid, :max_idle, :count
```
petyaslavova: I think the `max_idle` should be removed from this class. According to the official docs, max_idle is not supported for [ft.cursor-read](https://redis.io/docs/latest/commands/ft.cursor-read/), and this object seems to be used only for this command (the max_idle for ft.aggregate is specified above).

lib/redis/commands/modules/search/field.rb
```diff
@@ -0,0 +1,292 @@
...
+      # A +GEOSHAPE+ field, indexing geometric shapes under a coordinate system.
+      class GeoShapeField < Field
+        FLAT = "FLAT"
+        SPHERICAL = "SPHERICAL"
+
+        # Build a +GEOSHAPE+ field.
+        #
+        # @param [String, Symbol] name the document attribute the field indexes
+        # @param [String] coord_system the coordinate system, {FLAT} or {SPHERICAL}
+        def initialize(name, coord_system = FLAT, **options)
```
petyaslavova: The default server value is SPHERICAL. By providing here FLAT as default coord_system we never allow the default server behaviour to be used.

lib/redis/commands/modules/search/index.rb
```diff
@@ -0,0 +1,341 @@
...
+          # Validate fields
+          fields.each do |field_name, value|
+            field = @schema.field(field_name)
+            if field.is_a?(NumericField) && !value.is_a?(Numeric)
+              raise Redis::CommandError, "Invalid value for numeric field '#{field_name}': #{value}"
+            end
+          end
+
+          begin
+            json? ? @redis.json_set(key, "$", fields) : @redis.hset(key, fields)
```
petyaslavova, line 117: `add` for JSON indexes currently writes the provided kwargs directly at the JSON root via` json_set(key, "$", fields)`. 
That only works when schema paths are root-level (`$.name AS name`), but not for nested paths like `$.user.name AS name`.
`index.add("1", name: "Ann")` stores `{"name":"Ann"}`, while the index expects `{"user":{"name":"Ann"}}`.
We should either document/test that JSON add requires the full JSON document shape, or reject alias-shaped writes for nested JSON paths with a clear error.

test/lint/search.rb
```diff
@@ -0,0 +1,1452 @@
...
+      rescue Redis::CommandError
+        nil
+      end
+      r.flushdb
+    end
+
+    def wait_for_index(index_name, timeout = 5.0)
+      deadline = now + timeout
+      loop do
+        info = r.ft_info(index_name)
```
petyaslavova: You should have some error handling around this command, but still inside the loop - it will raise an error if the index is not yet existing and will basically silently just exit and won't wait.

test/lint/search.rb
```diff
@@ -0,0 +1,1452 @@
...
+      schema = Schema.build do
+        vector_field :embedding, :flat, type: :float32, dim: 4, distance_metric: :l2
+        tag_field :tag
+      end
+      r.create_index(@index_name, schema)
+
+      r.hset("doc1", embedding: [0.1, 0.9, 0.2, 0.8].pack("f*"), tag: "a")
+      r.hset("doc2", embedding: [0.2, 0.8, 0.3, 0.7].pack("f*"), tag: "b")
+      r.hset("doc3", embedding: [0.8, 0.2, 0.7, 0.3].pack("f*"), tag: "c")
+
```
petyaslavova, line 693: You can include wait_for_index() here as well.

test/lint/search.rb
```diff
@@ -0,0 +1,1452 @@
...
+    def test_hybrid_search_knn
+      schema = Schema.build do
+        text_field :title
+        vector_field :embedding, :flat, type: :float32, dim: 2, distance_metric: :cosine
+      end
+      r.create_index(@index_name, schema)
+
+      r.hset("doc1", title: "foo", embedding: [0.1, 0.9].pack("f*"))
+      r.hset("doc2", title: "bar", embedding: [0.8, 0.2].pack("f*"))
+
```
petyaslavova, line 717: ditto

lib/redis/commands/modules/search/query.rb
```diff
@@ -0,0 +1,676 @@
...
+        # @param field [String] the field name to return
+        # @param as_field [String, nil] an alias to expose the field under (+AS+)
+        # @param decode_field [Boolean] whether the returned value should be JSON-decoded in results
+        # @return [self]
+        def return_field(field, as_field: nil, decode_field: true)
+          @return_fields ||= []
+          @return_fields_decode ||= {}
+
+          @return_fields << field
+          @return_fields_decode[field] = decode_field
```
petyaslavova: When `as_field` is used, Redis returns the value under the alias, but `decode_fields` is keyed by the original field/path here. For `return_field("$.user", as_field: "user")`, the parser sees `"user"` and won’t decode because the map contains `"$.user"`. This should key decoding by `as_field || field`.

lib/redis/commands/modules/search/index_definition.rb
vladvildanov: Fixed

lib/redis/commands/modules/search/aggregation.rb
vladvildanov: Removed

lib/redis/commands/modules/search/field.rb
vladvildanov: Fixed

lib/redis/commands/modules/search/index.rb
vladvildanov: Documented given behaviour with root path only

test/lint/search.rb
vladvildanov: Got it, now it handles error but still continue loop in case the index info is not immediately accessible

test/lint/search.rb
vladvildanov: Done

test/lint/search.rb
vladvildanov: Done

lib/redis/commands/modules/search/query.rb
vladvildanov: Good catch, fixed by checking with OR condition

lib/redis/commands/modules/search/query.rb
```diff
@@ -0,0 +1,681 @@
...
+        # @param field [String] the text field name
+        # @param pattern [String] the text pattern to match
+        def initialize(field, pattern)
+          super(field)
+          @pattern = pattern
+        end
+
+        # @return [String] the query-string fragment, e.g. +(@title:hello)+
+        def to_s
+          "(@#{@field}:#{@pattern})"
```
petyaslavova: **Multi-word `match` binds the field modifier to only the first token.**

`TextMatchPredicate#to_s` renders `"(@#{field}:#{pattern})"`, so a multi-word
pattern isn't grouped. RediSearch applies `@field:` only to the term
immediately after it — the rest become global (all-field) terms.

```ruby
index = redis.create_index("books", schema, prefix: "book")
index.add("1", title: "red shoes",  body: "leather")
index.add("2", title: "blue hat",   body: "red shoes on sale")  # title is NOT "red shoes"

result = index.search { text(:title).match("red shoes") }
# query string built  -> "(@title:red shoes)"
# parsed as: @title:red  AND  shoes (any field)  -> this leads to results silently widened
# RediSearch reads it as: @title:red  AND  shoes(any field)
# => doc "2" ALSO matches (title has "red", body has "shoes")
```

Should render `@title:(red shoes)` so the modifier covers the whole phrase.

lib/redis/commands/modules/search/query.rb
```diff
@@ -0,0 +1,681 @@
...
+        # @param field [String] the tag field name
+        # @param value [String] the tag value to match
+        def initialize(field, value)
+          super(field)
+          @value = value
+        end
+
+        # @return [String] the query-string fragment, e.g. +(@color:{red})+
+        def to_s
+          "(@#{@field}:{#{@value}})"
```
petyaslavova: **Predicate values are interpolated without escaping query metacharacters.**

`TagEqualityPredicate`, `TextMatchPredicate`, and `RangePredicate` interpolate
user values verbatim, so any value containing RediSearch syntax changes the
query's meaning or breaks it.

```ruby
Query.new.tag(:status).eq("active|pending").to_redis_args.first
# => "(@status:{active|pending})"   # '|' is tag-OR: matches "active" OR "pending",
#                                    # not the literal value

Query.new.tag(:city).eq("New York").to_redis_args.first
# => "(@city:{New York})"           # space breaks the intended exact-tag match
```

`eq`/`match` should escape metacharacters (`|`, `{`, `}`, `:`, `-`, `@`,
spaces, …), as redis-py's query builder does.

lib/redis/commands/modules/search/field.rb
```diff
@@ -0,0 +1,293 @@
...
+        def to_args
+          args = [@name]
+          args << "AS" << @alias_name if @alias_name
+          args << @type.to_s.upcase
+          args << @coord_system if @coord_system
+
+          # Add suffix options
+          args << "NOINDEX" if @options[:no_index]
+          args << "SORTABLE" if @options[:sortable]
+
```
petyaslavova, line 222: GEOSHAPE also supports INDEXMISSING option

lib/redis/commands/modules/search/query.rb
vladvildanov: Fixed

lib/redis/commands/modules/search/query.rb
vladvildanov: Added escaping

lib/redis/commands/modules/search/field.rb
vladvildanov: Added missing parameter

petyaslavova (approved):
LGTM.

PR #1364, opened 2026-07-23
Added support for HIMPORT commands

vladvildanov:
## Summary

Implements client support for `HIMPORT PREPARE / SET / DISCARD / DISCARDALL` across all three client types, including first-class cluster fan-out and automatic recovery from connection-scoped fieldset loss.

### What HIMPORT is

`HIMPORT` is a fast-ingestion mechanism for hashes sharing one field schema: the client registers field names once (`HIMPORT PREPARE fieldset field ...`) and then creates hashes sending only values (`HIMPORT SET key fieldset value ...`). The hard part for a client is that fieldsets are server-side session state scoped to one physical connection: they are destroyed by `disconnect/RESET`, are invisible to other connections, and a `SET` on a connection that didn't `PREPARE` fails with `ERR no such fieldset`.

### API

```
redis.himport_prepare("users", ["name", "email", "age"])          # => "OK"
redis.himport_set("user:1", "users", ["alice", "a@x.com", "25"])  # => "OK"
redis.himport_discard("users")                                    # => 1
redis.himport_discard_all
```

- `lib/redis/commands/hashes.rb` — four transport-pure methods (available on standalone clients and inside pipelined/multi automatically). Client-side validation is limited to non-empty fields/values; fields/values are never deduplicated or reordered; empty strings are valid names; server errors propagate unmodified as `Redis::CommandError`.
- `Redis::Distributed` — `himport_prepare/himport_discard/himport_discard_all` fan out via `on_each_node` (per-node array replies, the existing Distributed convention); `himport_set routes` via `node_for(key)`.
- `Redis::Cluster` — see below.

### Fieldset-loss recovery (himport_auto_prepare, default on)

A `Redis` instance transparently replaces a dead connection (default reconnect_attempts: 1), silently destroying prepared fieldsets. The client therefore keeps a registry of the last schema prepared per fieldset name and, when a `himport_set` fails with no such fieldset and the registry has the schema, re-issues the `PREPARE` (cluster: re-fans it out to all masters) and retries the `SET` exactly once. Explicitly discarded fieldsets are removed from the registry and never resurrected. One mechanism covers reconnects, idle-timeout reconnects, sentinel failover, `RESET` by connection sanitizers, cluster failover/topology rebuilds and `MOVED/ASK` onto fresh connections.

Opt out with `Redis.new(himport_auto_prepare: false)` for a fully explicit fieldset lifecycle; the registry is still recorded to support manual recovery. Recovery applies to direct calls only — commands inside pipelined/multi blocks are untouched (the recommended high-throughput pattern is a single pipeline carrying the PREPARE and its SETs, which always executes as one batch on one socket).

### Cluster support and its risks

~The server advertises `request_policy:all_shards` for `PREPARE/DISCARD/DISCARDALL` and a key at position 2 for `SET`. `redis-cluster-client (0.16.x)` parses neither command tips nor per-subcommand key specs — a container command gets the parent spec (first key = 0), so without intervention `HIMPORT SET` would route to an arbitrary node and `PREPARE` would reach one node only. This PR adds explicit routing in the cluster gem:~

- ~`Redis::Cluster::Client#himport_fan_out` — executes on all primaries; partial failures raise `Redis::Cluster::CommandErrorCollection` (per-node #errors), with the same `renew_cluster_state` handling `FLUSHALL's` fan-out gets.~
- ~`Redis::Cluster::Client#himport_call_by_key` — slot-routes `SET` via the router's public `find_node_key_by_key/send_command_to_node`, preserving `MOVED/ASK/CLUSTERDOWN` handling.~
- ~`Redis::Cluster#himport_*` overrides aggregate fan-out replies for standalone/cluster API parity (FLUSHALL-style): "OK" for prepare, max across masters for the discard counters.~

### ⚠️ Risk: private upstream API usage

~The primaries fan-out reaches one private method of the underlying driver: `router.send(:send_command_to_primaries, ...)`. There is no public all-primaries primitive in `redis-cluster-client 0.16.x`. Mitigations:~

- ~the dependency is pinned to~ ~> 0.16.0, ~so the coupling can only break on a deliberate bump;~
- ~the cluster test suite exercises the call against a real cluster, so a future bump that removes or reshapes the method fails loudly in CI;~
- ~the call site carries a TODO. The durable fix is upstreaming command-tips (`request_policy:all_shards`) and subcommand key-spec parsing into `redis-cluster-client` (https://github.com/redis-rb/redis-cluster-client/issues/532); once that lands, both helper methods can be deleted without any public API change.~

~Everything else the cluster layer touches (`find_node_key_by_key, find_node, send_command_to_node, renew_cluster_state, router access from Redis::Cluster::Client`) is either public or has existing precedent in this repo (watch).~

~The issue has been addressed: https://github.com/redis-rb/redis-cluster-client/issues/534~

### ⚠️ Limitation: HIMPORT inside cluster transactions

~A cluster `MULTI` must pin to a single node, and the driver derives that node from the first command's key — which it cannot extract from any `HIMPORT` command (container-command spec). As a result:~

- ~A multi consisting only of `HIMPORT` commands raises `Redis::Cluster::TransactionConsistencyError` at `EXEC` — including the `himport_prepare + himport_set` combination. This is deliberate and covered by tests: even if the transaction could pin, a `PREPARE` inside it would reach one node's connection only, silently violating the all-shards contract; failing loudly is safer.~
- ~`himport_set` does work inside a multi pinned by a normal keyed command in the same hash slot (covered by tests, both orderings), because the fan-out `PREPARE` put the fieldset on every master's connection.~
- ~Inside a cluster pipelined block, `HIMPORT` commands route to the pipeline's arbitrary node — prepare does not fan out there. Documented as unsupported; use direct calls on cluster.~

~Standalone pipelined/multi have no such limitations and are fully supported (and tested).~

**All the risks about were solved in the downstream driver as a part of** https://github.com/redis-rb/redis-cluster-client/pull/535

---

> [!NOTE]
> **Medium Risk**
> New experimental ingestion API and automatic re-prepare change behavior on reconnect; cluster users inherit breaking command-routing semantics from the pinned redis-cluster-client 0.16.7.
> 
> **Overview**
> Adds **experimental** client support for Redis 8.10’s `HIMPORT` family (`himport_prepare`, `himport_set`, `himport_discard`, `himport_discard_all`) for bulk hash ingestion with a shared field schema.
> 
> On standalone `Redis`, the client keeps a per-fieldset schema registry and, by default (`himport_auto_prepare`), re-issues `PREPARE` and retries `SET` once when the server reports `no such fieldset` after reconnect or connection loss; discarded fieldsets are not restored. **`Redis::Distributed`** fans prepare/discard to every ring node, routes `SET` by key, replays registered fieldsets when nodes are added, and returns per-node reply arrays for fan-out commands. **`Redis::Cluster`** relies on `redis-cluster-client` command-tip routing and aggregates multi-master fan-out replies to match the standalone API; loss recovery re-fans out through the same path.
> 
> Pins **`redis-client`** to `0.30.1` and **`redis-cluster-client`** to `0.16.7` (exact versions instead of patch ranges). The cluster gem changelog documents **breaking** routing changes from driver 0.16.6+ (`SLOWLOG`, `FUNCTION *`, `XINFO` subcommands, etc.). README adds a “Bulk hash ingestion (HIMPORT)” section; shared lint tests cover standalone, distributed, and cluster.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3036d854c339120d0f33731a01ef76be91b264e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/test/distributed/himport_test.rb
+++ b/test/distributed/himport_test.rb
@@ -1,0 +1,131 @@
+# frozen_string_literal: true
+
+require "helper"
+
+class TestDistributedHimport < Minitest::Test
+  include Helper::Distributed
+  include Lint::Himport
+
+  # Fan-out commands return one reply per ring node (the Redis::Distributed
+  # convention, cf. flushdb/script); the default test ring has a single node.
+
+  def test_himport_prepare_returns_ok
+    target_version "8.9" do
+      assert_equal ["OK"], r.himport_prepare("fs", %w[f1 f2])
+    end
+  end
+
+  def test_himport_discard_semantics
+    target_version "8.9" do
+      r.himport_prepare("fs", %w[f1])
+      r.himport_set("k1", "fs", %w[v])
+
+      assert_equal [1], r.himport_discard("fs")
+      assert_equal [0], r.himport_discard("fs")
+
+      error = assert_raises(Redis::CommandError) do
+        r.himport_set("k2", "fs", %w[v])
+      end
+      assert_match(/no such fieldset/, error.message)
+
+      assert_equal({ "f1" => "v" }, r.hgetall("k1"))
+    end
+  end
+
+  def test_himport_discard_all_counts
+    target_version "8.9" do
+      r.himport_prepare("fs1", %w[f1])
+      r.himport_prepare("fs2", %w[f1 f2])
+
+      assert_equal [2], r.himport_discard_all
+      assert_equal [0], r.himport_discard_all
+    end
+  end
+
+  # A second ring node on another DB of the same server is still a distinct
+  # physical connection — which is the scope fieldsets live in — so these
+  # tests genuinely exercise two independent sessions.
+
+  def test_prepare_fans_out_to_all_nodes
+    target_version "8.9" do
+      r.add_node("redis://127.0.0.1:#{PORT}/14")
+      r.flushdb
+
+      key_a, key_b = keys_on_distinct_nodes
+
+      assert_equal %w[OK OK], r.himport_prepare("fs", %w[f1])
+      assert_equal "OK", r.himport_set(key_a, "fs", %w[a])
+      assert_equal "OK", r.himport_set(key_b, "fs", %w[b])
+    end
+  end
+
+  def test_set_fails_when_prepare_missed_a_node
+    target_version "8.9" do
+      r.add_node("redis://127.0.0.1:#{PORT}/14")
+      r.flushdb
+
+      key_a, key_b = keys_on_distinct_nodes
+
+      # prepare on one node's connection only
+      r.nodes[0].himport_prepare("fs", %w[f1])
+
+      assert_equal "OK", r.himport_set(key_a, "fs", %w[v])
+      error = assert_raises(Redis::CommandError) do
+        r.himport_set(key_b, "fs", %w[v])
+      end
+      assert_match(/no such fieldset/, error.message)
+    end
+  end
+
+  def test_discard_fans_out
+    target_version "8.9" do
+      r.add_node("redis://127.0.0.1:#{PORT}/14")
+      r.flushdb
+
+      r.himport_prepare("fs", %w[f1])
+      assert_equal [1, 1], r.himport_discard("fs")
+
+      r.himport_prepare("fs1", %w[f1])
+      r.himport_prepare("fs2", %w[f1])
+      assert_equal [2, 2], r.himport_discard_all
+    end
+  end
+
+  def test_key_tags_route_himport_set
+    target_version "8.9" do
+      r.add_node("redis://127.0.0.1:#{PORT}/14")
+      r.flushdb
+
+      assert_equal r.node_for("{tag}k1").id, r.node_for("{tag}k2").id
+
+      r.himport_prepare("fs", %w[f1])
+      assert_equal "OK", r.himport_set("{tag}k1", "fs", %w[v1])
+      assert_equal "OK", r.himport_set("{tag}k2", "fs", %w[v2])
+    end
+  end
+
+  def test_himport_set_recovers_per_node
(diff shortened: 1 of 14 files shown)


test/lint/himport.rb
```diff
@@ -0,0 +1,177 @@
+# frozen_string_literal: true
+
+module Lint
+  module Himport
+    def test_himport_prepare_returns_ok
+      target_version "8.9" do
+        assert_equal "OK", r.himport_prepare("fs", %w[f1 f2])
+      end
```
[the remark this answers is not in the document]
vladvildanov: It's fine, Redis 8.10 doesn't match API version, which is 8.9.x

test/distributed/himport_test.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      key_a, key_b = keys_on_distinct_nodes
+
+      assert_equal %w[OK OK], r.himport_prepare("fs", %w[f1])
+      assert_equal "OK", r.himport_set(key_a, "fs", %w[a])
+      assert_equal "OK", r.himport_set(key_b, "fs", %w[b])
+    end
+  end
+
+  def test_set_fails_when_prepare_missed_a_node
```
petyaslavova: I think this one can be renamed to focus on the fact that things won't work if prepare is called directly on a specific node.

petyaslavova (approved):
Overall LGTM. Just one tiny comment that is not a blocker at all.

test/distributed/himport_test.rb
vladvildanov: Resolved

PR #1375, opened 2026-08-13
Added support for Keyspace and Subkey notifications

vladvildanov:
## Summary

Adds first-class support for [Redis Keyspace Notifications](https://redis.io/docs/latest/develop/pubsub/keyspace-notifications/) - including the **Subkey Notifications** channel families introduced in Redis 8.8 - to both gems: typed channel builders, a binary-safe parser, a managed subscriber for standalone/sentinel, and a cluster-aware manager that closes the "pub/sub works, so notifications work" trap.

### Why a plain `subscribe` isn't enough

- **Wire formats are server-defined and subtle.** Classic channels split on the *first* `:` only (keys may contain `:`); the four Redis 8.8 subkey families (`subkeyspace`, `subkeyevent`, `subkeyspaceitem`, `subkeyspaceevent`) carry length-prefixed, binary-safe payloads where keys/subkeys may contain `|`, `,`, `:`, `\n` and arbitrary bytes. Naive string splitting corrupts them.
- **Cluster silently loses data.** Notifications are node-local — never forwarded on the cluster bus — and `RedisClient::Cluster::PubSub` routes `SUBSCRIBE`/`PSUBSCRIBE` to a *single* node. A cluster user subscribing to a notification channel today receives ~1/N of the events: correct-looking in a small test, lossy in production.
- **Delivery is fire-and-forget**, so consumers need first-class reconnect/reconcile hooks rather than hand-rolled loops.

### What's included

**Layer 1 — pure helpers** (`lib/redis/keyspace_notifications/`), usable directly inside the existing `subscribe`/`psubscribe` DSL:
- `Channels` — builders for all six families (`keyspace`, `keyevent`, `subkeyspace`, `subkeyevent`, `subkeyspaceitem`, `subkeyspaceevent`), db wildcards, BINARY-safe, `subkeyspaceitem` rejects keys containing `\n` (server never emits them).
- `Parser.parse(channel, payload)` — binary-safe, validated against a live 8.8.0 server; returns `nil` for foreign channels, raises `ParseError` (carrying raw channel/payload) for malformed bodies.
- `Notification` — immutable value object: `family`, `db`, `event` (opaque string — new server events need no gem update), `key`, `subkeys` (ordered, duplicates preserved), raw `channel`/`payload`, matched `pattern`.

**Layer 2 — standalone manager** (`Redis#keyspace_notifications`):
```ruby
manager = redis.keyspace_notifications
manager.subscribe_keyevent("expired") { |n| cache.delete(n.key) }
manager.subscribe_subkeyspace("session:*") { |n| n.subkeys.each { |f| invalidate(n.key, f) } } # 8.8+
manager.on_reconnect { cache.clear } # gaps are lossy by design — this is the reconcile hook
```
- Owns a dedicated connection (duplicated from the client's options) + one listener thread; the calling client stays fully usable.
- **Ack-then-commit consistency**: `subscribe`/`unsubscribe` block until the server acknowledges and commit local state only then. A raised `subscribe` rolls back completely (nothing fires later); a raised `unsubscribe` leaves you subscribed on both sides — just retry. In-flight messages for a pattern being unsubscribed go to its handler until the ack, never to the default handler afterwards. A listener-side invariant reverts any server-acked pattern that is no longer registered, so replay races self-heal.
- **Auto-resubscribe** on connection loss with the repo's standard `reconnect_attempts` semantics (Integer, or an Array of sleep durations; default: exponential 0.5s→30s ladder, 10 attempts, budget resets after every healthy session), then fires `on_reconnect`.
- Handler exceptions and parse errors are reported to the error handler (default: `warn`) and never kill the listener.

**Cluster manager** (`Redis::Cluster#keyspace_notifications`, `redis-clustering`):
- Fans every subscription out to **all primaries** (enumerated via public `CLUSTER SLOTS`; one dedicated sidecar connection per primary — the router stays untouched).
- N node readers feed **one bounded queue drained by a single dispatcher thread**: handlers need not be thread-safe; per-node ordering preserved, cross-node unspecified. The queue also serializes reactive topology **refresh** with dispatch (a failing node's own thread never repairs itself).
- Reactive refresh on node connection errors + public `#refresh` for scale-out (a brand-new primary emits no error signal); manual refresh raises `KeyspaceNotificationsRefreshError` with a per-node `#errors` hash. Refresh reconciles both directions: subscribes missing patterns and removes no-longer-registered ones.
- `unsubscribe` deliberately removes tracking *first* (documented): with N nodes partial failure is normal, and ack-gated tracking would make refresh re-subscribe the pattern on the N−1 nodes that already removed it. Per-node convergence comes from the core manager's ack-then-commit + refresh.

---

> [!NOTE]
> **Medium Risk**
> Large new concurrent pub/sub surface (standalone manager + per-primary cluster sidecars) with complex subscription lifecycle; changes to subscribed-client I/O synchronization affect all pub/sub users, though scoped to race/teardown fixes.
> 
> **Overview**
> Adds **keyspace notification** support end-to-end: channel builders, a binary-safe parser (classic `__keyspace@*` / `__keyevent@*` plus Redis 8.8 subkey families), typed `Notification` objects, and **`Redis#keyspace_notifications`** — a manager with its own connection/thread, ack-gated subscribe/unsubscribe, auto-resubscribe, and **`on_reconnect`** for lossy gaps. Server **`notify-keyspace-events`** is documented as operator-owned; **`Redis::Distributed`** is out of scope.
> 
> For **cluster**, **`Redis::Cluster#keyspace_notifications`** subscribes on **every primary** (node-local pub/sub), merges events through one dispatcher queue, and **reactively refreshes** topology on connection errors (plus public **`#refresh`**); raises **`KeyspaceNotificationsRefreshError`** when primaries cannot be subscribed.
> 
> **Reliability fixes:** concurrent **`unsubscribe`/`punsubscribe`** during teardown no longer surfaces **`NoMethodError`** (subscription client captured once); **`SubscribedClient`** uses sliced guarded reads and a closed flag to avoid hiredis use-after-free races; **`Redis#dup`** accepts option overrides for the manager’s connection.
> 
> **Cluster plumbing:** **`HashifyClusterNodeInfo`** handles single-slot and migrating markers; test orchestrator gets failback/reverse resharding, faster rebuild when topology is already consistent, and **`CLUSTER SHARDS`** convergence checks. **`redis-cluster-client`** pinned to **0.17.0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8d586d28722c9284404b5b4d99f565c0cefb595. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/redis/keyspace_notifications/manager.rb
```diff
@@ -0,0 +1,758 @@
...
+          until patterns.all? { |pattern| @confirmed.key?(pattern) || !@handlers[pattern].equal?(installed[pattern]) }
+            # A CommandError means the server REJECTED a command on this session
+            # (e.g. an ACL-forbidden pattern): retrying cannot fix it, so raise it
+            # promptly — the caller's rollback then removes the poisoned registration
+            # instead of the replay churning until the generic timeout. The session
+            # rejection is indivisible, so every concurrent waiter gets it; their
+            # rolled-back subscriptions are safe to retry.
+            if (rejection = @listener_error).is_a?(CommandError)
+              raise rejection
+            end
```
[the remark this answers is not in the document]
vladvildanov: This was already fixed

cluster/lib/redis/cluster/keyspace_notifications.rb
petyaslavova, line 1: Shouldn't the cluster manager also expose `on_reconnect`?

cluster/lib/redis/cluster/keyspace_notifications.rb
```diff
@@ -0,0 +1,650 @@
...
+      # @return [Boolean]
+      def closed?
+        @closed
+      end
+
+      # Close every node listener, stop the dispatcher and release all connections.
+      # Idempotent.
+      #
+      # @return [void]
+      def close
```
petyaslavova, line 400: `Cluster::KeyspaceNotifications#close` can block for a long time on a bigger cluster.
It waits on `@refresh_lock`, and a refresh in flight holds that lock across all of its per-node catch-ups (N primaries x up to 5 convergence passes x the 5s ack timeout).
After that it closes the listeners one at a time, and each `NodeListener#close` can spend up to 2 x 2s on its thread joins.
On a cluster with ten primary nodes that adds up to tens of seconds.

The root cause looks like the in-flight refresh not knowing a close is happening.
`refresh` checks `@closed` only once, on entry, so once it is past that line it runs to completion while `close` waits behind the lock.

Two changes would fix it without weakening cleanup:

- Set `@closed = true` under `@lock` before reaching for `@refresh_lock` (moving the idempotency guard up with it),
and recheck `@closed` inside the per-node loop and the convergence loop in `refresh` so it aborts promptly.
`close` can then keep the unbounded `@refresh_lock.synchronize`, so it never skips `@queue.close` or the listener teardown, but the wait becomes short.
- Close the listeners in parallel rather than serially. They are independent, so teardown becomes max-of-nodes instead of sum-of-nodes.

cluster/lib/redis/cluster/keyspace_notifications.rb
```diff
@@ -0,0 +1,650 @@
...
+        listeners.each do |node_key, listener|
+          begin
+            listener.subscribe(patterns)
+          rescue ::Redis::CommandError => error
+            # The server REJECTED a pattern (deterministic — e.g. ACL NOPERM), which
+            # best-effort handling must not swallow: left registered, the pattern
+            # would fail every future catch-up batch on every primary. Identify and
+            # evict the culprit(s) once, then raise below — matching the standalone
+            # contract that a rejected subscribe raises and leaves no registration.
+            rejected = evict_rejected(listener, node_key, patterns) if rejected.empty?
```
petyaslavova: One thing in the cluster `subscribe` fan-out logic: after the first node reports a rejected pattern, the errors from all the other nodes are lost.
Once `rejected` is set, the inner rescue skips both `evict_rejected` and the `raise`, so the error never reaches the outer `rescue StandardError` that calls `report_error` and `request_refresh`.
Those nodes end up with neither the valid patterns subscribed nor any error report, and the only thing that saves them is the refresh you request afterwards.

`refresh` already does this the right way: it collects `failures[node_key] = error` for every node and raises once with the full hash.
Same accumulation can be used here as well - collect every per-node error inside the loop, report them all through `report_error` after the loop, then raise.

petyaslavova (changes requested):

cluster/lib/redis/cluster/keyspace_notifications.rb
vladvildanov: Done

cluster/lib/redis/cluster/keyspace_notifications.rb
vladvildanov: Fixed. Now it's paralelised

cluster/lib/redis/cluster/keyspace_notifications.rb
vladvildanov: Done, now we accumulate errors

lib/redis/keyspace_notifications/manager.rb
```diff
@@ -0,0 +1,1181 @@
...
+
+          @closing = true
+          # Wake a listener parked in a reconnect backoff — closing the socket can't
+          # interrupt a plain sleep, and a 30s delay would outlive the bounded joins.
+          @cond.broadcast
+          @thread
+        end
+
+        begin
+          @redis.punsubscribe if thread&.alive? && @redis.subscribed?
```
petyaslavova, line 451: This is the hiredis CI failure, and it's a real crash, not a flake.

`close` writes PUNSUBSCRIBE from the closing thread while the listener thread may be closing the same connection. In redis-client  `PubSub#call_v` reads `@raw_connection` then writes to it, while `PubSub#close` frees the hiredis context first and nils the ivar second. Read-then-free = use-after-free.

In the failing job: SIGSEGV in `redisBufferWrite`, frames `manager_test.rb:26 -> manager.rb:489 -> subscribe.rb:11`. The `NoMethodError` rescue in `write_to_session` only covers the case where the ivar was already nil, and no Ruby rescue catches the native one. Every `write_to_session` from a caller thread has the same exposure, so this is reachable in normal use.

This can be fixed in `SubscribedClient`: hold the existing `@write_monitor` inside `close` too, and set a `@closed` flag that `call_v` checks and raises `SubscriptionError` on. Teardown then can't free the connection under an in-flight write. Please add a regression test that closes concurrently with a session-killing rejection.

cluster/lib/redis/cluster/keyspace_notifications.rb
```diff
@@ -0,0 +1,780 @@
...
+          stale = @lock.synchronize do
+            expect_subscribed = !@registry.empty?
+            gone_keys = @listeners.keys - primaries.keys
+            primaries.each_key do |node_key|
+              listener = @listeners[node_key]
+              gone_keys << node_key if listener && !listener.healthy?(expect_subscribed)
+            end
+            gone_keys.map { |node_key| @listeners.delete(node_key) }
+          end
+          stale.compact.each(&:close)
```
petyaslavova: The serial-teardown stall that was fixed in `#close` is still here. Line 335 (`stale.compact.each(&:close)`) and line 408 close listeners one at a time while holding `@refresh_lock`, and `#close` waits on that lock unboundedly.

After a cluster-wide blip, every listener is mid-reconnect, so `healthy?` is false for all of them, and the prune closes N listeners serially, each up to ~4s. Same fix as `#close`: fan the prune out over threads and join.

lib/redis/keyspace_notifications/manager.rb
```diff
@@ -0,0 +1,1181 @@
...
+        rescue StandardError
+          nil # the listener may be mid-teardown; the force-close below covers it
+        end
+
+        if thread && !thread.equal?(Thread.current) && !thread.join(timeout)
```
petyaslavova: `close` has no `ensure`, and `thread.join` re-raises whatever the listener thread died with. The loop tail at 688-704 isn't inside the inner rescue, so e.g. a `reconnect_attempts` that's neither Integer nor Array raises `NoMethodError` from `@reconnect_attempts[attempts]`.

Then `close` skips `force_close_redis` and `@closed = true`: connection leaked, `@closing` stuck true so every later `subscribe` raises, and `closed?` still says false. Wrap the teardown in `ensure`, and validate `reconnect_attempts` in the constructor so it fails at the call site instead of after a connection loss.

lib/redis/keyspace_notifications/manager.rb
```diff
@@ -0,0 +1,1181 @@
...
+              # through with stale ownership and leave the pattern deaf.
+              patterns.each { |pattern| @removing.delete(pattern) if @removing[pattern].equal?(owned[pattern]) }
+              raise SubscriptionError, "timed out waiting for unsubscription confirmation"
+            end
+
+            @cond.wait([remaining, 0.05].min)
+            pending = patterns.select do |pattern|
+              @confirmed.key?(pattern) && @handlers[pattern].equal?(owned[pattern])
+            end
+            punsubscribe_quietly(pending)
```
petyaslavova: The subscribe path was bounded to one re-issue per session (see the reasoning at 915-928), but this one re-issues PUNSUBSCRIBE every 50ms - up to ~100 duplicates against a slow node.

Each late ack then deletes `@confirmed[pattern]` and may re-issue a psubscribe. In the single-pattern case that makes `subscribed?` briefly false, which is exactly what `NodeListener#healthy?` prunes a whole node on. Please bound it like the subscribe path.

lib/redis/keyspace_notifications/manager.rb
```diff
@@ -0,0 +1,1181 @@
...
+              @cond.broadcast
+              entry = @handlers[key]
+              # Wanted again unless this ack answers the unsubscribe that targets the
+              # live registration (the normal flow, where deletion follows the ack).
+              !entry.nil? && !entry.equal?(@removing[key])
+            end
+            # A registered pattern lost its server-side subscription to an unsubscribe
+            # aimed at an older, since-replaced registration (the two block-less writes
+            # crossed on the wire): re-establish it.
+            psubscribe_quietly([pattern]) if still_wanted
```
petyaslavova: This is the only psubscribe write not atomic with its own `track_pending_acks` - the lock block closes at 837, so the write and the tracking are separate lock acquisitions. Every other site (205-207, `reissue_unconfirmed`, `issue_probes`, `mark_probes`) holds the lock across both.

A concurrent `subscribe` slotting in between leaves `@pending_acks[pattern]` in the opposite order from the wire, so a blocking batch retires from `@inflight_waits` on another command's ack - and a `CommandError` in that window then evicts a valid registration instead. That's the misattribution the comments at 130-138 guard against. Moving the call inside the existing lock block fixes it.

petyaslavova (changes requested):

lib/redis/keyspace_notifications/manager.rb
vladvildanov: Thanks to dive deeper into it before me. Now we do not have those race conditions and it's guarded with mutexes

lib/redis/keyspace_notifications/manager.rb
vladvildanov: Fixed

cluster/lib/redis/cluster/keyspace_notifications.rb
vladvildanov: Fixed

lib/redis/keyspace_notifications/manager.rb
vladvildanov: Fixed

lib/redis/keyspace_notifications/manager.rb
vladvildanov: Fixed

lib/redis/keyspace_notifications/manager.rb
```diff
@@ -0,0 +1,1323 @@
...
+        probe_queue = probing ? patterns.drop(1) : nil
+        # The session-opening command is one psubscribe per pattern, acknowledged
+        # like any later one: record the expected acks before issuing it (a failed
+        # connect raises into run_listener, whose session-end clear discards them).
+        # The session sequence is bumped alongside: waits key their single re-issue
+        # to it, and bumping before the connect lets a wait that sampled the old
+        # session retry against this one.
+        @lock.synchronize { @session_seq += 1 }
+        probing ? mark_probes(opening) : track_opening_acks(opening)
+        @redis.psubscribe(*opening) do |on|
```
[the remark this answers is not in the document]
vladvildanov: It's fixed

petyaslavova (approved):
I have just one small comment, but I suggest to merge this PR as is and adding the change in a separate PR. 
Cluster fan-out is still serial in cluster keyspace_notifications.rb:121-134. 
`subscribe` runs one ack-blocking call per pattern per node, sequentially, each with the 5 s SUBSCRIBE_ACK_TIMEOUT.

PR #1383, opened 2026-09-03
Added support for Redis functions (7.0)

vladvildanov:
Starting from 7.0 Redis supports server-side script execution API called Redis Functions. Check the [official documentation](https://redis.io/docs/latest/develop/programmability/functions-intro/) for detailed explanation.

---

> [!NOTE]
> **Medium Risk**
> New command surface and non-trivial distributed fan-out/kill semantics; behavior is gated to Redis 7.0+ and covered by tests but affects multi-node setups.
> 
> **Overview**
> Adds **Redis Functions** support (7.0+) alongside existing Lua scripting: `redis.function` for library lifecycle (load/replace, list, delete, flush, dump/restore with restore policies, stats, kill) and `redis.fcall` / `redis.fcall_ro` for invocation with the same keys/argv calling style as `eval`/`evalsha`.
> 
> **`FUNCTION LIST`** and **`FUNCTION STATS`** replies are normalized to nested hashes via new `HashifyFunctionList` and `HashifyFunctionStats`, so RESP2 flat arrays and RESP3 maps look the same in Ruby.
> 
> On **`Redis::Distributed`**, `function` fans out **once per physical server** (deduped by host/port/path, including sentinel-backed nodes) so server-global libraries are not loaded twice on different DBs; **`function(:kill)`** succeeds if any node returns OK and ignores idle nodes’ `NOTBUSY`. **`fcall`** / **`fcall_ro`** route like distributed `eval` (single key / same-slot keys only).
> 
> Coverage adds standalone, distributed, and cluster tests (CROSSSLOT, `no-writes` for `FCALL_RO`, pipelines).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff3723bd35dbb54ee33a20561007f3c243e61af8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/redis/commands.rb
+++ b/lib/redis/commands.rb
@@ -226,6 +226,34 @@ class Redis
       reply.map { |str| HashifyClusterNodeInfo.call(str) }
     }
 
+    # FUNCTION LIST: RESP2 replies with a flat [k, v, ...] array per library (and per
+    # function within it); RESP3 already returns maps. Converge on nested Hashes.
+    HashifyFunctionList = lambda { |reply|
+      reply.map do |library|
+        library = library.each_slice(2).to_h unless library.is_a?(Hash)
+        library["functions"] = library["functions"].map do |function|
+          function.is_a?(Hash) ? function : function.each_slice(2).to_h
+        end
+        library
+      end
+    }
+
+    # FUNCTION STATS: RESP2 replies with nested flat [k, v, ...] arrays; RESP3 with maps.
+    HashifyFunctionStats = lambda { |reply|
+      reply = reply.each_slice(2).to_h unless reply.is_a?(Hash)
+
+      running = reply["running_script"]
+      reply["running_script"] = running.each_slice(2).to_h if running.is_a?(Array)
+
+      engines = reply["engines"]
+      engines = engines.each_slice(2).to_h unless engines.is_a?(Hash)
+      reply["engines"] = engines.transform_values do |stats|
+        stats.is_a?(Hash) ? stats : stats.each_slice(2).to_h
+      end
+
+      reply
+    }
+
     Noop = ->(reply) { reply }
 
     # Sends a command to Redis and returns its reply.
--- a/lib/redis/distributed.rb
+++ b/lib/redis/distributed.rb
@@ -1357,8 +1357,33 @@ class Redis
     end
 
     # Manage Redis Functions libraries.
+    #
+    # Function libraries are server-global rather than per-database, so the fan-out runs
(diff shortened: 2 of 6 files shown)


lib/redis/commands.rb
```diff
@@ -298,6 +298,34 @@ module Commands
...
+      reply["running_script"] = running.each_slice(2).to_h if running.is_a?(Array)
+
+      engines = reply["engines"]
+      engines = engines.each_slice(2).to_h unless engines.is_a?(Hash)
+      reply["engines"] = engines.transform_values do |stats|
+        stats.is_a?(Hash) ? stats : stats.each_slice(2).to_h
+      end
+
+      reply
+    }
```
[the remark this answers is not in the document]
vladvildanov: The cluster driver document precisely that `special` policy is out of scope of request-response policies.

lib/redis/distributed.rb
```diff
@@ -1356,6 +1356,36 @@ def script(subcommand, *args)
       on_each_node(:script, subcommand, *args)
     end
 
+    # Manage Redis Functions libraries.
+    #
+    # Function libraries are server-global rather than per-database, so the fan-out runs
+    # once per physical server: ring nodes differing only by database would re-run
+    # mutations (e.g. FUNCTION LOAD) against the same server and fail on collision.
+    def function(subcommand, *args, **options)
+      function_nodes = nodes.uniq { |node| [node._client.path, node._client.host, node._client.port] }
```
petyaslavova: `node._client` is a plain `::RedisClient` for a sentinel-backed node (`Client.sentinel` passes `client_implementation: ::RedisClient`), and that class exposes only `config`/`id`/`server_url` - no `path`, `host` or `port`. So this line raises `NoMethodError` for any ring built like:

```ruby
Redis::Distributed.new([
  { name: "shard-a", sentinels: [{ host: "10.0.0.1", port: 26379 }] },
  { name: "shard-b", sentinels: [{ host: "10.0.1.1", port: 26379 }] }
])
```

It fails before any I/O, since `uniq` runs before the first command. Reaching one level deeper fixes it - both `RedisClient::Config` and `RedisClient::SentinelConfig` expose all three:

```ruby
nodes.uniq { |node| c = node._client.config; [c.path, c.host, c.port] }
```

petyaslavova (changes requested):

lib/redis/distributed.rb
vladvildanov: Fixed

PR #1386, opened 2026-09-08
Fixed XAUTOCLAIM omit Deleted entry from reply

vladvildanov:
New deleted entry from XAUTOCLAIM reply was added as a part of Redis 7.0. Currently, parser omits it and does not add it into a reply structure

---

> [!NOTE]
> **Low Risk**
> Additive response field with a safe default on older Redis; limited to stream consumer-group parsing and tests.
> 
> **Overview**
> **`xautoclaim`** now includes a **`'deleted'`** key in its return hash: an array of stream entry ids Redis removed from the pending list because the messages were already deleted from the stream (Redis **7.0+**). On **6.2**, that key is always **`[]`** because the server reply has no third element.
> 
> The **`HashifyStreamAutoclaim`** / **`HashifyStreamAutoclaimJustId`** parsers map `reply[2]` into **`'deleted'`**, and **`xautoclaim`** docs describe the full **`next` / `entries` / `deleted`** shape. Integration and mock tests cover deleted-stream cases and two-element 6.2 replies; **`redis_mock`** encodes **`nil`** as RESP null so 6.2-style deleted slots in the entries list can be exercised in tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da735b97e8c5527ec1c2cc2ca5999165621db070. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/redis/commands.rb
+++ b/lib/redis/commands.rb
@@ -219,19 +219,23 @@ class Redis
       end
     }
 
+    # XAUTOCLAIM: Redis 7.0 appends the ids of entries that were deleted from the stream and
+    # dropped from the PEL; 6.2 has no third element and reports them as nils inside entries.
     HashifyStreamAutoclaim = lambda { |reply|
       {
         'next' => reply[0],
         'entries' => reply[1].compact.map do |entry, values|
           [entry, values.each_slice(2)&.to_h]
-        end
+        end,
+        'deleted' => reply[2] || []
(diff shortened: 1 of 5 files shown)


petyaslavova (approved):
LGTM.

PR #1387, opened 2026-09-08
Added support for WAITAOF command (Redis 7.2)

vladvildanov:
> [!NOTE]
> **Medium Risk**
> Changes affect durability-wait semantics and cluster-wide aggregation; behavior is documented (including pipeline limitations) and covered by integration tests, but callers relying on strict per-connection guarantees in cluster pipelines should read the new docs.
> 
> **Overview**
> Adds **Redis 7.2 `WAITAOF`** across standalone, distributed, and cluster clients: a blocking server command that returns `[local, replicas]` fsync counts for writes on the current connection.
> 
> **Standalone** implements `waitaof` in `Commands::Server` via `send_blocking_command`, with YARD docs covering cluster/distributed behavior and pipeline caveats. **Distributed** fans out with `on_each_node` and returns one pair per ring node.
> 
> **Cluster** introduces driver-level **`command_routings`** (merged from `DEFAULT_COMMAND_ROUTINGS` + caller overrides) so `waitaof` uses `all_shards` fan-out; a thin `Redis::Cluster#waitaof` override collapses per-primary `[local, replicas]` replies with **per-position minimum** (`agg_min` semantics). Documentation in `AGENTS.md` and `specs/adding-commands.md` formalizes when to use routings vs reply-shaping overrides (alongside existing `himport_*` patterns).
> 
> Tests cover basic usage, AOF-enabled local fsync, errors, timeouts, argument ordering (mock), pipelines, cluster fan-out to all primaries, caller routing override, and distributed per-node results—all gated at Redis 7.2 where needed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64793ed664a0689922c2605329b6fe6cd9e5b500. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cluster/lib/redis/cluster.rb
```diff
@@ -194,12 +194,40 @@ def himport_discard_all
...
+      'waitaof' => { request_policy: 'all_shards' }.freeze
+    }.freeze
+    private_constant :DEFAULT_COMMAND_ROUTINGS
+
     def initialize_client(options)
+      command_routings = DEFAULT_COMMAND_ROUTINGS.merge(options[:command_routings] || {})
       # protocol defaults to 3 (RESP3) but a caller-provided protocol: in options overrides it.
       cluster_config = RedisClient.cluster(
         protocol: 3, **options,
+        command_routings: command_routings,
```
petyaslavova, line 236: This makes caller-supplied `command_routings:` a documented precedence contract, and the new test relies on it, but the option still is not in the `@option` list on `#initialize` (around line 83). One `@option options [Hash] :command_routings` line describing the merge order would make it discoverable.

Minor: the merge only dedupes String keys. A caller passing `{ waitaof: nil }` still wins, but only because the driver normalizes both entries and later insertions overwrite earlier ones - `.transform_keys(&:to_s)` on the caller hash would make that intentional rather than incidental.

lib/redis/commands/server.rb
```diff
@@ -134,6 +134,38 @@ def save
...
       end
 
+      # Block until the write commands previously sent on this connection have been fsynced to
+      # the append-only file of the local server and/or of the given number of replicas.
+      #
+      # The reply always reports how many local servers and replicas had fsynced the writes when
+      # the command returned, whether or not the requested levels were reached, so compare it
+      # against what you asked for. Inside `multi` the command does not block.
+      #
+      # On `Redis::Cluster` the command is sent to every primary and each count is the minimum
```
petyaslavova, line 156: This paragraph describes the non-pipelined path only. `WAITAOF` carries no `write` flag, so inside `pipelined`/`multi` the driver resolves it via `find_node_key` -> `any_replica_node_key(seed:)` and sends it to an arbitrary node - a random primary under the default `replica: false`, an actual replica under `replica: true` - not the node that carried the writes.

That makes the pipelined reply shape-correct but unrelated to the pipeline's writes: `pipelined { |p| p.set(k, v); p.waitaof(1, 0, 0) }` can come back `[1, n]` with the SET unfsynced. Could you add a sentence here (and mirror it in the `Redis::Cluster#waitaof` comment) saying the per-connection guarantee does not hold inside `pipelined`/`multi` on cluster?

cluster/test/commands_on_server_test.rb
```diff
@@ -198,4 +198,82 @@ def test_sync
...
+  def test_waitaof_reaches_every_primary
+    target_version "7.2.0" do
+      original = redis.config(:get, 'appendonly')['appendonly']
+      begin
+        redis.config(:set, 'appendonly', 'no')
+
+        # With AOF disabled every primary rejects numlocal=1, so the fan-out surfaces one error
+        # per primary rather than a single node's.
+        error = assert_raises(Redis::Cluster::CommandErrorCollection) { redis.waitaof(1, 0, 0) }
+        assert_equal redis.role.size, error.errors.size
```
petyaslavova, line 247: `redis.role.size` equals the primary count only under the default `replica: false` topology, where `PrimaryOnly` aliases `clients` to `primary_clients`. With `replica: true` `ROLE` fans out to all six nodes while the errors come from three primaries. Fine for this fixture, but worth a comment - or an explicit primary count - so the idiom is not copied into a replica-enabled test. `specs/adding-commands.md:286` recommends it generically, which is where it would spread from.

cluster/test/commands_on_server_test.rb
```diff
@@ -198,4 +198,82 @@ def test_sync
...
+      r.set('{a}foo', 'bar')
+
+      local, replicas = r.waitaof(0, 0, 0)
+
+      assert_equal 0, local
+      assert_kind_of Integer, replicas
+    end
+  end
+
+  def test_waitaof_in_pipeline_keeps_the_standalone_shape
```
petyaslavova: The name reads as an endorsement of the pipelined path, but this asserts the reply *shape*, not the durability semantics - the command lands on whichever node `any_replica_node_key` picks, so the pair says nothing about the pipeline's writes. Worth renaming to something like `test_waitaof_in_pipeline_returns_a_single_node_pair` and adding a comment that the pinning guarantee is lost there.

AGENTS.md
```diff
@@ -150,6 +150,14 @@ The `HIMPORT` command family (Redis 8.10) is the one place a command's state out
...
... override** (`waitaof`, `himport_*`): the driver still does the fan-out and hands back one reply per primary, and the override only collapses that array into the standalone shape. It has to be a method override, not a reply block on the shared command: the driver applies reply blocks **per node**, before it collects the fan-out, so a block never sees the aggregate. Aggregate the way the server's `response_policy` tip says (`agg_min` → per-position minimum, as `waitaof` does, so thresholds keep their single-node meaning; `agg_sum` → sum), not the way a sibling command happens to be handled. Guard it with `reply.first.is_a?(Array)` so a plain reply passes through (a caller who changed the routing, or a later driver that aggregates natively). Pipelines are unaffected by the missing override: `pipelined`/`multi` route each command to a single node, so the plain standalone shape comes back — add a cluster test that issues the command inside `pipelined` to lock that in. Prefer contributing the missing policy upstream over growing these overrides. Re-audit `DEFAULT_COMMAND_ROUTINGS` and the overrides on every `redis-cluster-client` bump (the gemspec pins an exact version for this reason).
```
petyaslavova: "Pipelines are unaffected by the missing override" is true of the reply shape but not of the command's meaning: single-node routing is exactly what breaks `WAITAOF`, because the node is chosen by `any_replica_node_key`, not by where the writes went. Please reword so the next person adding a fan-out command does not conclude the pipelined path is equivalent. Same wording in `specs/adding-commands.md:284`.

petyaslavova (changes requested):

cluster/lib/redis/cluster.rb
vladvildanov: Documented

lib/redis/commands/server.rb
vladvildanov: Fixed

cluster/test/commands_on_server_test.rb
vladvildanov: Fixed

cluster/test/commands_on_server_test.rb
vladvildanov: Fixed

AGENTS.md
vladvildanov: Fixed

petyaslavova (approved):
LGTM.

PR #1388, opened 2026-09-08
Fixed Redis 7.2 reply shapes

vladvildanov:
- Floatify crashes on "nan". 7.2 normalizes any NaN in a reply to the literal nan (RESP2 bulk string). Floatify (lib/redis/commands.rb:83) handles "inf"/"-inf" but falls through to Float(value), and         
Float("nan") raises ArgumentError. Add a when "nan" then Float::NAN branch (RESP3 already delivers a native double, so only the RESP2 path is affected).                                                     

- zrank(..., withscore: true) for a missing member likely returns [nil, nil] instead of nil. FloatifyPair (commands.rb:96) destructures a nil reply into [nil, Floatify.call(nil)]. Worth a test under both    
protocols and a nil guard.

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 3d4d5e44a0383dec1d4fe1f6db1cbd680f8dfe5b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/redis/commands.rb
+++ b/lib/redis/commands.rb
@@ -94,12 +94,16 @@ class Redis
       end
     }
 
+    # RESP2 carries doubles as bulk strings, including the non-finite ones Redis 7.2 normalizes
+    # to "inf", "-inf" and "nan"; RESP3 delivers native Floats, which pass through.
     Floatify = lambda { |value|
       case value
       when "inf"
         Float::INFINITY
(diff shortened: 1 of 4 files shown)


petyaslavova (approved):
LGTM.

PR #1389, opened 2026-09-10
Added support for missing HFE commands

vladvildanov:
Added support for missing HGETDEL/HGETEX/HSETEX commands

---

> [!NOTE]
> **Low Risk**
> Additive hash command wrappers with local argument validation; behavior depends on Redis 8.0 server support and can change field TTLs when expiry options are used.
> 
> **Overview**
> Adds Ruby client support for Redis **8.0** hash-field commands **`HGETDEL`**, **`HGETEX`**, and **`HSETEX`** on the main client and **`Redis::Distributed`**.
> 
> **`hgetdel`** returns values for the requested fields and removes them in one round trip, using the `FIELDS` count form like other hash-field APIs. **`hgetex`** reads fields and can optionally apply per-field expiry (`ex`/`px`/`exat`/`pxat`) or **`persist`**; incompatible expiry options are rejected in Ruby before sending. **`hsetex`** sets field/value pairs (array or hash, like **`hset`**) with optional **`fnx`**/**`fxx`** and the same expiry/ **`keepttl`** knobs, again with client-side mutual-exclusion checks.
> 
> Lint tests under **`target_version "8.0.0"`** cover splat/array fields, missing keys/fields, wrong types, TTL behavior, and **`ArgumentError`** cases for bad option combinations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50aed44c3b769f43aaf37a075ff2456fd7925fec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/redis/commands/hashes.rb
+++ b/lib/redis/commands/hashes.rb
@@ -158,6 +158,105 @@ class Redis
         send_command([:hdel, key].concat(fields))
       end
 
+      # Get the value of one or more hash fields and delete them.
+      #
+      # @example
+      #   redis.hset("hash", "f1", "v1", "f2", "v2")
+      #   redis.hgetdel("hash", "f1", "f2") # => ["v1", "v2"]
+      #   redis.hgetdel("hash", "f1", "f3") # => [nil, nil]
+      #
+      # @param [String] key
+      # @param [Array<String>] fields
+      # @return [Array<String, nil>] the value of each requested field, `nil` for fields that
+      #   did not exist
+      def hgetdel(key, *fields)
+        fields.flatten!(1)
+        send_command([:hgetdel, key, 'FIELDS', fields.length, *fields])
+      end
+
+      # Get the value of one or more hash fields and optionally set their expiration. HGETEX is
+      # similar to HMGET, but is a write command with additional options. When no options are
+      # provided, HGETEX behaves like HMGET.
+      #
+      # @example
+      #   redis.hset("hash", "f1", "v1", "f2", "v2")
+      #   redis.hgetex("hash", "f1", "f2") # => ["v1", "v2"]
+      #   redis.hgetex("hash", "f1", ex: 60) # => ["v1"]
+      #
+      # @param [String] key
+      # @param [Array<String>] fields
+      # @param [Hash] options
+      #   - `:ex => Integer`: Set the specified expire time on the fields, in seconds.
+      #   - `:px => Integer`: Set the specified expire time on the fields, in milliseconds.
+      #   - `:exat => Integer`: Set the specified Unix time at which the fields will expire, in
+      #      seconds.
+      #   - `:pxat => Integer`: Set the specified Unix time at which the fields will expire, in
+      #      milliseconds.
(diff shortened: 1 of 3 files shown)


lib/redis/commands/hashes.rb
```diff
@@ -158,6 +158,105 @@ def hdel(key, *fields)
...
+
+        args = [:hsetex, key]
+        args << "FNX" if fnx
+        args << "FXX" if fxx
+        args << "EX" << Integer(ex) if ex
+        args << "PX" << Integer(px) if px
+        args << "EXAT" << Integer(exat) if exat
+        args << "PXAT" << Integer(pxat) if pxat
+        args << "KEEPTTL" if keepttl
+        args.concat(['FIELDS', attrs.length / 2, *attrs])
```
petyaslavova, line 256: `numfields` is computed from the unflattened splat, so the array form documented in the `@param` tag on line 224 (`Array<String> | Hash<String, String>`, copied from `hset`) does not work. `r.hsetex("foo", ["f1", "v1"])` leaves `attrs == [["f1", "v1"]]`, so `attrs.length / 2` is `0`; redis-client's `flat_map` then flattens the nested array on the wire and the server receives `HSETEX foo FIELDS 0 f1 v1`, which `hsetexParseArgs` rejects with "invalid number of fields" (`numfields` is range-checked as 1..INT_MAX). `hset` accepts the same call, and `Redis::Distributed#hsetex` splats `*attrs` through unchanged, so it inherits this.

Adding `attrs.flatten!(1)` after the Hash normalization on line 245 fixes it and matches `hdel`/`hgetdel`/`hgetex`. Otherwise the array form should come out of the docstring.

test/lint/hashes.rb
```diff
@@ -68,6 +68,266 @@ def test_variadic_hdel
...
+      target_version "8.0.0" do
+        assert_equal 1, r.hsetex("foo", "f1", "v1", "f2", "v2")
+
+        assert_equal "v1", r.hget("foo", "f1")
+        assert_equal "v2", r.hget("foo", "f2")
+        assert_equal [-1, -1], r.httl("foo", "f1", "f2")
+      end
+    end
+
+    def test_hsetex_with_a_hash
```
petyaslavova, line 245: Could you add a `test_variadic_hsetex` here that passes the fields as a single array (`r.hsetex("foo", ["f1", "v1"])`), mirroring `test_variadic_hgetdel`? The splat and Hash forms are both covered, but the array form is not, which is why the `numfields` issue above passes CI.

petyaslavova (changes requested):

lib/redis/commands/hashes.rb
vladvildanov: Done

test/lint/hashes.rb
vladvildanov: Added

petyaslavova (approved):
LGTM

